PR #139, opened 2015-01-17
Parse uppercase headers in get_object and get_object_metadata

raincole:
There is no reason to assume server to always return HTTP headers in lowercase, so we had better handle both.

Also I fixed `put_bucket_policy` to use correct content-type(json).


ransomr:
Did you see an issue with header case? It looks to me like lhttpc already is making all headers lowercase. `lhttpc_client:read_response` calls `lhttpc_lib:canonical_header`.

sebmaynard:
We got bitten by this too - erlcloud's rebar.config refers to an lhttpc from talko's repository - that version calls canonical_header, but the version from esl (which we were using) doesn't.
Thankfully, as of right now, the talko repo is the same as 1.3.0 at esl with e17 fixes and that header fix, so was safe for us to switch; and that makes this change redundant (except for the extra header you added for bucket policy)

PR #141, opened 2015-01-19
Added a content_encoding to the headers returned from erlcloud_s3:get_object

sebmaynard:


ransomr:
Thank you for the contributions.

PR #143, opened 2015-01-23
Fix erlcloud_sqs:send_message/4 spec misses 'none' value

mkurkov:
It seems 'none' value is missing in send_message/4 spec, Dialyzer unhappy)


ransomr:
thank you

PR #144, opened 2015-02-05
Correct string start match

JoshRagem:
`["http"|_]` will match `[[$h,$t,$t,$p],_]` but not "http"
`[$h,$t,$t,$p|_]` will match "http"


ransomr:
Thank you.

PR #145, opened 2015-02-17
Added support for GetFederationToken

mbrix:
I added support for the STS::GetFederationToken call, and tried to match the assume role function's style and form.  I didn't add any specific Policy generating code, as I don't see that in the code base. Are you thinking of creating a policy record and support functions or just manually creating and encoding the policies? Some utility functions for that may be useful.

### DIFF
--- a/src/erlcloud_sts.erl
+++ b/src/erlcloud_sts.erl
@@ -4,6 +4,7 @@
 -include_lib("erlcloud/include/erlcloud_aws.hrl").
 
 -export([assume_role/4, assume_role/5,
+		get_federation_token/3,
 		get_federation_token/4]).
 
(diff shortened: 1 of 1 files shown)


ransomr:
Thank you. I don't personally use erlcloud_sts and don't have plans to enhance it. I welcome further pull requests.

mbrix:
How are you handling policies? Right now its just JSON encoded raw policy data, but could be a record, or a policy builder object. Were you heading in that direction?

ransomr:
I think it makes sense for calls like GetFederationToken to continue to take raw JSON. I think a policy builder module would be useful and could be used to generate the raw JSON for the input, but I think it makes sense to keep it separate from erlcloud_sts.

PR #146, opened 2015-02-20
adding S3 "Delete Multiple Objects" REST API implementation 

izinin:
implementation of http://docs.aws.amazon.com/AmazonS3/latest/API/multiobjectdeleteapi.html

used in http://www.jongla.com/

### DIFF
--- a/src/erlcloud_s3.erl
+++ b/src/erlcloud_s3.erl
@@ -12,6 +12,7 @@
          list_objects/1, list_objects/2, list_objects/3,
          list_object_versions/1, list_object_versions/2, list_object_versions/3,
          copy_object/4, copy_object/5, copy_object/6,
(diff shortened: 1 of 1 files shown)


ransomr:
thank you

ransomr:
Sorry, merged this too soon. Multiple issues:
1. This doesn't pass dialyzer - `make check-eunit` and make sure there isn't a new warning from s3
2. Why does Compose_xml need to be a fun? It is only used in one place. Please either just implement inline or as a private function.
3. erlcloud_aws:sign_v4 is called, but result is not used
4. please remove io:format call

Thanks.

izinin:
Hi Ransom,

Please see review my new pull request where all your comments are fixed (https://github.com/gleber/erlcloud/pull/149)

Regards,
Igor

PR #147, opened 2015-02-23
leveraging configure options for sns

marcelog:
Hello,

This pull allows one to configure the aws_config options in the same way as the other modules do. This is useful for including the region in the hostname and other use cases.


ransomr:
thank you

PR #150, opened 2015-03-08
Fix S3 rest api signing

### LINKED ISSUE #113: sqs list queues always fails with SignatureDoesNotMatch
master branch
receive_message however works well

snippet:
ok = ssl:start(),
ok = erlcloud:start(),
ok = erlcloud_sqs:configure(AwsAccessKey, AwsSecretKey, AwsHost),
Queues = try erlcloud_sqs:list_queues()
			catch Exception:Error -> io:format("exception: ~p, error: ~p~n", [Exception, Error]), []
			end.

http_error,403,"Forbidden",
                                 <<"<?xml version=\"1.0\"?><ErrorResponse xmlns=\"http://queue.amazonaws.com/doc/2012-11-05/\"><Error><Type>Sender</Type><Code>SignatureDoesNotMatch</Code><Message>The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.</Message><Detail/></Error><RequestId>b8262869-972d-5988-8342-64ce5cf34645</RequestId></ErrorResponse>">>}}

lrascao:
erlcloud_http:make_query_string can now take an extra parameter that informs it how to treat empty query parameters, either perform an empty assignment or no assignment at all

fixes #113

### DIFF
--- a/src/erlcloud_http.erl
+++ b/src/erlcloud_http.erl
@@ -1,12 +1,24 @@
 -module(erlcloud_http).
--export([make_query_string/1, url_encode/1, url_encode_loose/1]).
+-export([make_query_string/1, make_query_string/2, url_encode/1, url_encode_loose/1]).
 
+encode_query_term(Key, [], no_assignment) ->
+  [Key];
+encode_query_term(Key, [], empty_assignment) ->
+  [Key, "="];
+encode_query_term(Key, Value, _) ->
+  [Key, "=", url_encode(value_to_string(Value))].
+
+%% encode an empty value query string differently based on the
+%% argument provided, this is based on the fact that S3 and SQS
+%% sign url differently, S3 requires that empty arguments have no
+%% '=' (/?acl) while SQS requires it (/?QueuePrefix=)
+%% default behaviour is adding '='
 make_query_string(Params) ->
-    string:join([case Value of
-                     [] -> [Key, "="];
-                     _ -> [Key, "=", url_encode(value_to_string(Value))]
-                 end
-                 || {Key, Value} <- Params, Value =/= none, Value =/= undefined], "&").
+  make_query_string(Params, empty_assignment).
+
+make_query_string(Params, EmptyQueryOpt) ->
+    string:join([encode_query_term(Key, Value, EmptyQueryOpt) || {Key, Value} <-
+                  Params, Value =/= none, Value =/= undefined], "&").
 
 value_to_string(Integer) when is_integer(Integer) -> integer_to_list(Integer);
 value_to_string(Atom) when is_atom(Atom) -> atom_to_list(Atom);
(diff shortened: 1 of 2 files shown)


ransomr:
I'm a bit confused as I see two changes. One to make_query_string and one to the value of host that is used. This PR mentions make_query_string, and the issues mentions the host. Are both changes necessary?

Given that the changes are pretty subtle, could you please add code comments so that they won't be mistakenly reverted in the future? Thanks.

lrascao:
Apologies for the lack of clarity, both changes are necessary to get S3 working, root cause is the fact that Amazon signs the request differently in S3 and SQS, in S3 url's are expected to not have '='  (eg. /?acl) for empty arguments while SQS expects so (/?QueuePrefix=), this accounts for the make_query_string changes. While trying to find the issue brought up by @rawn i found out that by reverting the change i still could not get a successful S3 request, further digging revealed the host issue.

Will update the PR momentarily with the comments

lrascao:
@rawn could you please check that this branch fixes the issue you've detected?

rawn:
I will try to find time for it tomorrow.

rawn:
Sorry, this branch does not fix s3 for eu-west-1. In fact, no calls work now, all s3 calls throw "Sign atureDoesNotMatch" errors. At least the ones i tried.

Feel free to look at my quick 'hack' in git@github.com:campanja-forks/erlcloud.git.

lrascao:
sorry, my bad, turns out this whole host thing is unnecessary, got mislead in my tests and was actually validating against a v4 eu-central server, i think it should be fine now

ransomr:
@rawn do you have time to confirm that this fixes the issue? Thanks.

rawn:
I have tried it out and it works. Both S3 calls and SQS calls now work as expected, for my use case.

ransomr:
Great. Thanks for the contribution.

teodor-pripoae:
I'm seeing this error, when trying to upload files to `eu-central-1` region.

```
(ErlangError) erlang error: {:aws_error, {:http_error, 400, 'Bad Request', "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>InvalidRequest</Code><Message>The authorization mechanism you have provided is not supported. Please use AWS4-HMAC-SHA256.</Message><RequestId>9A0BEFF7D47D5BBA</RequestId><HostId>G737gwQTlOYHM15y1vOcim6QjqTZv72+k8GcUUM/tp7S7swPRr5CyrWp1+71ovVa1omkSS1TNJ4=</HostId></Error>"}}
```

I'm using `erlcloud_s3:put_object/5`. It used to work on `eu-west-1` until I moved my bucket to `eu-central-1`.

lrascao:
from what i gather, eu-central-1 exclusively uses v4 signing which erlcloud does not yet support for s3/sqs

PR #151, opened 2015-03-09
Use user's specified config rather than default_config

knoxcarey:
There's a small issue in the SQS code in which the default_config is being used rather than the one the user provided in the request. I believe that my edit reflects the original intention.


ransomr:
Thanks for the fix.

PR #152, opened 2015-03-12
Handling redrive policies in SQS 

knoxcarey:
SQS queues have an optional "redrive policy" that sends messages to a specified dead letter queue after a certain number of failed delivery attempts. When retrieving the attributes of a queue with such a policy, the attribute name ("RedrivePolicy") was not recognized. I added attribute encoder and decoder clauses, choosing not to further decode the JSON attribute value, in keeping with the way the "Policy" attribute is handled.


ransomr:
Thanks for the contribution.

knoxcarey:
You are quite welcome -- thanks for the great software!

-knox

PR #156, opened 2015-04-15
Add get_bucket_and_key/1 to extract them from URL

derek121:
Functions such as get_object/2 take a BucketName and Key. The new
convenience function get_bucket_and_key/1 pulls them out of a URL such
as "bucket_name.s3.amazonaws.com/path/to/key" (virtual-hosted-style URL)
or "s3.amazonaws.com/bucket_name/path/to/key" (path-style URL).

Details on S3 addressing at
http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html and
http://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAPI.html for

### DIFF
--- a/src/erlcloud_s3.erl
+++ b/src/erlcloud_s3.erl
@@ -28,13 +28,16 @@
          complete_multipart/4, complete_multipart/6,
          abort_multipart/3, abort_multipart/6,
          list_multipart_uploads/1, list_multipart_uploads/2,
-         get_object_url/2, get_object_url/3
+         get_object_url/2, get_object_url/3,
+         get_bucket_and_key/1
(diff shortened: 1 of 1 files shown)


derek121:
That is- this function extracts the BucketName and Key from a URL in order to then pass them to functions such as get_object/2.

derek121:
Slightly simplified the PR, since http_uri:parse/1 doesn't return the fragment portion of the URI (that's included via http_uri:parse/2 with appropriate option). So can do a single pattern match on its return tuple to get host and path.

ransomr:
Thanks for the contribution.

PR #157, opened 2015-04-16
Fix response header used to obtain version_id

derek121:
erlcloud_s3:get_object_torrent/3 used x-amz-delete-marker to obtain the
version_id from repsonse headers. Changed to x-amz-version-id (and
default to "null) as done in other functions such as put_object/6 and
copy_object/6.

This was mentioned in issue https://github.com/gleber/erlcloud/issues/138


ransomr:
Thanks for the fix

PR #158, opened 2015-04-16
Change default value for version_id to "null"

derek121:
erlcloud_s3:get_object_metadata/4 used a default value of "false" in its
proplists:get_value() call. Elsewhere, "null" is used as default for
version_id (e.g., get_object/4). That matches the docs at
http://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html

Related to issue https://github.com/gleber/erlcloud/issues/138


ransomr:
Why is "null" better than "false"? I would think undefined would be a better default value (used by most of the other attributes). But wouldn't either change potentially break existing clients?

derek121:
True- changing the default could cause problems for existing clients.

This change would just make it consistent with other functions in the module that look up x-amz-version-id in the response headers, which use null as a default (and matches the docs at the link I included).

But considering that existing clients may be relying on the false value, this indeed wouldn't be a safe change.

ransomr:
Thanks. Given that I'm going to close this for now. I labelled it api, so if there is a good time to make api changes in the future it can be found again.

PR #165, opened 2015-05-21
Parse erlcloud_elb:describe_loadbalancer output to tuples

supershal:
Added elastic load balancer parsing to retrieve ELB data.


ransomr:
Won't this break any existing clients that expect the current output?

PR #167, opened 2015-05-31
Support filtered describe instances request

lrascao:
Add filtering capability to the describe_instances api call, useful for example in obtaining
all ec2 instances that belong to an erlang cluster, they could either be grouped by placement group,
tag or by security groups. This offers a more flexible solution than the nodefinder_ec2 project (https://github.com/cstar/ec2nodefinder) which filters instances only by security group


ransomr:
Thanks for the contribution.

PR #168, opened 2015-06-01
Update elb API date version to 2012-06-01

mikev:
The updated 2012 API works with Amazon's VPC; 2009 API does not.  Tested register_instance.


ransomr:
Thank you.

PR #169, opened 2015-06-03
Fix generating direct links to S3 objects using temporary credentials

tomaszwojcikowski:
Existing implementation of EC2 metadata temporary credentials is incomplete (at least for S3).
While using temporary credentials it is required to include Security Token along with AWSAccessKey when accessing resources like S3 objects.

The following changes allow EC2 temporary credentials to be used to generate working direct links to S3 objects.


ransomr:
Thank you.

PR #170, opened 2015-06-05
DynamoDB new features implementation

smilerlee:
* New data types: BOOL, NULL, L, M
* Expressions: Projection, Condition, KeyCondition, Filter, Update
* ExpressionAttributeNames and ExpressionAttributeValues
* Improved ddb2_consumed_capacity record
* Global Secondary Index online creation and deletion
* DDB util: scan_all
* Other new features

Most of the code clean-ups are made in [Resort and rewrite some code to improve readability and flexibility](https://github.com/gleber/erlcloud/commit/f28932018da89a4020af379cae8d02db9044fbef),  you might prefer compare one-commit-by-one-commit, sorry for any inconvenience.

I believe there wouldn't be any backward compatibility breaking.

Hope you like it.
Thanks.

Smiler


ransomr:
Thanks. It will probably be a few days before I have a chance to take a look.

smilerlee:
Alright, take your time.

src/erlcloud_ddb2.erl
```diff
@@ -248,13 +261,42 @@ default_config() -> erlcloud_aws:default_config().
...
+                      {include, [attr_name()]} |
+                      all.
+
+-type global_secondary_index_def() :: {index_name(), key_schema(), projection(), read_units(), write_units()}.
 
 -type return_value() :: none | all_old | updated_old | all_new | updated_new.
 
+-type expression() :: binary().
+-type expression_attribute_names() :: [{binary(), attr_name()}].
+-type expression_attribute_values() :: [{binary(), in_attr_value()}].
```
ransomr: should these two use expression() instead of binary() for the key type?

ransomr:
Reviewed the code and it looks great. Before I merge I'm going to test using it in our service. It would be nice to update the doc examples to use the new attributes, since the old ones are deprecated.

src/erlcloud_ddb2.erl
smilerlee: I'm afraid not, neither of them is expression.
Attribute names act as placeholders for reserved or long names according to the document.
Attribute values are also placeholders. Obviously, typed or complex values couldn't be embed into expressions that easy.

smilerlee:
I'll update that a little later.

ransomr:
Thank you - great to have this updated.

src/erlcloud_ddb2.erl
```diff
@@ -1858,32 +2042,46 @@ update_item(Table, Key, Updates, Opts, Config) ->
 
 -type update_table_return() :: ddb_return(#ddb2_update_table{}, #ddb2_table_description{}).
 
--type global_secondary_index_update() :: {index_name(), pos_integer(), pos_integer()}.
--type global_secondary_index_updates() :: [global_secondary_index_update()].
+-type global_secondary_index_update() :: {index_name(), pos_integer(), pos_integer()} |
+                                         {index_name(), delete} |
```
alexturkin: @smilerlee out of curiosity: why order is like  `{index_name(), delete}` instead of more erlangis `{delete, index_name()} `?

smilerlee: The prefer way may be {create, ...}, {update, ...} and {delete, ...}  
But to keep a consistent style with the remaining exist code, I have chosen the {index_name(), delete} format.

smilerlee: If there will be any future DynamoDB updates which couldn't fit into this exist style, I will consider to add a secondary format as you said, coexist with current one.
Thanks for your reply.

alexturkin: @smilerlee thank you!

PR #173, opened 2015-06-17
take vsn from git itself

motobob:
this should address https://github.com/gleber/erlcloud/issues/172

### DIFF
--- a/src/erlcloud.app.src
+++ b/src/erlcloud.app.src
@@ -3,7 +3,7 @@
 
 {application, erlcloud,
  [{description, "Erlang cloud computing library"},
-  {vsn, "0.9.2"},
+  {vsn, git},
   {registered, []},
   {applications, [stdlib,
                   kernel,


smilerlee:
Thanks, but I don't think it's a good idea.

Searched the mail list, rebar would do something magic like: \(plugin source code could be found [here](https://github.com/erlware/rebar_vsn_plugin/blob/master/src/rebar_vsn_plugin.erl)\)
```
git describe --always --tags `git log -n 1 --pretty=format:%h .`
```

and I test it on `git version 1.9.5 (Apple Git-50.3)`...

* If `HEAD` is on the same commit with a tag (eg: `v0.10.0`), it would get the tag name:
```
v0.10.0
```
That's not what we need, we need `0.10.0`

* If `HEAD` is not on a tagged commit, it would get something like:
```
v0.9.2-39-gdee12f4
```
The `39` is the distance from `v0.9.2`, and the `dee12f4` is the commit id.  

I don't think neither of the results above is what we need.

Additionally, I also manually change the erlcloud app vsn to `{vsn, git}` in deps folder of my project, and typed in eshell
```
application:get_key(erlcloud, vsn).
```
It also get the similar output:
```
{ok,"v0.8.0-412-gda18104"}
```

**As a conclusion, I think this PR might cause potential problems.**

motobob:
your investigation is valid. This only one of the ways to solve this.
If your top level app will take not a tagged commit it will result in things like `{ok,"v0.8.0-412-gda18104"}`.
However, i think it is a valid behaviour. 
We use this all over the place in several projects and it does help to clearly identify the module version via `application:which_applications().`   
Secondly, if vsn will continue to be hardcoded - it will be always forgotten to be updated.
Lastly, I can hardly imagine anybody binding a specific vsn version in their top level apps. Ultimately it's just a string. 
Probably I miss your use case so could you clarify why you need `0.10.0` instead of `v0.10.0`?

smilerlee:
I agree totally.

Maybe the compromise solution could be changing the tag name strategy subsequently, strip off the v- prefix.

As why I prefer `0.10.0` instead of `v0.10.0` is because in Erlang world most of the vsn take the `major.minor.patch[-suffix]` pattern, without any prefix, I think it's better to obey this behavior.

ransomr:
Thanks.

PR #174, opened 2015-06-24
Support for lhttpc options

tomekowal:
`erlcloud_httpc` makes it possible to change underlying http clients,
however it doesn't allow to pass options to them.
This PR adds possibility to pass options to `lhttpc`
by adding `lhttpc_options` in `erlcloud_aws` record.
This comes in handy when you want to add a pool of workers to http client.

### DIFF
--- a/include/erlcloud_aws.hrl
+++ b/include/erlcloud_aws.hrl
@@ -35,6 +35,7 @@
           timeout=10000::timeout(),
           cloudtrail_raw_result=false::boolean(),
           http_client=lhttpc::erlcloud_httpc:request_fun(),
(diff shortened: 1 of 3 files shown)


ransomr:
I'd prefer not to add anything lhttpc specific to configuration. Since lhttpc is no longer supported I'd like to switch to a different default client. Instead you can set http_client in your aws_config to be something like the following, which won't break if the default client does change:

```
fun(URL, Method, Hdrs, Body, Timeout, _Config) ->
    lhttpc:request(URL, Method, Hdrs, Body, Timeout, [your_options])
end
```

PR #176, opened 2015-07-06
Support for UTF-8 characters

artefactop:
With this change the lib is able to send messages to SQS with UTF-8 characters


ransomr:
Thanks for the fix.

PR #177, opened 2015-07-08
Update erlcloud to use AWS v4 signature and use for AS and EC2

ncrothe:
Finally got around to fixing that v4 signature. I followed the initial idea of @ransomr and extended it to allow GETs as well (using canonical query string, cf. http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html).

I used it for EC2 and AS on eu-central-1 (which only allows v4) and as far as I can tell that works with both GET and POST.


ransomr:
Thank you. Glad to have this done.

PR #180, opened 2015-07-11
Add optional provisioned_throughput opt to update_table

smilerlee:
Since `ProvisionedThroughput` is an optional parameter to `UpdateTable`, click [here](http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateTable.html#DDB-UpdateTable-request-ProvisionedThroughput) for more information.

Backwards compatibility is fully considered in this PR.

### DIFF
--- a/src/erlcloud_ddb2.erl
+++ b/src/erlcloud_ddb2.erl
@@ -298,7 +298,8 @@ default_config() -> erlcloud_aws:default_config().
 -type return_consumed_capacity() :: none | total | indexes.
 -type return_item_collection_metrics() :: none | size.
 
--type out_attr_value() :: binary() | number() | boolean() | undefined | [binary()] | [number()] | [out_attr_value()] | [out_attr()].
+-type out_attr_value() :: binary() | number() | boolean() | undefined |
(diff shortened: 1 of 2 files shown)


nalundgaard:
+1... This is very similar to the work that I did in the [alertlogic fork](https://github.com/alertlogic/erlcloud/commit/2096c8c83c5173b648dcf21700349416fe3920e1#diff-bd982b35b7ba4ac9c8c4a766560403eeR2033), which was supposed to get PRed back here, but never did. It looks like this part of your change is 100% compatible with my change, which is awesome. I had also implemented support for adding/updating G2I in the above commit, and it wasn't QUITE compatible with how you implemented it in #170 (discussion [here](https://github.com/gleber/erlcloud/pull/170#discussion_r32824168)), and we ended up merging your change over mine and modified our client code to use your format.

Glad to see that in this case, it should merge clean for us!

smilerlee:
@nalundgaard thanks! I did read your work indeed before I wrote this PR. I was going to cherry pick that commit, but I messed it up. Additionally, there also be some other code need to be refactored, I wrote it again based on your code.

ransomr:
Thank you for the fix.

PR #181, opened 2015-07-11
Add legacy Exists and Value parameters to Expected

smilerlee:
Just in case someone might need that.

Assuming there's an existing expected condition which is _`{Name, Value, Op}`_, of which

  * _`Value`_ is atom `true`, which assumed to be inferred as `{"S": "true"`}; and
  * _`Op`_ is one of `eq`, `ne`, `le`, `lt`, `ge`, `gt`, `contains`, `not_contains`, `begins_with`.

Iff in this case, it becomes _equals to atom string `Op`_ (eg. `{"S": "lt"}`), which used to be _comparing with `{"S": "true"}`_

Actually, seldom people would have used atoms as string values. I believe no one would come into this circumstance incidentally.

### DIFF
--- a/src/erlcloud_ddb2.erl
+++ b/src/erlcloud_ddb2.erl
@@ -27,7 +27,7 @@
 %% `42'. To specify the AWS binary or set types an explicit `Type'
 %% must be provided. For example: `{b, <<1,2,3>>}' or `{ns,
 %% [4,5,6]}'. Note that binary values will be base64 encoded and
-%% decoded automatically. Since some atoms (such as `false', `not_null',
+%% decoded automatically. Since some atoms (such as `true', `false', `not_null',
 %% `null', `undefined', `delete', etc) have special meanings in some cases,
 %% use them carefully.
 %%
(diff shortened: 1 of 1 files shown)


ransomr:
I don't know why anyone would need this. {Name, Value, eq} and {Name, Value} are logically equivalent. Don't think we need to provide a way to use a logically equivalent way of doing this that has been deprecated twice.

smilerlee:
I totally agree that we should not use these twice-deprecated parameters any longer, and I did.

Actually, DynamoDB doesn't allow us to use `Exists`/`Value` combine with `AttributeValueList`/`ComparisonOperator` together, otherwise, a `ValidationException` will be returned.

But I think we'd better support both of them; or neither of them.
Adding these all relevant parameters, rather than a partial support, could help our users using them precisely, decreasing confusions, not mix-using them together by mistake.

And I tested that using `Exists = true` without `Value` parameter is also not allowed by DynamoDB, thus I only added `{attr_name(), true, in_attr_value()}`, not `{attr_name(), true}`.

It'll be cool if you could reconsider my proposal.

smilerlee:
By the way, the `Exists` can be omitted if `Value` presents, but I added it in dynamizing result for clarity.

ransomr:
Thank you

PR #182, opened 2015-07-13
Adding support for Amazon KMS

rod-eligible:
I just added support for Amazon KMS:

http://docs.aws.amazon.com/kms/latest/APIReference/Welcome.html


ransomr:
Thanks for the PR. A few issues:
1. Having optional parameters always passed to the methods is problematic. It's not extensible when the AWS API is extended. It's unwieldy for the callers and not self documenting at the call site. And as implemented it requires a function clause for each possible combination of omitted parameters which could lead to exponential implementation size. You could follow the pattern in erlcloud_ddb2, or just take all parameters as proplists.
2. Can you use erlcloud_retry instead of duplicating ddb_impl into kms_impl? I wrote erlcloud_retry with the hope that we wouldn't need to continue duplicating that code. Also be sure to update the error handling to deal intelligently with any KMS errors that need to be handled.
3. Please use spaces instead of tabs consistently.
4. Would be great to have a test suite.

rod-eligible:
ok, first time I try to commit something to a erlang library so I'm kind of new on this, I'll follow the points and let you know once done.
thx

zacharyfox:
You might want to take a look at my refactor here: https://github.com/alertlogic/erlcloud/pull/13. It's the start of addressing the comments above.

PR #183, opened 2015-07-13
hackney as a configurable http client

hemantkr:
Added support for using hackney as an http client with erlcloud. 

(cc @benoitc)


ransomr:
cherry-picked the one relevant commit

okeuday:
I attempted to use this with hackney tagged at 1.3.2 and got a `{tls_alert,"certificate unknown"}` error reason.  Is there something special that needs to happen with hackney before this works?

hemantkr:
Nothing else needed. Does lhttpc work fine?

okeuday:
@hemantkr Yes.  To avoid the problem I commented out the `#aws_config{http_client = hackney}` modification and stuff works fine.  However, it would be nice to be able to use hackney instead of lhttpc.

okeuday:
@hemantkr when testing this, I was using the `erlcloud_ddb2` module.

benoitc:
hrm what is the version of erlang? also seems odd since hackney have been
well tested with aws. does it work with an earlier version? ( without
partial chain handling)

benoitc:
the diference with lhttpc there afaik is that hackney is validating the
certificate. can anyone provides me a simple way to reproduce it?

okeuday:
@benoitc Using 18.0.  I can get a short example.

benoitc:
@okeuday thanks!

okeuday:
@benoitc https://gist.github.com/okeuday/c72d4f346ee0d83d8b5f is a short example of the problem (it requires replacing INVALID with the proper keys)

smilerlee:
@okeuday @benoitc DO NOT use 18.0 for production usage yet. I came across OOM crashes due to an unknown binary memory leak with 18.0 two months ago, and switching back to 17.x got it fixed. [Here](https://github.com/elixir-lang/elixir/issues/3504) is a related issue.

okeuday:
@smilerlee There are too many erts bugs in the downloadable 17.5 release, and with my testing I have found 18.0 to be more stable.

smilerlee:
@okeuday I have been using [17.4 rev. 2](http://packages.erlang-solutions.com/site/esl/esl-erlang/FLAVOUR_1_general/esl-erlang_17.4-2~ubuntu~trusty_amd64.deb) from erlang-solutions.com, maybe you can give it a try.

PR #184, opened 2015-07-17
Return status and do not raise errors in sns:publish/5 & derivatives

jasonwebster:
This begins the process to starting to align the sns module with many of the others in the app.

I'm not sure what the policy is on this project for introducing backwards incompatible API changes, but this does change the return spec of publish/5, publish_to_topic/* and publish_to_target/*.

### DIFF
--- a/src/erlcloud_sns.erl
+++ b/src/erlcloud_sns.erl
@@ -89,6 +89,8 @@
 
 -type(sns_subscribe_protocol_type () :: http | https | email | 'email-json' | sms | sqs | application).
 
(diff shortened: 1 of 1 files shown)


ransomr:
hi Jason. Thanks for the PR. You say that this begins the process. What are the full set of changes you have in mind? As much as possible I've been trying not to break backwards compatibility. It might make more sense to do this as a new module (erlcloud_sns2).

jasonwebster:
Sure, I'd be happy to do that.

I'll create a new sns module that updates _all_ exported functions to return status tuples instead of raise errors.

PR #186, opened 2015-07-20
Support multi-byte characters in erlcloud_http:url_encode/1

jasonwebster:
Fixes https://github.com/gleber/erlcloud/issues/185

I went the path of least resistance on this one by augmenting the current system. I think the ideal solution would be to replace this completely with something that operates on binaries instead of lists, similar to what hackney does https://github.com/benoitc/hackney/blob/ae65c297dab8bbfba886b80c8d04a3593ca022bf/src/hackney_lib/hackney_url.erl#L264.

### DIFF
--- a/src/erlcloud_http.erl
+++ b/src/erlcloud_http.erl
@@ -38,9 +38,8 @@ url_encode([Char|String], Accum)
        Char =:= $-; Char =:= $_;
        Char =:= $.; Char =:= $~ ->
     url_encode(String, [Char|Accum]);
-url_encode([Char|String], Accum)
-  when Char >=0, Char =< 255 ->
-    url_encode(String, [hex_char(Char rem 16), hex_char(Char div 16),$%|Accum]).
+url_encode([Char|String], Accum) ->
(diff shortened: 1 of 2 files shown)


ransomr:
Thank you. Is there a reason this doesn't implement support for all character ranges (4 byte UTF-8 codes)? Would be preferable to crash on those instead of encoding them incorrectly. Please also document that this is doing a UTF-8 encoding, preferably by using a `utf8_encode` method, or at least with a comment. Thank you for adding a unit test. Have you tested against AWS and verified it behaves as expected?

jasonwebster:
Thanks for the feedback. I'm going to refactor and clean it up a bit, push some new commits, and you can take another look shortly.

jasonwebster:
Alright, updated. Let me know if how you feel about that and I'll squash the commits before you merge.

jasonwebster:
Going to make a couple more changes first.

jasonwebster:
Should be all done.

I haven't tested other AWS APIs, but I can confirm that in the case of SNS this works correctly, and AWS does handle the UTF-8 encoding.

ransomr:
Thank you.

PR #191, opened 2015-07-29
Return sessionToken from erlcloud_sts:get_federation_token

### LINKED ISSUE #190: Return "sessionToken" from erlcloud_sts:get_federation_token
Federated user requests need to be authorized with access_key_id and secret_access_key. But amazon javascript sdk needs sessionToken to make requests (if it's not provided, then sdk will throw error "The AWS Access Key Id you provided does not exist in our records"). This I don't know why exactly SDK needs this sessionToken, but anyway it's better to provide all info about federated user.

alexdruzhilov:
fixes #190


ransomr:
Thank you for the PR

Issue #196, opened 2015-08-04
unicode tests do not pass on R16 but 


motobob:
@jasonwebster unicode conversion tests do not pass on R16 but work fine on R17.4+
I'll appreciate if you would be able to assist to fix this

```erlang
module 'erlcloud_http'
  module 'erlcloud_http_tests'
    erlcloud_http_tests:6: url_encode_test_...*failed*
in function erlcloud_http_tests:'-url_encode_test_/0-fun-0-'/0 (test/erlcloud_http_tests.erl, line 6)
**error:{assertMatch_failed,[{module,erlcloud_http_tests},
                     {line,6},
                     {expression,"erlcloud_http : url_encode ( \"foo bar¡~!\" )"},
                     {pattern,"\"foo%20bar%C2%A1~%21\""},
                     {value,"foo%20bar%C3%82%C2%A1~%21"}]}
    erlcloud_http_tests:7: url_encode_test_...*failed*
in function erlcloud_http_tests:'-url_encode_test_/0-fun-2-'/0 (test/erlcloud_http_tests.erl, line 7)
**error:{assertMatch_failed,[{module,erlcloud_http_tests},
                     {line,7},
                     {expression,"erlcloud_http : url_encode ( \"�\\200\\230single quotes�\\200\\231\" )"},
                     {pattern,"\"%E2%80%98single%20quotes%E2%80%99\""},
                     {value,"%C3%A2%C2%80%C2%98single%20quotes%C3%A2%C2%80%C2%99"}]}
    erlcloud_http_tests:8: url_encode_test_...*failed*
in function erlcloud_http_tests:'-url_encode_test_/0-fun-4-'/0 (test/erlcloud_http_tests.erl, line 8)
**error:{assertMatch_failed,[{module,erlcloud_http_tests},
                     {line,8},
                     {expression,"erlcloud_http : url_encode ( \"¢\" )"},
                     {pattern,"\"%C2%A2\""},
                     {value,"%C3%82%C2%A2"}]}
    erlcloud_http_tests:9: url_encode_test_...*failed*
in function erlcloud_http_tests:'-url_encode_test_/0-fun-6-'/0 (test/erlcloud_http_tests.erl, line 9)
**error:{assertMatch_failed,[{module,erlcloud_http_tests},
                     {line,9},
                     {expression,"erlcloud_http : url_encode ( \"��\\201� ��\\201�\" )"},
[...]

PR #197, opened 2015-08-05
erlcloud_http_tests: set encoding to utf-8 explicitly

### LINKED ISSUE #185: erlcloud_http:url_encode/1 does not support characters > 255
I noticed this when trying to send an SNS payload where the message contents included a right single quote `’`, or character `8217`. Trying to encode a something that contains a character outside of the ascii plane, it will eventually raise an error when it gets to that character:

```
{function_clause,
    [{erlcloud_http,url_encode,
        [[8217,115,32 ....
```

Code in question is https://github.com/gleber/erlcloud/blob/master/src/erlcloud_http.erl#L34-L43

I'm not entirely sure what the best path forward is here. I checked `http_uri:encode/1` and it just passes through the characters un-encoded. I'd be happy to submit a PR after the correct way forward is decided upon.

nalundgaard:
Resolves #185.

### DIFF
--- a/test/erlcloud_http_tests.erl
+++ b/test/erlcloud_http_tests.erl
@@ -1,3 +1,4 @@
+%% -*- coding: utf-8 -*-
 -module(erlcloud_http_tests).
 -include_lib("eunit/include/eunit.hrl").
 


ransomr:
thanks

nalundgaard:
Oops, I referenced the wrong issue. It closes #196. I think that will have to get closed manually.

PR #198, opened 2015-08-17
Configurable erlcloud_s3:delete_objects_batch

alexdruzhilov:
Add erlcloud_s3:delete_objects_batch/3 (with custom aws config). It's may be more useful than just working with a default aws config.


ransomr:
Thank you for the PR.

PR #199, opened 2015-08-26
bug fix : AWS S3 get bucket location call returns empty result for buckets in US region

izinin:
To address the problem reference implementation AWS tool s3cmd (http://s3tools.org/s3cmd) has been taken


ransomr:
Thank you

PR #200, opened 2015-08-27
Added a little flexibility to S3 to make this library easier to use with things like CEPH.

TBK145:
When using this library for our own server (which uses a S3 architecture/ implementation) we ran into a couple of problems. One was that we're using an http:// scheme and the other being we don't have a domain, but rather an IP.
Therefore I created 2 things, a function that replaces the entire config with the config you provide (erlcloud_aws.change_config/1), and 2 cases to create a different URI ([scheme, default_host, port, "/", Host, Path] instead of [scheme, Host, ".", default_host, port, Path]) when a boolean in the default_config record (s3_bucket_after_host, false at default) is set to true.
These changes were made to make this library a bit more flexible to use with your own S3 implementation (like CEPH).

### DIFF
--- a/include/erlcloud_aws.hrl
+++ b/include/erlcloud_aws.hrl
@@ -6,6 +6,7 @@
           s3_scheme="https://"::string(),
           s3_host="s3.amazonaws.com"::string(),
           s3_port=80::non_neg_integer(),
+          s3_bucket_after_host=false::boolean(),
           sdb_host="sdb.amazonaws.com"::string(),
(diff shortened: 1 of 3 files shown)


TBK145:
Thanks for the feedback, I made the changes you suggested.
I know that you can explicitly pass the config to all methods, but to me it just made no sense to pass the same parameter again and again. Therefore I think that this configure function would still be useful.

ransomr:
Thanks for the PR.

PR #201, opened 2015-08-27
ACL Grantee does not contain a nested 'owner'

Tolk-Haggard:
Noticed this bug when trying to set object acls. The grantee element has no owner sub-element.

Also, added a commit to handle URI based grantees.

http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGETacl.html

### DIFF
--- a/src/erlcloud_httpc.erl
+++ b/src/erlcloud_httpc.erl
@@ -53,8 +53,8 @@ request_httpc(URL, Method, Hdrs, <<>>, Timeout, _Config) ->
                                  [{body_format, binary}]));
 request_httpc(URL, Method, Hdrs, Body, Timeout, _Config) ->
     {value,
-     {_, ContentType}, HdrsRest} = lists:keytake(<<"content-type">>, 1, Hdrs),
-    HdrsStr = [{to_list_string(K), to_list_string(V)} || {K, V} <- HdrsRest],
+     {_, ContentType}, _HdrsRest} = lists:keytake("content-type", 1, Hdrs),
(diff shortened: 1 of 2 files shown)


src/erlcloud_httpc.erl
```diff
@@ -53,8 +53,8 @@ request_httpc(URL, Method, Hdrs, <<>>, Timeout, _Config) ->
                                  [{body_format, binary}]));
 request_httpc(URL, Method, Hdrs, Body, Timeout, _Config) ->
     {value,
-     {_, ContentType}, HdrsRest} = lists:keytake(<<"content-type">>, 1, Hdrs),
-    HdrsStr = [{to_list_string(K), to_list_string(V)} || {K, V} <- HdrsRest],
+     {_, ContentType}, _HdrsRest} = lists:keytake("content-type", 1, Hdrs),
+    HdrsStr = [{to_list_string(K), to_list_string(V)} || {K, V} <- Hdrs],
```
ransomr: Why is this change needed? I thought httpc would add the content-type header based on the content type passed in. In any case the headers should always be passed in as binary strings. If this is the right change, then you should use lists:keyfind instead of lists:keytake.

Tolk-Haggard: In our testing and reviewing the httpc request logic (in OTP 17.4) we did not see the content-type header being added. We ended up switching to httpc and modifying the header logic because we were generating empty content-type headers with lhttpc which caused our backend s3 store to choke. Good catch on using keyfind instead of keytake and searching after any binaries had been converted to strings.

ransomr:
Thanks for the contribution.

PR #202, opened 2015-09-02
SNS: added api for ListTopics

alexturkin:
This PR adds "ListTopics" action for SNS service. 
Also a bunch of tests for SNS module are added.

### DIFF
--- a/src/erlcloud_sns.erl
+++ b/src/erlcloud_sns.erl
@@ -585,9 +585,9 @@ list_all(Fun, Type, Config, Token, Acc) ->
     List = proplists:get_value(Type, Res),
     case proplists:get_value(next_token, Res) of
         "" ->
-            lists:append(Acc, List);
+            lists:foldl(fun erlang:'++'/2, [], [List | Acc]);
         NextToken ->
-            list_all(Fun, Type, Config, NextToken, lists:append(Acc, List))
+            list_all(Fun, Type, Config, NextToken, [List | Acc])
     end.
 
 extract_endpoint(Nodes) ->
--- a/src/erlcloud_sns.erl
+++ b/src/erlcloud_sns.erl
@@ -11,6 +11,8 @@
          create_topic/1, create_topic/2,
          delete_endpoint/1, delete_endpoint/2, delete_endpoint/3,
          delete_topic/1, delete_topic/2,
+         list_topics/0, list_topics/1, list_topics/2,
+         list_topics_all/0, list_topics_all/1,
          list_endpoints_by_platform_application/1,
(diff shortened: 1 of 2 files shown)


src/erlcloud_sns.erl
```diff
@@ -543,6 +580,15 @@ sns_request(Config, Action, Params) ->
             erlang:error({sns_error, Reason})
     end.
 
+list_all(Fun, Type, Config, Token, Acc) ->
+    Res = Fun(Token, Config),
+    List = proplists:get_value(Type, Res),
+    case proplists:get_value(next_token, Res) of
+        "" ->
+            lists:append(Acc, List);
```
ransomr: Calling `lists:append(Acc, List)` will need to copy `Acc` each call. More efficient to use `List ++ Acc` and then reverse the result when you are done.

ransomr:
I'm seeing a couple of new warnings from the tests:
```
test/erlcloud_sns_tests.erl:85: Warning: variable 'Key' exported from 'case' (line 78)
test/erlcloud_sns_tests.erl:89: Warning: variable 'Value' exported from 'case' (line 78)
```

src/erlcloud_sns.erl
smilerlee: Way like `erlcloud_ddb_util:flatreverse/1` is much more efficient

alexturkin:
@ransomr @smilerlee addressed

I've also moved flatreverse/1 to common module erlcloud_util, since it can be used from a lot of places which implement paging queries

smilerlee:
@alexturkin I would prefer not to extract this as a utility function since it's just a tiny implementation trick.

Any way, if I were you to implement this, I would declare an opaque type (for `Acc`), and three functions like:
* `new() -> []`
* `append(Acc, List) -> [List | Acc]`
* `final(Acc)` or `reverse(Acc)`

Perhaps they should reside in a standalone module like `list_appender` or something similar, maybe `erlcloud_` prefix is also needed.

However, I'd rather not to add this little convenience at the expense of complexity and exposing implementation details.

alexturkin:
@smilerlee I think you right, being an utility function is too much. I'll redo this

alexturkin:
@smilerlee please, take a look - I've put this trick just in `erlcloud_sns:list_all` function.

smilerlee:
@alexturkin seems great. I've never seen the usage of `fun erlang:'++'/2` before, it awesome!

ransomr:
Thanks for the contribution.

alexturkin:
@smilerlee @ransomr thank you guys!

PR #203, opened 2015-09-07
Implement a simple manager for multi-part uploads

jacobgreenleaf:
The existing API for consumers of this library for multi-part uploads is tedious to use, so I made a wrapper for it that will handle the lifecycle of calling `start_multipart`, then calculating the number of pieces necessary, then for each piece calling `upload_part`, accumulating the ETags, and finally calling `complete_multipart`, or calling `abort_multipart` if an error occurs. 

Notes:

* Currently uploads one at a time (part one, then part two, then part three). Could be done in parallel. 
* Currently split parts exactly on 10MB chunks. The smallest part size is 5MB, and it looks like the largest part size is 500MB. It could support variable numbers of parts. Files less than the smallest part size (5MB) are trivially supported. 
* Currently ONLY uses the `default_config()` AWS config and default HTTP options. I was happy to see the default configuration has a timeout, but with a default value of 10,000 ms, and a static chunk size of 10 MB, that means your uplink to S3 must be at least 1 megabyte per second, else uploading a ten megabyte file will time out. 
* Each file that is to be uploaded is allocated a unique upload ID by S3, and each upload ID is assigned a single worker by this library. The supervision strategy for the workers is `temporary`, so uploads will not be retried per se, but there is a `gen_server` that `monitor()`'s the workers as they are spawned, so that crashing workers should be cleaned up. This is important because partial uploads consume resources on S3 and you will be charged for partial uploads unless you clean up after them by calling `abort_multipart`.
* The interface to this library is a `gen_server:call`, i.e. synchronous. It might be nice to have a `cast` as well so that users can "fire-and-forget" an upload.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ erl_crash.dump
 .edts
 *[#]*[#]
 *[#]*
+
+*.sw[o-p]
--- a/README.md
+++ b/README.md
@@ -90,6 +90,16 @@ For usage information, consult the source code and refer to the API reference at
 - http://docs.aws.amazon.com/AmazonS3/latest/API/Welcome.html
 - http://docs.amazonwebservices.com/AmazonSimpleDB/latest/DeveloperGuide/
 
+## Multi-part uploads ##
+
+A small library is included for managing multi-part uploading of files in a fault-tolerant manner, `erlcloud_s3_multipart`. Example usage: 
+
+```
+erlcloud:start_s3_multipart_sup(). %% starts the multi-part supervisor; intended to be started by one of your supervisors
+erlcloud:put_object_multipart("/path/to/file", "bucket", "key").
+
+```
+
 ## Roadmap ##
 
 v0.8.0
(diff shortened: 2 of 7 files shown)


jacobgreenleaf:
Unfortunately there is an issue with uploading large files that causes a crash because the node runs out of memory. I think it is related to the way the worker gen_server handles the binaries it gets from reading each piece of the file. I'm keeping this PR open for overall review but don't advise merging.

ransomr:
Thanks - this looks like useful functionality.

It seems like the process model is overly complicated though. For serial part uploads it would be much simpler to just use the calling process. Use try/catch to ensure cleanup. This can be extended to support parallel uploads. See `erlcloud_ddb_util:get_all` for one possible approach.

Removing the supervisors would also remove map usage, which is important, since erlcloud is used on older Erlang versions. I'm fine with new functionality not supporting older versions, but you'd need to do something to make sure it still compiled. Maybe an approach like jsx uses? In any case simpler to remove the supervisors.

jacobgreenleaf:
I agree. I will refactor it to be a simpler method in erlcloud_s3.

jacobgreenleaf:
Closing for #206.

PR #206, opened 2015-09-13
Implement a simpler interface for multi-part uploads

jacobgreenleaf:
The existing API for consumers of this library for multi-part uploads is tedious to use, so I made a wrapper for it that will handle the lifecycle of calling `start_multipart`, then calculating the number of pieces necessary, then for each piece calling `upload_part`, accumulating the ETags, and finally calling `complete_multipart`, or calling `abort_multipart` if an error occurs. 

This is a refactor of my earlier pull request (#203). 

Notes:
- Currently uploads one at a time (part one, then part two, then part three). Could be done in parallel. 
- Currently split parts exactly on 10MB chunks. The smallest part size is 5MB, and it looks like the largest part size is 500MB. It could support variable numbers of parts. Files less than the smallest part size (5MB) are trivially supported. 
- Currently uses the default HTTP options. I was happy to see the default configuration has a timeout, but with a default value of 10,000 ms, and a static chunk size of 10 MB, that means your uplink to S3 must be at least 1 megabyte per second, else uploading a ten megabyte file will time out. 
- Files are opened in `raw` mode so that binaries aren't passed around on messages. I found that with `raw` mode disabled, that lots of uploads on a machine with small amounts of memory would cause either the OOM killer on Linux to kill BEAM, or a [crash dump](https://gist.github.com/jacobgreenleaf/ccece7c18274b525fb8e) would occur, with exactly two processes consuming 200MB each: the caller and the `file_io_server` process spawned because of no `raw` mode. Turning on raw mode fixed the issue, but I am not very experienced with Erlang's refc binaries and how they interact with messages to know if this is the correct solution.


ransomr:
Looks much better. Is the memory issue you mentioned before fixed? Is there a reason you didn't use a try/catch to make sure that abort is called? Seems like that would be a good idea.

jacobgreenleaf:
Memory issue is fixed after using raw mode for the file. I will add a catch
statement, thanks.

ransomr:
@jacobgreenleaf any update on this?

waisbrot:
I'm going to close this due to inactivity. If Jacob or anyone else wants to reactivate this, I think that's fine, but I'd rather not leave a PR lying around when it's not going to get merged without someone giving it more attention than it's gotten. Sorry for letting this go to rot, Jacob.

PR #210, opened 2015-10-01
erlcloud_ses: add support for MessageRejected errors

nalundgaard:
I missed this error code in my original implementation (mentioned [here](http://docs.aws.amazon.com/ses/latest/DeveloperGuide/ses-errors.html) in the AWS SES docs). I noticed it in a testing environment where SES wasn't configured. This makes erlcloud_ses return the error gracefully, instead of with an unceremonious `function_clause` error.


nalundgaard:
cc @motobob: once this is approved, I would like for us to pull it into alertlogic master.

ransomr:
thank you

PR #211, opened 2015-10-01
Function mismatch with httpc

mikebeyer:
Tuple mismatch on the 2nd param of httpc:request for puts/posts.  Mirrored the guard clause that exists in otp/httpc.erl

### DIFF
--- a/src/erlcloud_httpc.erl
+++ b/src/erlcloud_httpc.erl
@@ -46,7 +46,14 @@ request(URL, Method, Hdrs, Body, Timeout,
 request_lhttpc(URL, Method, Hdrs, Body, Timeout, _Config) ->
     lhttpc:request(URL, Method, Hdrs, Body, Timeout, []).
 
-request_httpc(URL, Method, Hdrs, <<>>, Timeout, _Config) ->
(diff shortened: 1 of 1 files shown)


ransomr:
If there is already a guard test in httpc.erl, why duplicate it here?

mikebeyer:
The issue we ran into is regarding a put in s3 for an object that has no body it causes a function clause error.  

https://github.com/erlang/otp/blob/OTP-17.4/lib/inets/src/http_client/httpc.erl#L147-L195

ransomr:
Thanks - now I understand. Could you please add a comment to the effect that the method needs to support put with no body and I'll merge.

mikebeyer:
I'd be happy to add some comments.  I'm not finding a lot of them throughout the code, so I want to ensure I match what you would expect there.  Do you want something right above the function explaining the guard?

ransomr:
Yes, that would be great. Thanks.

mikebeyer:
Ok, comment has been added.  Let me know if you'd like any more detail.

ransomr:
Thank you

PR #212, opened 2015-10-05
Removed deprecated erlang:now.

TBK145:
erlang:now is deprecated in Erlang 18 (http://www.erlang.org/news/88), so I replaced it with erlang:monotonic_time in accordance to the Erlang user guide (http://www.erlang.org/doc/apps/erts/time_correction.html#Dos_and_Donts)

### DIFF
--- a/src/erlcloud_sdb.erl
+++ b/src/erlcloud_sdb.erl
@@ -333,7 +333,7 @@ extract_item(Item) ->
     ].
 
 sdb_request(Config, Action, Params) ->
-    case sdb_request_with_retry(Config, Action, Params, 1, ?SDB_TIMEOUT, erlang:now()) of
+    case sdb_request_with_retry(Config, Action, Params, 1, ?SDB_TIMEOUT, erlang:monotonic_time()) of
         {ok, {Doc, Metadata}} ->
             {Doc, Metadata};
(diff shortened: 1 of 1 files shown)


smilerlee:
`erlang:monotonic_time` was newly added in Erlang 18, but we should also support old Erlang versions.

motobob:
I would agree with @smilerlee. This library is used by many and with older versions e.g. we still have R16B3 and R17.5, so unless it's declared at compile time it should be introduced since it will break backward compatibility.

TBK145:
Ah, yes totally didn't think about that.
I guess it would be possible to just replace erlang:now with erlang:timestamp, still making it Erlang 18 compatible?

ransomr:
You could probably just switch to use `os:timestamp`. Or you could use http://www.erlang.org/doc/apps/erts/time_compat.erl

smilerlee:
Yeah, `erlang:timestamp` is new too, and equals to old `os:timestamp`

TBK145:
I removed the monotonic timestamps and used os:timestamp instead. Now it should be Erlang 18 and backwards compatible.

TBK145:
Ok, this is getting messy, I'll create a new one (it's been a long day).

PR #213, opened 2015-10-05
Erlang 18 compatibility

TBK145:
Changed erlang:now() (deprecated in Erlang 18) to os:timestamp.
This way it should be compatible with Erlang 18 as well as older versions.
(Old PR: https://github.com/gleber/erlcloud/pull/212)


ransomr:
Thanks for the fix

PR #214, opened 2015-10-05
Travis

motobob:
Introduce use of Travis into the project alike it was done in our fork.

to make it working @ransomr need to register & allow access For Travis to gleber/erlcloud. Please follow http://docs.travis-ci.com/user/getting-started/.

Goal:
this will make auto running `make eunit` upon every PR against two Erlang versions.

Issue #209, opened 2015-09-22
Merging alertlogic fork into upstream


motobob:
Hello @ransomr  
I would really like fix the past problem of our fork being significantly diverged from the upstream while having quite a lot of good features needed in the upstream either.
I'm talking about submitting a PR from: https://github.com/gleber/erlcloud/compare/master...alertlogic:master.
Would you be okay reviewing such a big PR? I'm not sure how to split it into just several ones while keeping the feature set not broken,

As you may notice it has a major differences as here is a brief overview:
- Full support for SigV4 across all modules. According to AWS, all services support SigV4 in all regions and only a few still have SigV2 therefore SigV2 funs were not removed if smb still needs them 
- S3 follow redirect option was added
- KMS update, basically reworked #182
- DirectConnect described APIs added
- RDS APIs added 
- Route53 added
- Enabling travis CI builds for PRs. To enable it for you, permission should be set correctly
- `erlcloud_autoscalling` added along with some edits `erlcloud_as` to have backward compatibility
- Large number of maintenance defect fixes here and there
- enhancements to  DDB2, EC2 (adding all the gateways)  and Kinesis
- enriching ELB APIs
- adding more testing

smilerlee:
@motobob Wow! What a big PR!

How about...
* create a new branch from `gleber/master` (let's call it `A`), and...
* reproduce features one by one on this newly created branch `A` (this can be archived by checkout `master`, reset to `A` with `--mixed`, then commit hunks by hunks), and...
* revert unnecessary changes on your local `master`...
* finally branch `A` and your local `master` will become identical
* push `master` to `alertlogic/master`, push `A` to upstream
* after merging `A`, you'll be able to merge upstream changes I suppose.

Maybe you can make some tests.

ransomr:
Wow - took a quick look and there's a ton of good stuff there. I'd love to get this merged, so while I'd much prefer smaller PRs going forward, I'm willing to take this as one large PR. Please open it from a branch that isn't going to change while I'm reviewing it though - might take me a while to get through it all.

motobob:
@ransomr @smilerlee thank you for feedback.
Let me do the following:
- i'll pin our master to some branch and open the "master" PR to keep track of it.
- i'll open several small PRs for auxiliary things and where completely new module was added. That will minimize the delta of the "master PR". I doubt i'll be able two major features Sigv4 and S3 redirect, but at least that will simplify reviewing. 

hope that works with you

[...]

PR #215, opened 2015-10-05
[Do not merge] - for review purposes. Bringing Alertlogic 1.0.5 into upstream

motobob:
@ransomr as agreed in #209 here is the locked `master` PR to bring our features into upstream.

I would really like fix the past problem of our fork being significantly diverged from the upstream while having quite a lot of good features needed in the upstream either.

As you may notice it has a major differences as here is a brief overview:
- Full support for SigV4 across all modules. According to AWS, all services support SigV4 in all regions and only a few still have SigV2 therefore SigV2 funs were not removed if smb still needs them
- S3 follow redirect option was added
- KMS update, basically reworked #182
- DirectConnect described APIs added
- RDS APIs added
- Route53 added
- Enabling travis CI builds for PRs. To enable it for you, permission should be set correctly
- erlcloud_autoscalling added along with some edits erlcloud_as to have backward compatibility
- Large number of maintenance defect fixes here and there
- enhancements to DDB2, EC2 (adding all the gateways) and Kinesis
- enriching ELB APIs
- adding more testing

me, @alexturkin and @lilrooness will work on splitting this is several small one like #214 to help reduce the delta within this PR doing it module by module. I think over the course of next two week we will nail it down.

This is not to be merged as is therefore is marked **[Do not merge]**  unless you're really comfortable bringing such a huge delta at once.


smilerlee:
@motobob I think you misread my comment. What I meant was keeping the binary format is enough.

Showing you my unfinished work was for revealing the complexity; and I won't make any further changes or PRs for now.

I'll make some use cases analyses later.

motobob:
@smilerlee oh...sorry, this makes sense now. I'll submit and enhancement request to our fork to address your proposal.

motobob:
this is not going to happen this way. 
AL fork is still going to be merged into upstream (via FF) but is to be communicated well to the rest of `erlcloud` community as major version bump. 
stay tuned for a Readme roadmap update.

PR #218, opened 2015-10-08
erlcloud_s3 module now uses signv4

lilrooness:
### Problem
- Amazon s3 uses some different approach to sign their requests

### Solution
- Amended to work with Amazon s3

### DIFF
--- a/include/erlcloud_aws.hrl
+++ b/include/erlcloud_aws.hrl
@@ -1,88 +1,88 @@
 -record(aws_config, {
-  as_host="autoscaling.amazonaws.com"::string(),
-  ec2_host="ec2.amazonaws.com"::string(),
-  iam_host="iam.amazonaws.com"::string(),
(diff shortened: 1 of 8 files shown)


include/erlcloud_aws.hrl
```diff
@@ -6,9 +6,11 @@
           s3_scheme="https://"::string(),
           s3_host="s3.amazonaws.com"::string(),
           s3_port=80::non_neg_integer(),
+          s3_follow_redirect=false::boolean(),
```
motobob: this if for other feature namely support for S3 follow redirect. please amend it for this PR

motobob:
this is part and preparation for #215

alexturkin:
I think it is better to move  s3-related changes into separated s3-related PR.

I've created  https://github.com/gleber/erlcloud/pull/219 removing them and also other things you've commented here.

@lilrooness please close this PR

PR #219, opened 2015-10-09
erlcloud_aws changes, merging code from alertlogic repo

alexturkin:
this PR is part and preparation for https://github.com/gleber/erlcloud/pull/215

the purpose of this PR is mostly to commit common changes from `erlcloud_aws` module which will simplify further PRs to come

### What's changed? 
* adding x-amz-content-sha256 to header for signv4 (it will be needed for s3 related changes in further s3-related PRs)
* refactoring: move common code for retrieving region for host to one place
* https://github.com/alertlogic/erlcloud/pull/20 - added Reporting of AWS statuses using

All further PRs are expected to be as one PR per each AWS service.

### DIFF
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -433,24 +433,45 @@ authorization(Config, CredentialScope, SignedHeaders, Signature) ->
 
 %% This function fetches http://status.aws.amazon.com/data.json
 %% and examine "current" section for on going AWS issues/failures.
--spec get_service_status(string()) -> ok | list().
-get_service_status(ServiceName) when is_list(ServiceName) ->
-    {ok, Json} = aws_request_form(
-        get, "http", "status.aws.amazon.com", undefined, 
+%% Example of a return status:
+%% [{<<"service_name">>,
+%%   <<"Amazon Elastic Compute Cloud (Frankfurt)">>},
+%%  {<<"summary">>,<<"[RESOLVED] Internet Connectivity ">>},
+%%  {<<"date">>,<<"1436972949">>},
+%%  {<<"status">>,1},
+%%  {<<"details">>,<<>>},
+%%  {<<"description">>,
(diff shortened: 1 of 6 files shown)


src/erlcloud_aws.erl
```diff
@@ -194,15 +202,23 @@ format_timestamp({{Yr, Mo, Da}, {H, M, S}}) ->
 default_config() ->
     case get(aws_config) of
         undefined ->
-            #aws_config{access_key_id=os:getenv("AWS_ACCESS_KEY_ID"),
-                        secret_access_key=os:getenv("AWS_SECRET_ACCESS_KEY")};
+            AccessKeyId = case os:getenv("AWS_ACCESS_KEY_ID") of
+                              false -> "";
```
ransomr: Is there a reason that access_key_id needs to be a string? It seems like `undefined` would be a more conventional value if it isn't set, and would be less likely to break things like guard tests that look for lists.

src/erlcloud_aws.erl
```diff
@@ -409,9 +424,54 @@ signing_key(Config, Date, Region, Service) ->
...
+            lists:any(
+                fun(InputService)-> 
+                    ServiceNameBin = list_to_binary(InputService),
+                    ServiceNameLen = byte_size(ServiceNameBin),
+                    case proplists:get_value(<<"service">>, S) of 
+                        <<ServiceNameBin:ServiceNameLen/binary, _/binary>> -> true;
+                        _ -> false
+                    end
+                end, 
+            ServiceNames)
```
ransomr: please fix indent - `ServiceNames` and `Statuses` should match corresponding `fun`

src/erlcloud_aws.erl
alexturkin: agree. I'm not sure there was specific reason for that. Anyway, tests are fine with that both in your repo and ours, so I've changed this to be undefined

src/erlcloud_aws.erl
alexturkin: ok

alexturkin:
@ransomr I've addressed your comments, please take a look.

We have a bunch of branches coming up with a lot of other changes, but they're based on this ones, so after this will be merged, there will be a lot of other PRs :)

ransomr:
Thank you very much for the contributions - looking forward to the rest of it.

PR #228, opened 2015-11-03
Set explicit version on dependencies

ransomr:


motobob:
@ransomr just noticed this when was trying to merge upstream. 
was there any specific reason for the change? any observed issues?

ransomr:
Referencing master (instead of a tagged release) means erlcloud could have relied on untested or broken code. I don't think there was an issue at the time I made this change, but there have been in the past where erlcloud has broken as a result of changes in dependencies.

PR #229, opened 2015-11-06
Added get/list/describe api bindings in new erlcloud_cloudformation module

lilrooness:
### Problem
* Would like to be able to call get/list/describe functions for aws cloudformation

### Solution
* New Cloudformation module with all non PUT/DELETE API functions


ransomr:
thanks for the contribution - great to have some cloudformation support.

PR #232, opened 2015-11-25
Return AZ information in DescribeSpotPriceHistory response

sportlane:
Different AZ's have different spot price history. New API version returns that information in the response.


ransomr:
Thanks for the contribution.

PR #234, opened 2015-12-07
Change cloudwatch service name for monitoring

nogates:
Hello,

I tried to push some metrics to AWS Cloudwatch and I got this error:

```
"<ErrorResponse xmlns="http://monitoring.amazonaws.com/doc/2010-08-01"><Error>    <Type>Sender</Type> 
<Code>SignatureDoesNotMatch</Code>
<Message>Credential should be scoped to correct service: 'monitoring'. </Message>
```

I think that the name of the Service should be `monitoring` instead of `cloudwatch`. Changing it fixed the problem for me.

### DIFF
--- a/src/erlcloud_mon.erl
+++ b/src/erlcloud_mon.erl
@@ -260,7 +260,7 @@ mon_query(Config, Action, Params, ApiVersion) ->
                                  Config#aws_config.mon_port,
                                  "/",
                                  QParams,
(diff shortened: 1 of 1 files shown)


motobob:
+1, yes it should be `monitoring`

ransomr:
thanks for the fix

nogates:
thanks for the project! :)

PR #236, opened 2015-12-21
add volumeType to create_volume/describe_volumes

SergejJurecko:


ransomr:
Thank you for the contribution.

PR #237, opened 2015-12-23
Add delete_tags

lhaskins:
This adds a call to the EC2 `delete_tags` function. Works pretty much the same as `create_tags`, except when no tags are supplied, it deletes all tags for the supplied instance.


ransomr:
Thank you for the contribution.

PR #238, opened 2015-12-29
[DynamoDB] Add write_all, put_all and delete_all

smilerlee:
* Add `write_all` to `erlcloud_ddb_util.erl`, using parallel `batch_write_item` calls;
* Add two convenient functions sets: `put_all` and `delete_all`, use `write_all` as backend;
* Update some **unexported** type specs in `erlcloud_ddb_util.erl` for clarity.

### DIFF
--- a/src/erlcloud_ddb2.erl
+++ b/src/erlcloud_ddb2.erl
@@ -68,7 +68,7 @@
 %% to handle conditional check failures, match `{error,
 %% {<<"ConditionalCheckFailedException">>, _}}'.
 %%
-%% `erlcloud_ddb_impl' provides a higher level API that implements common
+%% `erlcloud_ddb_util' provides a higher level API that implements common
 %% operations that may require multiple DynamoDB API calls.
 %%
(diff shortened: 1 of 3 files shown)


nalundgaard:
This is very cool, but I am wondering if it's a good idea to parallel batch write? Doesn't that have the potential to consume a lot of write throughput and trigger rate limiting? I suppose that each batch process must get the benefit of erlcloud_ddb2's retry logic on throttling, but I can only imagine that running a number of requests in parallel significantly increases the chances that one of them gets throttled beyond the threshold. I guess this could be mitigated by a client with a custom retry function, but I wonder if it isn't advisable to provide some simpler control for this.

smilerlee:
@nalundgaard Instead of writing hundreds of or even thousands of large items, I suppose these functions are mean to small batches of small items: ensure unprocessed items are handled properly, and return as soon as possible. Anyway, 25 items is too small in some cases, and handling unprocessed items manually is a bit of trivial.

ransomr:
Thanks!

PR #240, opened 2016-01-05
correct typing for s3 functions

JoshRagem:
`iodata() :: iolist() | binary()`

since the functions take either 'iolist' or 'binary', 'iodata' is the correct type. It is unclear if `make_get_url` should still return `iolist`.


motobob:
thank you for contribution. Full `dializer` cleanup it what was always missing in the library and I hope will find some time in future for this.

PR #241, opened 2016-01-19
EC2 spot fleet support

sportlane:
request_spot_fleet, cancel_spot_fleet_requests, describe_spot_fleet_instances, modify_spot_fleet_request functionality.

### DIFF
--- a/src/erlcloud_ec2.erl
+++ b/src/erlcloud_ec2.erl
@@ -104,6 +104,7 @@
          create_spot_datafeed_subscription/3,
          delete_spot_datafeed_subscription/0, delete_spot_datafeed_subscription/1,
          describe_spot_datafeed_subscription/0, describe_spot_datafeed_subscription/1,
+         describe_spot_fleet_instances_all/1, describe_spot_fleet_instances_all/2,
          describe_spot_fleet_instances/1, describe_spot_fleet_instances/2,
          describe_spot_fleet_instances/3, describe_spot_fleet_instances/4,
          describe_spot_instance_requests/0, describe_spot_instance_requests/1,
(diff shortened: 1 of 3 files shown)


src/erlcloud_ec2.erl
```diff
@@ -2415,6 +2420,178 @@ request_spot_instances(Request, Config) ->
             {error, Reason}
     end.
 
+
+decode_spot_fleet_state("submitted") -> submitted;
```
motobob: is there really need this `list_to_atom()` conversion?
could you make it inline with the rest of the module and just leave it as `{atom(), list()}`

src/erlcloud_ec2.erl
```diff
@@ -2415,6 +2420,178 @@ request_spot_instances(Request, Config) ->
...
+-spec(describe_spot_fleet_instances/2 :: (string(), string()) -> {ok, [proplist()]} | {error, term()}).
+describe_spot_fleet_instances(SpotFleetRequestId, NextToken) ->
+    describe_spot_fleet_instances(SpotFleetRequestId, NextToken, undefined, default_config()).
+
+-spec(describe_spot_fleet_instances/3 :: (string(), string(), pos_integer()) -> {ok, [proplist()]} | {error, term()}).
+describe_spot_fleet_instances(SpotFleetRequestId, NextToken, MaxResults) ->
+    describe_spot_fleet_instances(SpotFleetRequestId, NextToken, MaxResults, default_config()).
+
+-spec(describe_spot_fleet_instances/4 :: (string(), string(), pos_integer(), aws_config()) -> {ok, [proplist()]} | {error, term()}).
+describe_spot_fleet_instances(SpotFleetRequestId, NextToken, MaxResults, Config) ->
```
motobob: it would be really beneficial if you add `describe_spot_fleet_instances_all` right away also which will handle the token pagination and return the full list. Examples can be found in other modules

src/erlcloud_ec2.erl
```diff
@@ -2415,6 +2420,178 @@ request_spot_instances(Request, Config) ->
...
+
+-spec(modify_spot_fleet_request/4 :: (string(), non_neg_integer(), default | no_termination, aws_config()) -> ok | {error, term()}).
+modify_spot_fleet_request(SpotFleetRequestId, TargetCapacity, ExcessCapacityTerminationPolicy, Config) ->
+    Params = [
+        {"ExcessCapacityTerminationPolicy", encode_termination_policy(ExcessCapacityTerminationPolicy)},
+        {"SpotFleetRequestId", SpotFleetRequestId},
+        {"TargetCapacity", TargetCapacity}],
+
+    case ec2_query2(Config, "ModifySpotFleetRequest", Params, ?NEW_API_VERSION) of
+        {ok, _Doc} -> ok;
```
motobob: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ModifySpotFleetRequest.html this request may fail so please return the actual error in the doc

motobob:
in general, I see no tests added along with the PR. I understand that ec2_tests module does not contain much at the moment but I would like to change this going forward for all new code added
at minimum this function `request_spot_fleet` must be covered.

src/erlcloud_ec2.erl
sportlane: As far as I understand AWS API "return" field can only be "true". Upon any error HTTP 400 is returned and {error, term()} tuple is returned.

motobob: yeah, sorry missed that.

sportlane:
Tests added

motobob:
@sportlane thank you for contribution.

PR #242, opened 2016-01-29
Add support for AWS Autoscaling API for DetachInstances

ncrothe:
Allows to using the AWS AS detach instances API function including the option to decrease or keep desired capacity.

### DIFF
--- a/src/erlcloud_as.erl
+++ b/src/erlcloud_as.erl
@@ -15,7 +15,8 @@
          terminate_instance/1, terminate_instance/2, terminate_instance/3,
 
          suspend_processes/1, suspend_processes/2, suspend_processes/3,
-         resume_processes/1, resume_processes/2, resume_processes/3]).
+         resume_processes/1, resume_processes/2, resume_processes/3,
+         detach_instances/2, detach_instances/3, detach_instances/4]).
 
 -define(API_VERSION, "2011-01-01").
 -define(DEFAULT_MAX_RECORDS, 20).
@@ -49,6 +50,11 @@
         "/SuspendProcessesResponse/ResponseMetadata/RequestId").
 -define(RESUME_PROCESSES_ACTIVITY, 
         "/ResumeProcessesResponse/ResponseMetadata/RequestId").
+
+%% xpath for detach instances:
+-define(DETACH_INSTANCES_ACTIVITY, 
+        "/DetachInstancesResponse/DetachInstancesResult/Activities/member").
+
 
 %% --------------------------------------------------------------------
 %% @doc Calls describe_groups([], default_configuration())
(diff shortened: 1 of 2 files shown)


src/erlcloud_as.erl
```diff
@@ -367,6 +373,44 @@ resume_processes(GroupName, ScalingProcesses, Config) ->
...
+    detach_instances(InstanceIds, GroupName, erlcloud_aws:default_config()).
+
+%% --------------------------------------------------------------------
+%% @doc Detach the given instances from the group.  The 3rd parameter can be:
+%% 'false' to not decrement the desired capacity of the group
+%% 'true' to decrement the desired capacity of the group
+%% Config a supplied AWS configuration.
+%% @end
+%% --------------------------------------------------------------------
+detach_instances(InstanceIds, GroupName, false) ->
```
kkuzmin: Let's make the 3rd parameter as a list of options instead of bool value, i.e. ```[{decrement_desired_capacity, true}]```.

ncrothe: Could be worth it, although according to API doc it is the only parameter there currently is. And it's the main use case, i.e. detach instance X from group Y and decrease or don't decrease capacity - could make an inner function accepting a list of params instead. Your thoughts?

kkuzmin:
Please add some test into ```test/erlcloud_as_tests.erl```

src/erlcloud_as.erl
```diff
@@ -367,6 +373,44 @@ resume_processes(GroupName, ScalingProcesses, Config) ->
     end.
 
 
+%% --------------------------------------------------------------------
+%% @doc Detach the given instances from the group using the default configuration
+%% without decrementing the desired capacity of the group.
+%% @end
+%% --------------------------------------------------------------------
+-spec detach_instances(list(string()),string()) -> aws_autoscaling_activity().
```
kkuzmin: what about specs for other functions below with different arity.

src/erlcloud_as.erl
kkuzmin: Actually I don't like the set of functions below. What about replacing them with just one with guard expression, i.e.
```
detach_instances(InstanceIds, GroupName, ShouldDecrementDesiredCapacity) when is_bool(ShouldDecrementDesiredCapacity) ->
    detach_instances(InstanceIds, GroupName, ShouldDecrementDesiredCapacity, erlcloud_aws:default_config())
```

src/erlcloud_as.erl
ncrothe: Added specs and some cleanup.

src/erlcloud_as.erl
ncrothe: Did the guard cleanup, but decided against the param list version - would just clutter up the export list and/or decrease readability of the code (because arity-2 function is already there).

ncrothe:
Added test case as well.

kkuzmin:
Thank you. +1

PR #246, opened 2016-02-02
Cleanup code formatting

sthadka:
* Add newlines to group similar things together
* Remove extra newlines
* Remove trailing spaces
* Remove newlines at the end of files
* Try to wrap lines more than 100 lines long (e.g. avoid this for list comprehensions)
* Try to wrap lines to 80 characters where possible
* Remove unused functions
* Some formatting changes from file to file. Stick to the file's formatting style


sthadka:
The goal of this PR is some basic cleanup.
While going through all the files, I found that the style varies a lot across them. Would there be any interest in having a uniform code style across all the modules?

nalundgaard:
I certainly understand the motivation behind this, but I would argue that it's somewhat misguided on a project of this scope, with the number of individual contributors, without having some explicit style guidelines and a linting tool that people could use to validate their work.

My major concern would be that we have a fork, [alertlogic/erlcloud](https://github.com/alertlogic/erlcloud), which my company has contributed to. It is a pretty long way ahead of the master branch here, and we are slowly working our way towards getting our contributions put into this repo. Merging this type of change seems like it will make that process significantly more confusing for developers to work through, messing up diffs all over the place. And, for that confusion, does this add significant value to the repo? Not much, I'd say. I do like your reformats, but I don't think it's the an opportune time to apply them.

Uniform code style (particularly with typing and returns), on the other hand, I would be very interested in, but commitments to backwards compatibility make doing this in a complete way rather impractical.

README.md
```diff
@@ -10,14 +10,19 @@ Service APIs implemented:
 - Amazon SimpleDB
 - Amazon Mechanical Turk
 - Amazon CloudWatch (MON)
-- Amazon DynamoDB (ddb2) 
+- Amazon DynamoDB (ddb2)
 
-All API functions have been implemented.  Not all functions have been thoroughly tested, so exercise care when integrating this library into production code.  Please send bug reports and patches.
+All API functions have been implemented.  Not all functions have been
```
motobob: it looks like this was done by some tool like a plugin to iDEA and clearly it's broken as this is not what .md files need.

nalundgaard: Actually, a single line break is harmless to markdown. See [here](https://daringfireball.net/projects/markdown/syntax#p):

    A paragraph is simply one or more consecutive lines of text, separated by
    one or more blank lines. (A blank line is any line that looks like a blank
    line — a line containing nothing but spaces or tabs is considered blank.)
    Normal paragraphs should not be indented with spaces or tabs.

    The implication of the “one or more consecutive lines of text” rule is that
    Markdown supports “hard-wrapped” text paragraphs. This differs significantly
    from most other text-to-HTML formatters (including Movable Type’s “Convert
    Line Breaks” option) which translate every line break character in a
    paragraph into a <br /> tag.

Some people are bothered by this, but it's not broken.

smilerlee: Yep, but two or more whitespaces at the end of a line will be converted to a new line.

sthadka:
@nalundgaard a completely valid point. Like you said, maybe the better approach would be for the project to adopt a specific style, have a way to validate that style for incoming PRs and then be open to PRs like this one.

PR #247, opened 2016-02-08
Add support for non-standard DynamoDB and S3 endpoints

efcasado:
This PR makes switching from AWS DynamoDB and S3 to, for instance, DynamoDB local and FakeS3 a bit more seamless.


nalundgaard:
Looks good to me. @motobob any reason not to accept this?

motobob:
LGTM too. sorry, was out of reach of internet for a bit:)

PR #251, opened 2016-02-11
Do not crash process due to S3 errors

efcasado:
The current implementation crashes the process calling `erlcloud_s3:create_bucket/{n}` if the bucket already exists. This sounds a bit inconsistent to me, specially when compared with how the ddb2 module handles the scenario of a user attempting to create a DynamoDB table that already exists `erlcloud_ddb2:create_bucket/{n}`.

Do you think it makes sense to do this standardisation?


nalundgaard:
This is a non-starter that will break existing handlers that expect the current behavior. The disparate interface is frustrating, but if you want to make this change, you need to make a new module (much the way `erlcloud_ddb2` has an evolved interface from `erlcloud_ddb`, which is essentially deprecated.

We probably should have a plan for deprecating interfaces, but IMO modifying an existing one is a recipe for breaking library users in subtle, confusing ways that is incredibly hostile to them.

motobob:
@efcasado yes, it's unfortunate, but this will brake exiting users of this not-really intuitive API. 
this was already brought up several times for s3, sns, etc...

Please `try...catch` it on your application level

waisbrot:
I don't think it would be hostile to make a new major-version release that revamped how things behave. It's a larger effort than one PR and I imagine that a higher-priority goal is to get all the AlertLogic improvements into the main code here, but as you say it's come up in a bunch of places.

motobob:
@waisbrot we are thinking about this...basically merging alertlogic fork and rev it as erlcloud v2.0.0.
however, i know so many case where folks pin to master branch that it will brake them all.

waisbrot:
I have pinned dependencies to master myself, so I don't think I'm being _too_ unfair when I suggest "screw them, every tool warns you not to do that and this is why". :-)

Erlcloud's always been good about having tagged releases, so it's not like anyone was forced to do it the wrong way.

nalundgaard:
I'm not saying categorically that we shouldn't do it, but I do think that it needs to be a larger effort with some prior warning in the form of deprecation warnings, a reference in the README, etc.

motobob:
I'm in the process of preparing a major version bump which will remove all `erlang:error()`. I'll merge your branch to ours and will merge soonish. 
sorry for taking SO long with the PR but that's what happens to breaking changes.

Issue #254, opened 2016-02-18
0.13.0 hex package


Licenser:
It would be great to get the 0.13.0 as a hex package :)

waisbrot:
I said I would handle the Hex packaging for them, but I missed that there was a 0.13.0 release. Sorry.

I just published a 0.13.0, and I removed the listed "meck" dep while I was at it.

Licenser:
awesome! thank you so much! :D

smilerlee:
Thanks @waisbrot , but how about publish the tag 0.12.0 instead?

PR #256, opened 2016-02-18
Support Rebar3 and Rebar2

waisbrot:
Inspired by #254

* rebar.config is written for rebar3 -- Meck is a test-only dependency, and all dependencies draw from Hex.pm
* rebar.config.script is extended to provide backwards-compatibiliy with rebar2.x (code for this copied from [the rebar3 docs](https://www.rebar3.org/docs/from-rebar-2x-to-rebar3#section-maintaining-backwards-compatibility-while-using-hex-packages))
* erlcloud.app.src has a couple extra properties that Hex displays for package info

In my limited testing (running both `rebar3 eunit` and `rebar eunit`) it works.

A couple bits to note:
* lhttpc 1.3.0 has a misconfigured .app file or something, so it doesn't work with rebar2. But there's no lhttpc 1.4.0 published on Hex (and 1.3.0 does work) so there's an unfortunate version mismatch right now
* I didn't fill in the "maintainers" list since all the project collaborators here are listed as private. Hex is happy with an empty list
* Running `rebar3 edoc` (or `rebar doc`) fails for me. Should that work? I find the code pretty readable, and I haven't missed having edocs, but Hex would accept edocs as part of the publishing if they were generated


motobob:
@waisbrot great. give us some time to do some more extended testing here to see whether it's good to go.

couple observations:
1) https://github.com/talko/lhttpc/compare/1.3.0...1.4.0 has quite an impacting change for headers handling. We should be really cautious about downgrading it. In general, lhttpc is end of support and not that many people still support it. Probably we should hex publish our version...
  
2) please update travis.yml file to make the builds via rebar3 too. just add more targets.

motobob:
having rebar.config defaulted to rebar3 type I suggest to update `Makefile` too

waisbrot:
I've updated the Makefile and the travis setup. The Makefile is uglier for it, so let me know if I should try harder to do something nice there.

The existing Hex package for lhttpc 1.3.0 is also from Talko's fork. Maybe @ransomr knows who pushed that version and can ping them to do 1.4.0?

motobob:
@waisbrot great effort. yeah, makefile looks little messy now. let me check if that is truly backward compatible for those who not yet on rebar3. I'll be back in couple days to this.

rebar.config.script
```diff
@@ -8,4 +9,16 @@ case erlang:system_info(otp_release) =< "R15B01" of
...
         end;
     false -> CONFIG
+end,
+case erlang:function_exported(rebar3, main, 1) of
+    true -> % rebar3
+        Config1;
+    false -> % rebar 2.x or older
+        %% Use git-based deps
+        %% profiles
+        [{deps, [{meck, ".*",{git, "https://github.com/eproxus/meck.git", {tag, "0.8.3"}}},
```
ddosia: I would rather prefer to have separate config for `rebar2` and invoke it through the make target with `rebar -C rebar2.config`

Makefile
```diff
@@ -1,6 +1,29 @@
...
+ REBAR_VSN=2
+else
+ REBAR=$(REBAR3)
+ REBAR_VSN=3
+endif
+ifeq ($(REBAR_VSN),2)
+ TARGET_ALL_DEPS = get-deps compile
+ TARGET_EUNIT_DEPS = compile
+ TARGET_CHECK_DEPS = dializer-2
+ TARGET_CHECK_EUNIT_DEPS = dializer-2-eunit
```
nalundgaard: Isn't this misspelled from the target definition [below](https://github.com/erlcloud/erlcloud/pull/256/files#diff-b67911656ef5d18c4ae36cb6741b7965R85)?

Makefile
```diff
@@ -1,6 +1,29 @@
...
+ REBAR=$(REBAR2)
+ REBAR_VSN=2
+else
+ REBAR=$(REBAR3)
+ REBAR_VSN=3
+endif
+ifeq ($(REBAR_VSN),2)
+ TARGET_ALL_DEPS = get-deps compile
+ TARGET_EUNIT_DEPS = compile
+ TARGET_CHECK_DEPS = dializer-2
```
nalundgaard: Looks like this spelling is also disparate from [here](https://github.com/erlcloud/erlcloud/pull/256/files#diff-b67911656ef5d18c4ae36cb6741b7965R85)

Makefile
```diff
@@ -42,23 +65,47 @@ compile:
...
-check: compile
+check: $(TARGET_CHECK_DEPS)
+check-eunit: $(TARGET_CHECK_EUNIT_DEPS)
+
+# Rebar3 knows how to run dialyzer
+dialyzer:
+	@$(REBAR) dialyzer
+
+# Manually run dialyzer when under rebar2
+dialyzer-2: compile
```
motobob: how about having:
```
dialyzer:
ifeq ($(REBAR_VSN),3)
    @$(REBAR) dialyzer
else
    ...old_way
endif
```

nalundgaard: +1 to this—it would make the Makefile more user-friendly and straightforward, I think.

motobob:
@waisbrot sorry for late response. 
please tune the makefile a little as the last change. 
I do need to check if lhttpc downgrade to 1.3.0 will cause big issues but probably we should just build hex 1.4.0 ourselves.

waisbrot:
@motobob thanks for the comments. I've rebased and modified the Makefile to put the conditionals inside each target instead of trying to squish them up top. I think it's an improvement.

@ddosia I'm inclined against that change, because I typically use straight `rebar`/`rebar3`, even when there is a Makefile. Making separate config files would break the ability for rebar2 users to run tests, and the rebar.config.script file solution seems reasonably friendly and easy to discard later on.

motobob:
:+1:

waisbrot:
I rebased this PR, but I'm still hoping someone with access to [the lhttpc package on Hex.pm](https://hex.pm/packages/lhttpc) is around and can push up the 1.4.0 version.

If not, I can hack something together. Either see if the Hex-keepers can give me access to lhttpc or just put up my own lhttpc package.

waisbrot:
Ah, I saw your post in the other thread.

Hex packages have a list of "owners" which are the user accounts allowed to push releases. For erlcloud, the current list is me (it's probably worthwhile for some other people to make Hex accounts at some point so that I'm not the only one with access).

However, I can't see the owner list for packages that I'm not the owner of, so I can't tell who's allowed to push to lhttpc.

I'll email Hex support and see what our options are.

waisbrot:
Talko's lhttpc fork was community-maintained at Hex, so I've been added to the list of accounts that can push to it. I've pushed 1.4.0 of lhttpc to Hex (and can push other versions and give access to others as needed).

I updated the rebar.config in this PR to use lhttpc 1.4.0 from Hex, so rebar2 and rebar3 now both use the same version.

I believe this PR is ready now, unless there are other changes you'd like.

motobob:
@waisbrot great news! you effort on this journey is greatly approached.

Issue #235, opened 2015-12-09
More control over retries, timeouts and backoff for DynamoDB


jonasschmidt:
Right now, the number of retries/attempts is hard-coded to `10` in `erlcloud_ddb_impl`. I think it would be very beneficial to have this configurable.

The same goes for the timeout value for the first request, which is hard-coded to 1s (`#aws_config.timeout` is only used for subsequent requests). There's a comment in the code saying `Shorter timeout on first request.`, but we might actually want to have a shorter timeout than 1s for all requests (including the first).

For the backoff, the option to pass a backoff function as a config setting could also be a useful addition.

ransomr:
The number of attempts and the backoff time can be changed by providing your own `ddb_retry` function. Timeout is not currently easy to configure - I would welcome a PR that would let your provide a timeout function.

jonasschmidt:
Ah ok, that's great. I'll try to prepare a PR for the timeout configuration.

motobob:
this has been addressed in #261

PR #257, opened 2016-02-18
Allow custom ddb timeout fun

derek121:
Motivated by https://github.com/erlcloud/erlcloud/issues/235

### DIFF
--- a/include/erlcloud_aws.hrl
+++ b/include/erlcloud_aws.hrl
@@ -27,6 +27,7 @@
           ddb_streams_scheme="https://"::string(),
           ddb_streams_host="streams.dynamodb.us-east-1.amazonaws.com"::string(),
           ddb_streams_port=80::non_neg_integer(),
+          ddb_timeout=fun erlcloud_ddb_impl:timeout/2::erlcloud_ddb_impl:timeout_fun(),
           kinesis_scheme="https://"::string(),
           kinesis_host="kinesis.us-east-1.amazonaws.com"::string(),
           kinesis_port=80::non_neg_integer(),
--- a/src/erlcloud_ddb_impl.erl
+++ b/src/erlcloud_ddb_impl.erl
@@ -64,13 +64,14 @@
 -export([backoff/1, 
          retry/1, retry/2,
          request_id_from_error/1,
-         error_reason2/1
+         error_reason2/1,
+         timeout/2
         ]).
 
 %% Internal impl api
 -export([request/3]).
 
--export_type([json_return/0, attempt/0, retry_fun/0]).
+-export_type([json_return/0, attempt/0, retry_fun/0, timeout_fun/0]).
 
 -type json_return() :: {ok, jsx:json_term()} | {error, term()}.
 
@@ -109,6 +110,9 @@ backoff(1) -> ok;
 backoff(Attempt) ->
     timer:sleep(random:uniform((1 bsl (Attempt - 1)) * 100)).
 
+-type timeout_fun() :: fun((pos_integer(), #aws_config{}) -> non_neg_integer()).
+
+-spec timeout(pos_integer(), #aws_config{}) -> non_neg_integer().
 %% HTTPC timeout for a request
 timeout(1, _) ->
     %% Shorter timeout on first request. This is to avoid long (5s) failover when first DDB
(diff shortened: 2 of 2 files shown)


motobob:
lgtm... @nalundgaard your call.

nalundgaard:
Looks great to me. My one comment is that I would have put the `timeout_fun` entry in the `#aws_config{}` record right below the `retry_fun` entry. Otherwise this looks fully backwards compatible and implements the desired customization. Thanks @derek121.

smilerlee:
Looking deeply into `erlcloud_ddb_impl`, the timeout of first attempt sticks to 1s, and continuous retries timeout are `Config#aws_config.timeout`. Seems it should be safe to delete this behavior and no one would and should have relied on the fast timeout-retry magic.

nalundgaard:
@smilerlee interesting. The original question in #235 was to be able to specify a smaller than 1s timeout for the initial call. Are you sure that is not a desired behavior when the main timeout is 5s? It looks like @ransomr wrote that, so it would be good to know his motivation before removing the default timeout.

If it needs to stay, it would be a good compromise to make the default look like this:

```erlang
-spec timeout(pos_integer(), #aws_config{}) -> non_neg_integer().
%% HTTPC timeout for a request
timeout(1, Config) ->
    %% If the configured timeout is > 1s, use a shorter 1s timeout on first request.
    %% This is to avoid long (5s) failover when first DDB endpoint doesn't respond
    min(1000, Config#aws_config.timeout);
timeout(_, Config) ->
    %% Use the configured timeout on subsequent requsets - The default of 5s results
    %% in less timeouts when system is under heavy load
    Config#aws_config.timeout.
```

Even if we can remove the custom DDB timeout logic, it might be worthwhile to still have this feature.

smilerlee:
@nalundgaard good point.

IMHO, a complex network request timeout strategy depends on attempt times is a kind of meaningless, and the default 10s (not 5s) may be too long. It would be really great if @ransomr could tell us the initial motivation :)

Inspired by you @nalundgaard and I just came up with a better solution and I believe it's an awesome one, as long as we have to keep all these legacy behaviors (1s timeout of `cloudtrail` and `s3`, and short timeout of `ddb` initial call).
We can change the default value of `#aws_config.timeout` to an atom named `default`:
* If users specify the timeout explicitly (`pos_integer()` or `infinity`, i.e. `timeout()`), we use this user-specified timeout on every request/retry;
* Otherwise (`default`), we use the original timeout strategies.

The ability to specify the timeout in the same way for every service, and no backwards compatibility broken. What do you think?
CC: @motobob

smilerlee:
I suppose `undefined`, rather than `default`, makes better sense.

ransomr:
Initial motivation was based on observed behaviors of timeouts within a single AWS region. Typical DDB calls were on the order of 10ms. When a call did timeout the subsequent call almost always succeeded quickly. So in those network conditions a 1s timeout was usually sufficient, and was much preferable to 5s. However with some operations (large scan or query) the 1s was sometimes too small even within a region. Because I anticipated that people would want to heavily customize the the timeout behavior, I just made it possible to override the function as I thought that would be better than having many separate values you could tune. For defaults I used what we used for our service, since that's what I was able to verify could work well at least in one set of conditions.

smilerlee:
Thanks @ransomr. It's true than in some cases, the initial call is much longer than subsequent calls. As you said, may be longer than 1s, not only for large scan or query, but also for simple small put_item call (in some cases, indeed). If it comes to a conditional write, this might cause unexpected problems.
What's the problem with "if the server not responds, I just wait" strategy? What's the necessity to retry in a hurry before the server responds? As I see it, users can specify a sufficient timeout according to their network conditions and business logics, that's enough.

jannekai:
One use case why I had to fork erlcloud and add configuration to specify longer than 1 second initial timeout was running tests on local machine against real dynamodb database in us-east AWS environment. The 1s non-overridable configuration caused tests to randomly fail due to timeouts.

motobob:
@derek121 may I ask you to implement the default timeout as @nalundgaard suggested with 
```erlang
min(1000, Config#aws_config.timeout);
```
after that that's ready to go from both backward compatibility standpoint and new functionality to allow retry timeout customisation. @smilerlee agree?

smilerlee:
@motobob don't agree with this, it's powerful but too complex to config. APIs should be easy to use. Adding `min(1000, Config#aws_config.timeout);` also increases this complexity. I will make a new PR later for discuss.

motobob:
@derek121 alternative implementation was done via https://github.com/erlcloud/erlcloud/pull/261.
Anyway thank you for contribution. 
closing this.

PR #258, opened 2016-03-01
Fix requests made in the China region

lrascao:
AWS China endpoints all have a .cn suffix, account
for that when parsing endpoints to extract the region.

### DIFF
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -107,7 +107,17 @@ aws_request2_no_update(Method, Protocol, Host, Port, Path, Params, #aws_config{}
 
 aws_region_from_host(Host) ->
     case string:tokens(Host, ".") of
-        [_, Value, _, _] ->
+        %% the aws endpoint can vary depending on the region
+        %% we need to account for that:
+        %%  us-west-2: s3.us-west-2.amazonaws.com
+        %%  cn-north-1 (AWS China): s3.cn-north-1.amazonaws.com.cn
+        %% it's assumed that the first element is the aws service (s3, ec2, etc),
+        %% the second is the region identifier, the rest is ignored
+        %% the exception (of course) is the dynamodb streams which follows a different
+        %% format
+        ["streams", "dynamodb", Value | _Rest] ->
+            Value;
+        [_, Value, _, _ | _Rest] ->
             Value;
         _ ->
             "us-east-1"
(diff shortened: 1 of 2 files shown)


motobob:
lgtm. @kkuzmin please double check

smilerlee:
It would be great if the region extraction could also be fixed in `erlcloud_ddb_streams.erl`.

kkuzmin:
+1

motobob:
@lrascao  i would agree with @smilerlee. It would be really nice if you could update 
https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_ddb_streams.erl#L899-L904 too
to just simply
```erlang
Region = erlcloud_aws:aws_region_from_host(Config#aws_config.ddb_streams_host)
```
this remove useless copy-paste.

lrascao:
no problem, PR updated

smilerlee:
Sorry @lrascao, but there're two bugs in your patch.
* As to host `s3.amazonaws.com`, your patch would interpret the region to `amazonaws`, which is wrong;
* It not as simple as delete the copy-paste code, since the host of DynamoDB Streams is differ from of other services: `streams.dynamodb.us-east-1.amazonaws.com`, it's the third part, not the second one.

Maybe the match patterns should be:
```
["streams", "dynamodb", Value | _Rest] -> Value;
[_, Value, _, _ | _Rest] -> Value;
_ -> "us-east-1"
```

lrascao:
> As to host s3.amazonaws.com, your patch would interpret the region to amazonaws, which is wrong;
If i'm not mistaken s3 requests completely bypasses this credentials business, which probably explains why nobody complained until now, the old code has the same issue.

> It not as simple as delete the copy-paste code, since the host of DynamoDB Streams is differ from of other services: streams.dynamodb.us-east-1.amazonaws.com, it's the third part, not the second one.
That seems reasonable, i'll update the PR

lrascao:
My point is that there is no bug since S3 requests never call this method (which explains why nobody complained).
Related to this, there is another issue (which is best reported in a different ticket) that is S3 working in AWS China due to it requiring the V4 signature

smilerlee:
Not only `s3`, you may take a look at [`erlcloud_aws.hrl`](https://github.com/erlcloud/erlcloud/blob/master/include/erlcloud_aws.hrl).
IMHO, we'd better not change the original behavior, that's why I recommend the pattern `[_, Value, _, _ | _Rest]`.

lrascao:
updated as per @smilerlee's suggestion

PR #259, opened 2016-03-04
AWS WAF API support

ptrakhtman:


src/erlcloud_waf.erl
```diff
@@ -0,0 +1,1152 @@
...
+
+-spec create_byte_match_set/3 ::
+        (ChangeToken :: string() | binary(),
+         Name :: string() | binary(),
+         Config :: aws_config()) ->
+        waf_return_val().
+create_byte_match_set(ChangeToken, Name, Config) ->
+    Json = [{<<"ChangeToken">>, to_binary(ChangeToken)},
+            {<<"Name">>, to_binary(Name)}],
+    waf_request(Config, "AWSWAF_20150824.CreateByteMatchSet", Json).
```
motobob: suggest hiding this API version as define and putting inside `waf_request()`. if new version of API will be released we will use ec2 module approach

motobob: like
https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_ec2.erl#L164
and 
https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_ec2.erl#L2989-L2992
there its passed as options but nothing prevents concatenating two strings atm.

i wish AWS was consistent in versioning their APIs. May be we can ask for that for WAF as it's stiff pretty new ?;)

src/erlcloud_waf.erl
```diff
@@ -0,0 +1,1152 @@
...
+
+%%%------------------------------------------------------------------------------
+%%% ListByteMatchSets
+%%
+%% http://docs.aws.amazon.com/waf/latest/APIReference/API_ListByteMatchSets.html
+%%%------------------------------------------------------------------------------
+-spec list_byte_match_sets/1 ::
+       (Limit :: pos_integer()) ->
+       waf_return_val().
+list_byte_match_sets(Limit) when is_number(Limit) ->
```
motobob: i understand that AWS WAF API claims Limit as required=YES (mindblown) for list APIs but I would suggest the following api with options and we did for inspector(https://github.com/alertlogic/erlcloud/blob/master/src/erlcloud_inspector.erl#L538). 
also I would not be surprised if Limit is actually optional and there is a default value
```erlang
list_byte_match_sets() ->
    list_byte_match_sets([], default_config()).
list_byte_match_sets(Limit) when is_number(Limit) ->
   list_byte_match_sets([{<<"Limit">>, Limit}], default_config());
list_byte_match_sets(Options) when is_list(Options) ->
   list_byte_match_sets(Options, default_config());
list_byte_match_sets(Config) when is_record(Config, aws_config) ->
   list_byte_match_sets([], Config).
list_byte_match_sets(Options, Config) when is_list(Option), is_record(Config, aws_config) ->
   list_byte_match_sets(Options, Config).
```

do you find it reasonable @ptrakhtman ?

motobob: also API doc says 100 is the max so i would suggest guarding for that.

ptrakhtman: @motobob The Limit parameter is actually required and there is no default. I opted out to more explicit because of that... I will add a guard though.

motobob: @nalundgaard @smilerlee what is your input of preferred API for new modules?

motobob:
:+1: 
@nalundgaard @smilerlee given this is a completely new module any feedback?

smilerlee:
After I get time to read the official APIs tomorrow :)

src/erlcloud_waf.erl
```diff
@@ -0,0 +1,1191 @@
...
+               [] -> <<"{}">>;
+               _ -> jsx:encode(Body)
+           end,
+    Headers = headers(Config, Operation, Payload),
+    Request = #aws_request{service = waf,
+                           uri = uri(Config),
+                           method = post,
+                           request_headers = Headers,
+                           request_body = Payload},
+    io:format("~p~n", [Request]),
```
nalundgaard: Does this need to be here? seems inappropriate.

src/erlcloud_waf.erl
```diff
@@ -0,0 +1,1191 @@
...
+%%% Internal Functions
+%%%------------------------------------------------------------------------------
+-spec transform_to_proplist/1 ::
+        (waf_byte_match_set_update() |
+         waf_ip_set_update() |
+         waf_rule_update() |
+         waf_size_constraint_update() |
+         waf_sql_injection_match_set_update() |
+         waf_web_acl_update()) ->
+       proplists:proplists().
```
nalundgaard: `proplists:proplist()`, no?

src/erlcloud_waf.erl
```diff
@@ -0,0 +1,1191 @@
...
+    
+
+-spec record_to_proplist/1 ::
+        (waf_rule_predicate() |
+         waf_ip_set_descriptor() |
+         waf_byte_match_tuple() |
+         waf_size_constraint() |
+         waf_sql_injection_match_tuple() |
+         waf_activated_rule()) ->
+        proplists:proplists().
```
nalundgaard: same here, `proplists:proplist()` I think

nalundgaard:
Kind of a lot to unpack here, and I've just had time to skim a few times so far. It looks good, albeit I'm not hugely thrilled about the record-y, erlang-y input with the very untransformed, AWS JSON-y output. Still, that can be amended in the future with optional output formats, if someone wants to do it, so I think this is fine for the first go.

src/erlcloud_waf.erl
```diff
@@ -0,0 +1,1191 @@
+-module(erlcloud_waf).
+-author('pavel@alertlogic.com').
+
+-include_lib("erlcloud_aws.hrl").
+-include_lib("include/erlcloud_waf.hrl").
```
smilerlee: Should file inclusion be one of `-include("x.hrl").`, `-include("../include/x.hrl").` and `-include_lib("erlcloud/include/x.hrl").`?

src/erlcloud_waf.erl
```diff
@@ -0,0 +1,1191 @@
...
+
+%%% Library initialization.
+-export([configure/2, configure/3, configure/4, new/2, new/3, new/4, waf_result_fun/1]).
```
smilerlee: There is no need to export `waf_result_fun/1`,

src/erlcloud_waf.erl
```diff
@@ -0,0 +1,1191 @@
...
+-spec waf_result_fun/1 ::
+          (Request :: aws_request()) -> aws_request().
+waf_result_fun(#aws_request{response_type = ok} = Request) ->
+    Request;
+waf_result_fun(#aws_request{response_type = error,
+                                  error_type = aws,
+                                  response_status = Status} = Request) when Status >= 500 ->
+    Request#aws_request{should_retry = true};
+waf_result_fun(#aws_request{response_type = error, error_type = aws} = Request) ->
+    Request#aws_request{should_retry = false}.
```
smilerlee: It'll be better to place this function after `waf_request_no_update/3` where is its first usage.

ptrakhtman:
@nalundgaard and @smilerlee your comments are addressed.

smilerlee:
The params order in official documentation is alphabetical.
I suppose the function arguments order should be logical, it's easy to understand.
I'll list several comments below, which are only my personal opinions. Welcome to discuss.

src/erlcloud_waf.erl
```diff
@@ -0,0 +1,1188 @@
...
+%%
+%% http://docs.aws.amazon.com/waf/latest/APIReference/API_CreateRule.html
+%%%------------------------------------------------------------------------------
+-spec create_rule/3 ::
+        (ChangeToken :: string() | binary(),
+         MetricName :: string() | binary(),
+         Name :: string() | binary()) ->
+        waf_return_val().
+create_rule(ChangeToken, MetricName, Name) ->
+    create_rule(ChangeToken, MetricName, Name, default_config()).
```
smilerlee: `create_rule(ChangeToken, Name, MetricName)`

src/erlcloud_waf.erl
```diff
@@ -0,0 +1,1188 @@
...
+%% http://docs.aws.amazon.com/waf/latest/APIReference/API_CreateWebACL.html
+%%%------------------------------------------------------------------------------
+-spec create_web_acl/4 ::
+        (ChangeToken :: string() | binary(),
+         DefaultActionType :: block | allow | count,
+         MetricName :: string() | binary(),
+         Name :: string() | binary()) ->
+        waf_return_val().
+create_web_acl(ChangeToken, DefaultActionType, MetricName, Name) ->
+    create_web_acl(ChangeToken, DefaultActionType, MetricName, Name, default_config()).
```
smilerlee: `create_web_acl(ChangeToken, Name, MetricName, DefaultActionType)`

src/erlcloud_waf.erl
```diff
@@ -0,0 +1,1188 @@
...
+    waf_request(Config, "GetRule", Json).
+
+
+%%%------------------------------------------------------------------------------
+%%% GetSampledRequests
+%%
+%% http://docs.aws.amazon.com/waf/latest/APIReference/API_GetSampledRequests.html
+%%%------------------------------------------------------------------------------
+-spec get_sampled_requests/5 ::
+        (MaxItems :: pos_integer(),
```
smilerlee: `1..100`

src/erlcloud_waf.erl
```diff
@@ -0,0 +1,1188 @@
...
+         RuleId :: string() | binary(),
+         StartTime :: pos_integer(),
+         EndTime :: pos_integer(),
+         WebAclId :: string() | binary()) ->
+        waf_return_val().
+get_sampled_requests(MaxItems, RuleId, StartTime, EndTime, WebAclId) ->
+    get_sampled_requests(MaxItems, RuleId, StartTime, EndTime, WebAclId, default_config()).
```
smilerlee: `get_sampled_requests(WebAclId, RuleId, StartTime, EndTime, MaxItems)`

http://docs.aws.amazon.com/waf/latest/APIReference/API_GetSampledRequests.html#AWS%20WAF-GetSampledRequests-request-MaxItems
MaxItems
The number of requests that you want AWS WAF to return from among the first 5,000 requests that your AWS resource received **during the time range**.

Therefore, `MaxItems` is after `TimeWindow`.

src/erlcloud_waf.erl
```diff
@@ -0,0 +1,1188 @@
...
+%% http://docs.aws.amazon.com/waf/latest/APIReference/API_ListByteMatchSets.html
+%%%------------------------------------------------------------------------------
+-spec list_byte_match_sets/0 ::
+        () ->
+       waf_return_val().
+list_byte_match_sets() ->
+    list_byte_match_sets(?LIMIT_MAX, default_config()).
+
+-spec list_byte_match_sets/1 ::
+       (Limit :: pos_integer()) ->
```
smilerlee: `1..100`

src/erlcloud_waf.erl
```diff
@@ -0,0 +1,1188 @@
...
+%%
+%% http://docs.aws.amazon.com/waf/latest/APIReference/API_UpdateWebACL.html
+%%%------------------------------------------------------------------------------
+-spec update_web_acl_default_action/3 ::
+       (ChangeToken :: string() | binary(),
+        WebAclId :: string() | binary(),
+        DefaultAction:: waf_acl_action_type()) ->
+       waf_return_val().
+update_web_acl_default_action(ChangeToken, WebAclId, DefaultAction) ->
+    update_web_acl_default_action(ChangeToken, WebAclId, DefaultAction, default_config()).
```
smilerlee: Updating default action only may be inappropriate since rules and default action can be updated at the same time.

The params combination is a little complex :)
`update_web_acl(ChangeToken, WebAclId, Updates)`
`update_web_acl(ChangeToken, WebAclId, Updates, Config)`
`update_web_acl(ChangeToken, WebAclId, DefaultAction)`
`update_web_acl(ChangeToken, WebAclId, DefaultAction, Config)`
`update_web_acl(ChangeToken, WebAclId, Updates, DefaultAction)`
`update_web_acl(ChangeToken, WebAclId, Updates, DefaultAction, Config)`

`DefaultAction` is after `Updates`

src/erlcloud_waf.erl
smilerlee: The two args `StartTime` and `EndTime` can be compress into a tuple `TimeWindow = {StartTime, EndTime}`.

smilerlee:
This input style is ok as long as Amazon won't add more optional params to the APIs. Otherwise, it'll be a disaster to keep backwards compatibilities.
Need we consider the ability to add the second-to-last `Opts` argument in future? (just like `ddb2`)

src/erlcloud_waf.erl
smilerlee: `DefaultAction` can be renamed to `DefaultActionType` like [this](https://github.com/erlcloud/erlcloud/pull/259/files#diff-bb24148281aa8e81f08cbb145df91fc0R240)

smilerlee:
@motobob @nalundgaard any idea?

nalundgaard:
I agree that it would be a good idea to think about supporting Options, either to add erlcloud-specific ones in the future (primarily output options), or to add support for optional parameters added by AWS to the waf API.

The main problem that I see is that a bunch of the inputs support strings, which are more, let's say, difficult to rationally differentiate from an Options proplist. Seems that the natural thing to do would be to not support string inputs.

nalundgaard:
Maybe I am wrong about that, though? It seems like the options can be added at the final arity of each function, where it only needs to be differentiated from Config, i.e.

```erlang
waf_fun(..., #aws_config{} = Config) ->
    waf_fun(..., [], #aws_config{});
waf_fun(... Options) ->
    waf_fun(..., Options, default_config()).

waf_fun(..., Options, Config) -> 
   % ...
```

In that sense, I suppose it's OK as is?

smilerlee:
@nalundgaard you mean `Options` and `Config` are both optional in either way? interesting.
Or, we can just use the `ddb2` approach:
```erlang
waf_fun(...)
waf_fun(..., Options)
waf_fun(..., Options, Config)
```
Seems great, but it may be a little trivial to use some APIs:
```erlang
list_byte_match_sets([{limit, 50}]) %% and also other list* functions
update_web_acl(ChangeToken, WebACLId, [{updates, ...}])
update_web_acl(ChangeToken, WebACLId, [{default_action, ...}])
update_web_acl(ChangeToken, WebACLId, [{updates, ...}, {default_action, ...}])
```
Does it?

smilerlee:
Additionally, the api below is meaningless, as it doesn't update any thing!
```erlang
update_web_acl(ChangeToken, WebACLId)
```

nalundgaard:
I think it still works? Anything that doesn't have to be given explicitly should go in Options, and the Options param (like the Config param) should not be required, as in your `waf_fun` above with the `(...)` (no Options/Config). 

And yes, I see what you're saying about the list functions, IMO the optional stuff should be put in an Options proplist.

nalundgaard:
Regarding `update_web_acl`, ugh. I see what you mean by the Updates/DefaultAction being both optional, with the intent being that you provide either one or both. Technically, in this case the current implementation is artificially limited, since it doesn't support providing a `Updates` without *also* providing `DefaultAction`. It would be nicer to see these as Options. It seems in this case there is no use for a `update_web_acl(ChangeToken, WebACLId)`, and it should just start out as `update_web_acl(ChangeToken, WebACLId, Options)`.

nalundgaard:
In cases like that, where you have API functions that let you do any of a set of things, but no thing that is explicitly, always required, I don't think we should implement that logic in erlcloud. Making Optional parameters go in an Options list will fix the above issue, and if AWS's API enforces that you do at least one of the optional things, then I propose that such business logic should reside with them.

smilerlee:
I do agree with the `Options` list idea, I did.
Let's take a further think, required parameter might become optional later, right? (this did happen in DynamoDB)
Thus, I prefer to make `Updates` required in `update_web_acl`, to keep consist with other `list_*` APIs.
The reason I insist this:
* consistency
* Amazon might add some more attributes to, e.g., `Rule` and make `Updates` in `update_rule` optional, allowing users to only update the added attributes

This is only one point above, I'll take some more detailed think later.

nalundgaard:
Yes, I didn't think about that... Since Updates can be an empty list, it makes sense to make it a first-class argument. I am not sure that really makes the interface any better, but I am ok with that.

Regarding the potential for AWS to demote required parameters to optional, I think the main concern there for future updates would be differentiating string parameters from Options. The main solution I see to that is to just not support string inputs.

include/erlcloud_waf.hrl
```diff
@@ -0,0 +1,126 @@
...
+-type waf_text_transformation() :: none | compress_white_space | html_entity_decode | lowercase | cmd_line | url_decode.
+-type waf_positional_constraint() :: exactly | starts_with | ends_with | contains | contains_word.
+-type waf_comparison_operator() :: eq | ne | le | lt | ge | gt.
+-type waf_predicate_type() :: ip_match | byte_match | sql_injection_match | size_constraint.
+-type waf_acl_action_type() :: block | allow | count.
+
+-record(waf_field_to_match, {
+    type :: uri | 'query_string' | header | method | body,
+    data :: string()
+}).
```
smilerlee: It might be pretty clear to use simple tuples instead of complex records.
Just like we did in `erlcloud_ddb2`:
https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_ddb2.erl#L2051-L2055

smilerlee:
I'm also ok with any form, as long it's easy to understand, to use, and to extend. Additionally, it couldn't be better if it's consistent among all WAF APIs.

Let's wait for @motobob's options :)

>The main solution I see to that is to just not support string inputs.

Agree.

src/erlcloud_waf.erl
motobob: >I suppose the function arguments order should be logical

tend to agree but the logical aspect here is to be on par with AWS API defined order which happens to be alphabetical, no?

src/erlcloud_waf.erl
motobob: true. `TimeWindow = {StartTime, EndTime}` even in AWS api

include/erlcloud_waf.hrl
motobob: here records are definitely over used but in general, records are preferable if four or more  elements in tuple. obviously, this falls more under the code style practices and not functional requirements...

motobob:
so let me try to summarize:
- records input is fine as long there is a way to pass `Options` and newly created optional fields. But this is AWS API version change either, no?
- > "The main solution I see to that is to just not support string inputs."
I see why this came up but I do not see this feasible. majority of API in string based already and switching to binaries will not help greatly.
- I agree that current output is not transformed back to records(and should not) and thus it looks strange, but I'm very against having fully open `Options` proplists input. That is hard to maintain, document and consume.

nalundgaard:
@motobob AFAIK AWS is in the habit of adding backwards compatible features to their APIs without specifically incrementing the API version. New optional parameters would likely fall in this category.

nalundgaard:
I also am not proposing "open-ended" options, but options inputs specific to any relevant API that has optional parameters. IMO this should work as much like erlcloud_ddb2 as is practical, I think it's an effective pattern.

include/erlcloud_waf.hrl
smilerlee: That's true. Too many nested records result to hard to write client code.

src/erlcloud_waf.erl
smilerlee: Since the form order in the [AWS Console](https://console.aws.amazon.com/waf/home?region=us-east-1#/rules/new) is `Name` then `CloudWatch metric name`, I believe it's enough to suggest the preferable order.

smilerlee:
> AWS is in the habit of adding backwards compatible features to their APIs without specifically incrementing the API version.

This fact can be seen from their [document history](http://docs.aws.amazon.com/waf/latest/developerguide/doc-history.html).

Since there're few optional params here thus I agree with not adding `Options` list. However, we need reserve the ability to add the `Options` argument in the future.

ptrakhtman:
@smilerlee @motobob and @nalundgaard addressed your comments... Please review.

src/erlcloud_waf.erl
```diff
@@ -0,0 +1,1326 @@
...
+        ]).
+
+
+-define(API_VERSION, "20150824").
+-define(LIMIT_MAX, 100).
+
+%%%------------------------------------------------------------------------------
+%%% Shared types
+%%%------------------------------------------------------------------------------
+%% -type waf_opts() :: list().
```
nalundgaard: does this need to be here?

nalundgaard:
I think my comments have been addressed, 👍

src/erlcloud_waf.erl
```diff
@@ -501,24 +549,24 @@ get_rule(RuleId, Config) ->
...
+        (RuleId :: string() | binary(),
          StartTime :: pos_integer(),
          EndTime :: pos_integer(),
+         MaxItems :: 1..100,
          WebAclId :: string() | binary()) ->
         waf_return_val().
-get_sampled_requests(MaxItems, RuleId, StartTime, EndTime, WebAclId) ->
-    get_sampled_requests(MaxItems, RuleId, StartTime, EndTime, WebAclId, default_config()).
+get_sampled_requests(RuleId, StartTime, EndTime, MaxItems, WebAclId) ->
+    get_sampled_requests(RuleId, StartTime, EndTime, MaxItems, WebAclId, default_config()).
```
smilerlee: Seems the `WebAclId` should come first.

src/erlcloud_waf.erl
```diff
@@ -0,0 +1,1324 @@
...
+%%%------------------------------------------------------------------------------
+-spec get_sampled_requests/5 ::
+        (RuleId :: string() | binary(),
+         StartTime :: pos_integer(),
+         EndTime :: pos_integer(),
+         MaxItems :: 1..100,
+         WebAclId :: string() | binary()) ->
+        waf_return_val().
+get_sampled_requests(RuleId, StartTime, EndTime, MaxItems, WebAclId) ->
+    get_sampled_requests(RuleId, StartTime, EndTime, MaxItems, WebAclId, default_config()).
```
smilerlee: Seems the `WebAclId` should come first.

smilerlee:
@nalundgaard should we make `limit` and `next_marker` in the `list_*` functions an options list?

src/erlcloud_waf.erl
```diff
@@ -0,0 +1,1324 @@
...
+
+-spec get_sampled_requests/6 ::
+        (RuleId :: string() | binary(),
+         StartTime :: pos_integer(),
+         EndTime :: pos_integer(),
+         MaxItems :: 1..100,
+         WebAclId :: string() | binary(),
+         Config :: aws_config()) ->
+        waf_return_val().
+get_sampled_requests(RuleId, StartTime, EndTime, MaxItems, WebAclId, Config) when MaxItems > 0, MaxItems =< 100->
```
smilerlee: Would it be better if we add `is_integer(MaxItems)` here?

nalundgaard: probably would be good, if indeed AWS doesn't like non-integer numbers.

nalundgaard:
On the list_* functions, I do agree that it would be more natural to treat those as options. `next_marker` is explicitly optional, and it seems reasonable, to some extent, to make `limit` "optional" by defaulting it to the maximum value, as is currently implemented.

I do think that would make for more readable calls, e.g. `list_byte_match_sets([{limit, 50},{next_marker, <<"marker">>}])`. it looks like that could be added in the future if desired without breaking the existing interface, though, so I don't know if it *needs* to be done now. @ptrakhtman what do you think?

smilerlee:
@nalundgaard if it could be, I would prefer to use the below combination instead of options list, **for now**.

```erlang
update_web_acl(ChangeToken, WebACLId, Updates)
update_web_acl(ChangeToken, WebACLId, Updates, Config)
update_web_acl(ChangeToken, WebACLId, DefaultAction)
update_web_acl(ChangeToken, WebACLId, DefaultAction, Config)
update_web_acl(ChangeToken, WebACLId, Updates, DefaultAction)
update_web_acl(ChangeToken, WebACLId, Updates, DefaultAction, Config)
```
The 3rd argument is either a list of `waf_web_acl_update` record or an atom, leaves us the ability to add the options list fashion in future:
```erlang
update_web_acl(ChangeToken, WebACLId, Options)
update_web_acl(ChangeToken, WebACLId, Options, Config)
```
What do you think?

smilerlee:
In the best case, we don't even need the options list at all, no matter for now or in future.

nalundgaard:
I guess I don't really care at this point, as long as it's something that's functional and reasonable—I'm starting to feel pretty bad for @ptrakhtman! I am not terribly fond of adding a whole bunch of different ways to do the same thing, and I like the predictability of putting optional inputs in the `Options` list—so you know where to expect them. Positional argument soup sounds like the kind of thing I would only do if I were trying to maintain backwards compatibility, not so much something I would plan from the beginning. But that's just my perspective. As you say, both ways can be extended, so I wouldn't insist on my preference if you are in disagreement. @motobob what do you think?

smilerlee:
I also feed terrible bad for @ptrakhtman, thank you @ptrakhtman for your effort, sincerely.
My only intention to use positional arguments instead of options list is keep a consistent fashion among functions :) After all, current implementation is acceptable as I see it.

ptrakhtman:
@smilerlee @motobob @nalundgaard I believe I've addressed all your comments and concerns. Please review.

motobob:
@nalundgaard @smilerlee  I'm :+1: :shipit: for this PR in the current form. 
@ptrakhtman made quite a number of improvement to original implementation and asking to polish it further within this PR will be too much. "better" is worse then "good" here.
I think we should get some more feedback after people start using it and enhance iteratively.

nalundgaard:
I agree. Let's take it in the current form. If there is a desire to go with @smilerlee's suggested change, we can use also pattern matching to add support for that style too. :shipit:

smilerlee:
Agree. 👍

PR #261, opened 2016-03-08
Add a universal way to config timeout

smilerlee:
A compact and universal way to specify http request timeout.

Backwards compatibility won't be broken as long as users didn't specify the timeout explicitly.
Nevertheless, in case of specified, this change is what they actually needed.

Previous complete discusses are in #257.

### DIFF
--- a/include/erlcloud_aws.hrl
+++ b/include/erlcloud_aws.hrl
@@ -38,7 +38,11 @@
           access_key_id::string()|undefined|false,
           secret_access_key::string()|undefined|false,
           security_token=undefined::string()|undefined,
-          timeout=10000::timeout(),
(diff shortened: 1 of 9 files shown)


motobob:
thanks @smilerlee this looks simpler and more generic. closing both.

smilerlee:
Thanks @motobob. I didn't expect this PR will be merged so quickly without further discuss.

motobob:
sorry, does it really require any further discussion? :)

smilerlee:
Has been concerned, since there are no feedbacks at all :)

nalundgaard:
After the fact, but I endorse this change. :+1:

Issue #260, opened 2016-03-05
Update meck dependency


webzepter:
It seems [meck](https://github.com/eproxus/meck/releases) dependency is outdated. Please, update it! :pray:

motobob:
it used to be meck/master and 0.8.4 specifically and afterward it was downgraded to 0.8.3 by https://github.com/erlcloud/erlcloud/pull/228.

is the request to update to 0.8.4?

webzepter:
@motobob yes

motobob:
please explain why this is necessary. 
as it's minor version difference and I also do not see major changes https://github.com/eproxus/meck/compare/0.8.3...0.8.4

if you really need 0.8.4 in your project because of other components i would encourage pinning meck to 0.8.4 above erlcloud in rebar.config file.

webzepter:
Our project is covered by tests written with [Espec](https://github.com/antonmi/espec), but latest versions of `espec` [depend](https://github.com/antonmi/espec/blob/master/mix.exs#L25) on `meck` 0.8.4

smilerlee:
thanks @webzepter and @motobob

PR #264, opened 2016-03-15
Update rebar.config

motobob:
bump meck to address https://github.com/erlcloud/erlcloud/issues/260


motobob:
@smilerlee @nalundgaard any objections?

smilerlee:
Reasonable. thanks @motobob

nalundgaard:
Looks good to me, but I haven't messed with it. Don't care as long as it works.

PR #268, opened 2016-03-29
 Added creation_date to result returned by describe_images.

waynerasmuss:
Had to use NEW_API_VERSION to get the field in the result

### DIFF
--- a/src/erlcloud_ec2.erl
+++ b/src/erlcloud_ec2.erl
@@ -1130,33 +1130,78 @@ extract_permissions([_|Nodes], Accum) ->
     extract_permissions(Nodes, Accum).
 
 %%
+%% Returns a list of all images Executable by the current user.
+%% Responses in the form:
+%% [[
+%%    image_id: 'ami-001af160',
+%%    image_location: '<accountid>/<image-name>',
+%%    image_state: 'available',
+%%    image_owner_id: '<accountid>',
+%%    is_public: false, architecture: 'x86_64',
+%%    image_type: 'machine',
+%%    kernel_id: [],
+%%    ramdisk_id: [],
+%%    image_owner_alias: :none,
+%%    name: '<image-name>',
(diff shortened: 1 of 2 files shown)


src/erlcloud_ec2.erl
```diff
@@ -1188,6 +1188,7 @@ extract_image(Node) ->
      {description, get_text("description", Node)},
      {root_device_type, get_text("rootDeviceType", Node)},
      {root_device_name, get_text("rootDeviceName", Node)},
+     {creation_date, erlcloud_xml:get_time("creationDate", Node)},
```
motobob: sounds simple but I cannot find it in AWS API docs http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeImages.html.
could you please provide an example response and would be even better if you put into the tests

msayler: I think it's from http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Image.html 

> creationDate
>
>    The date and time the image was created.
>   Type: String
>   Required: No

waynerasmuss: I had previously checked for an existing describe_instances test. There does not appear to be one.  I can probably add one, but I'm very new to erlang, so it will be a bit of a job for me. I did test the make sure it works for my code that is using it. For an example, would you like to me to add a whole image term as a comment like some of the other functions have? An image term is pretty lengthy.

motobob: you're right. this fork does not have that test at all.
here is an example - https://github.com/alertlogic/erlcloud/blob/master/test/erlcloud_ec2_tests.erl#L430-L475

waynerasmuss:
I have added a unit test now thanks to motobob's direction.

motobob:
@waynerasmuss thank you for contribution.

PR #269, opened 2016-03-31
fix global default_timeout

motobob:
Problem:
follow up on https://github.com/alertlogic/erlcloud/issues/99 and https://github.com/alertlogic/erlcloud/pull/100

Solution:
remove define. update how the library uses timeout.

@smilerlee @nalundgaard please review and merge before it hits other consumers of the library

### DIFF
--- a/src/erlcloud_cloudtrail.erl
+++ b/src/erlcloud_cloudtrail.erl
@@ -163,7 +163,7 @@ request_impl(Method, _Protocol, _Host, _Port, _Path, Operation, Params, Body, #a
                 erlcloud_httpc:request(
                      url(Config), Method, 
                      [{<<"content-type">>, <<"application/x-amz-json-1.1">>} | Headers],
-                     Body, erlcloud_aws:get_timeout(Config), Config)) of
+                     Body, timeout(Config), Config)) of
        {ok, {_RespHeader, RespBody}} ->
             case Config#aws_config.cloudtrail_raw_result of
                 true -> {ok, RespBody};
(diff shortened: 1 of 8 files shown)


msayler:
I assume the subtle change to unifying the SimpleDB version (`SDB_TIMEOUT`) with this isn't worth it?

motobob:
I can tackle it here if needed but whole SDB retry logic is somewhat really messy and to be revised as a whole so I would defer to the next round. Also it's defined in the module so it will not bring issues to library consumers.

src/erlcloud_cloudtrail.erl
```diff
@@ -192,8 +192,3 @@ port_spec(#aws_config{cloudtrail_port=80}) ->
 port_spec(#aws_config{cloudtrail_port=Port}) ->
     [":", erlang:integer_to_list(Port)].
 
-
-timeout(#aws_config{timeout = undefined}) ->
-    1000;
```
smilerlee: This commit changed the timeout for cloudtrail from 1s to 10s.

smilerlee:
Thanks @motobob, reviewed, this PR can be merged as long as cloudtrail timeout get fixed.

src/erlcloud_cloudtrail.erl
motobob: good spotting. addressed. 
in future I would really like to understand why DDB and CT are that different and the actual reason for that.

smilerlee:
@motobob 
As to DDB, @ransomr has explained in [#257](https://github.com/erlcloud/erlcloud/pull/257#issuecomment-186407850)

motobob:
@smilerlee thank you.

Issue #267, opened 2016-03-26
rebar 2 build still requires meck, hex package no longer requires it


nickveys:
Just updated to 0.13 and everything worked fine until I cleaned my dependencies and re-built.  Now meck is no longer pulled down, but the rebar2 config (https://github.com/erlcloud/erlcloud/blob/master/rebar.config.script) still requires it.  Breaks compilation of erlcloud with:

````
WARN:  Expected <<myapp>>/_build/dev/lib/meck to be an app dir (containing ebin/*.app), 
  but no .app found.
==> erlcloud (compile)
WARN:  Expected <<myapp>>/_build/dev/lib/meck to be an app dir (containing ebin/*.app), 
  but no .app found.
Dependency not available: meck-.* ({git,
                                       "https://github.com/eproxus/meck.git",
                                       {tag,"0.8.4"}})
ERROR: compile failed while processing <<myapp>>/deps/erlcloud: rebar_abort
** (Mix) Could not compile dependency :erlcloud, "/Users/nick/.asdf/installs/elixir/1.2.0/.mix/rebar 
  compile skip_deps=true deps_dir="<<myapp>>/_build/dev/lib"" command failed. You can
  recompile this dependency with "mix deps.compile erlcloud", update it with 
  "mix deps.update erlcloud" or clean it with "mix deps.clean erlcloud"
````

If I comment meck out, it builds, but I'm sure that's not the proper fix, since your tests must still need it, otherwise I'd send a PR!

waisbrot:
When I run locally (or under the Travis tests) with Rebar 3, I either run `rebar3 compile` which does _not_ fetch meck and is happy or I run `rebar3 eunit` which _does_ fetch meck and is also happy. When I run with Rebar 2, the `rebar.config.script` injects meck as a dependency (rebar 2 doesn't differentiate between test dependencies and build dependencies) and things work.

From what you've posted, I assume you're using "mix" as your build tool. I haven't tested this (and am not actually sure how to test, since I've never done an Elixir project). If you made a PR with some additions to the `Makefile` and `.travis.yml` files to demonstrate the failure, that would be very helpful.

motobob:
@nickveys thank you for reporting the issue. strangely that worked great on my setup and same as inside travis.ci where we specifically check that rebar2 build passes.
Just in case, notice the `get-deps` vs `deps` difference.

this is my output:
```sh
mbr13_ebob:erlcloud ebob$ rm -rf deps
[...]

PR #270, opened 2016-04-02
Adding Mixfile to support Elixir builds

nickveys:
From issue #267, builds using mix after v0.13 were failing due to dependency resolution issues.  This adds a simple Mixfile that allows erlcloud builds in Elixir along with its dependencies.

Tested locally, as a dependency to my application via a mix dependency of: `{:erlcloud, "~> 0.13", path: "/path/to/dir"}` and it downloaded dependencies, compiled and executed as expected.

Can tweak with feedback, let me know!

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -1,0 +1,37 @@
+defmodule Erlcloud.Mixfile do
+use Mix.Project
+
+    def project do
+        [
+            app: :erlcloud,
+            version: "0.13.1",
+            description: "Cloud Computing library for Erlang (Amazon EC2, S3, SQS, SimpleDB, Mechanical Turk, ELB)",
+            deps: deps,
+            language: :erlang
+        ]
+    end
+
+    def application do
+        [applications:
+            [
+                :kernel,
+                :stdlib,
+                :kernel,
+                :crypto,
+                :public_key,
+                :ssl,
+                :xmerl,
+                :inets,
+                :jsx,
+                :lhttpc
+            ],
+        ]
+    end
+
+    def deps do
+        [
+            {:jsx, "2.8.0"},
+            {:lhttpc, "1.4.0"},
+        ]
+    end
+end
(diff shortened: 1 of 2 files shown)


mix.exs
```diff
@@ -0,0 +1,38 @@
+defmodule Erlcloud.Mixfile do
+use Mix.Project
+
+    def project do
+        [
+            app: :erlcloud,
+            version: "0.13.1",
```
motobob: it's already 0.13.2 and I wonder if MIX support `git` based versioning alike rebar like [here](https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud.app.src#L6)

waisbrot:
I tried `mix test` and it wanted `test/test_helper.exs`. Would it be easy to add a stub version of this that just called EUnit to run the existing tests? 

I think there should be a Travis build for the mixfile (even if it's just trying to compile without error).

motobob:
I would like to nail this one.
could someone show me an example of how to integrate this elixir building into travis so that we can automate that?

waisbrot:
Here is my in-progress attempt: WhoopInc/dogstatsde#15

I think I'm close, but I've been thinking that for the last 10 commits...

nickveys:
Is this worthwhile if the issue that drove this (https://github.com/erlcloud/erlcloud/issues/267) will (seemingly) go away with Elixir 1.3?

waisbrot:
In Elixir 1.3, Mix started ignoring the mixfile for me and just running rebar3 to compile. I don't know if ignoring mix.exs in favor of rebar.config is a bug, but I believe Elixir should handle Erlcloud happily as-is.

It would be nice to figure out how to get Travis to run some Elixir tests for us. Something where it builds an Elixir project with an erlcloud dependency.

I'm going to close this PR, but give me a kick if someone's got a better idea.

Issue #266, opened 2016-03-24
readme is outdated and requires modifications to current 


motobob:
submitting this for myself to not to forget to update it

motobob:
#319

PR #271, opened 2016-04-12
Fix `rebar3 edoc` command

waisbrot:
I've published the result to https://hexdocs.pm/erlcloud/ so the Hex package now shows docs available.

Are there still plans to do a big merge in from the AlertLogic fork? I'd like to do some clean-up and improvement of the edocs, but I can just maintain these docs as-is in a separate branch (closing this PR) if that's better.

### DIFF
--- a/src/erlcloud_ddb2.erl
+++ b/src/erlcloud_ddb2.erl
@@ -2229,13 +2229,13 @@ update_table(Table, Opts) ->
 %% ===Example===
 %%
 %% Update table "Thread" to have 10 units of read and write capacity.
-%% Update secondary index <<"SubjectIdx">> to have 10 units of read write capacity 
-%% `
+%% Update secondary index `<<"SubjectIdx">>' to have 10 units of read write capacity
+%% ```
 %% erlcloud_ddb2:update_table(
 %%   <<"Thread">>,
 %%   [{provisioned_throughput, {10, 10}},
 %%    {global_secondary_index_updates, [{<<"SubjectIdx">>, 10, 10}]}])
-%% '
+%% '''
 %% @end
 %%------------------------------------------------------------------------------
 -spec update_table(table_name(), update_table_opts(), aws_config()) -> update_table_return();
--- a/src/erlcloud_ddb_impl.erl
+++ b/src/erlcloud_ddb_impl.erl
@@ -38,7 +38,7 @@
 %%
 %% Here is an example retry function that provides logging and an updated error reason:
 %%
-%% `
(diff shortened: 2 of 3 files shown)


smilerlee:
Thanks @waisbrot. I think reading source code and inline documentation is a must-have habit of every Erlang programmer, thus completing the inline documentation and the README file is a preferable approach.

src/erlcloud_ddb2.erl
```diff
@@ -2229,13 +2229,13 @@ update_table(Table, Opts) ->
...
-%% `
+%% Update secondary index `<<"SubjectIdx">>' to have 10 units of read write capacity
+%% ```
 %% erlcloud_ddb2:update_table(
 %%   <<"Thread">>,
 %%   [{provisioned_throughput, {10, 10}},
 %%    {global_secondary_index_updates, [{<<"SubjectIdx">>, 10, 10}]}])
-%% '
+%% '''
 %% @end
```
smilerlee: Is this a markdown code block? I think it both begins with and ends with three backquotes.

waisbrot: Markdown does, but EDoc uses some [Wiki notation](http://erlang.org/doc/apps/edoc/chapter.html#Wiki_notation)

waisbrot:
Sorry, that wasn't clear to me. I would call the edocs "the inline documentation". Are you talking about adding more [comments on the same line as the code](https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_s3.erl#L188)?

What sort of improvements to the README would you like? I'm interested in helping, especially if the work can be broken down into relatively small units. 

The main thing I'd wish for myself in terms of documentation would be adding `spec`s everywhere, getting Dialyzer to run cleanly, using more `type` aliases to document function args, and usage examples for each module.

src/erlcloud_ddb2.erl
smilerlee: I see. my mistake :)

smilerlee:
I think a brief guide and small example pieces for each services could be added to the README file, do you agree? see #266 
As to the edocs: specs, modules/funs/types/args/options comments (if necessary), and also examples.
Please feel free to contribute.

motobob:
@waisbrot the plan to merge in AL fork is still in mind - just can't find timeslot to properly plan/communicate it and merge.
Therefore let fix is here. also i quite often merge upstream to AL to stay on top.

motobob:
merging this :+1:

PR #272, opened 2016-04-21
Typo fix

x0id:
There was a typo the erlcloud_mon module making impossible to put statistics to CloudWatch.


motobob:
thank you. 
that is an unfortunate mistake in original implementation.

PR #274, opened 2016-04-27
Add support for aws cli credentials profiles

fp:
Allow config credentials to be sourced from the `~/.aws/credentials`
and support both `source_profile` and `role_arn` specifications.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -17,8 +17,18 @@
 
 {deps, [
         {jsx, "2.8.0"},
-        {lhttpc, "1.4.0"}
+        {lhttpc, "1.4.0"},
+        {eini, ".*",
+         {git, "https://github.com/accense/eini.git",
+          {tag, "1.2.1"}}}
        ]}.
+
+{overrides,
+ [
+  %% do not pull in the covertool plugin or repo, cause it fetches rebar and
+  %% breaks rebar3!
+  {override, eini, [{plugins, []},{deps, []}]}
+ ]}.
 
 {profiles, [
             {test, [{deps, [{meck, "0.8.4"}]}]}
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -15,7 +15,8 @@
          sign_v4_headers/5,
          sign_v4/8,
          get_service_status/1,
-         get_timeout/1
+         get_timeout/1,
+         profile/0, profile/1, profile/2
 ]).
 
 -include("erlcloud.hrl").
@@ -29,6 +30,13 @@
          security_token=undefined :: string(),
          expiration_gregorian_seconds :: integer()
         }).
+
+-record(profile_options, {
+          session_name :: string(),
+          session_secs :: 900..3600,
+          external_id :: string()
+}).
+
 
 aws_request_xml(Method, Host, Path, Params, #aws_config{} = Config) ->
     Body = aws_request(Method, Host, Path, Params, Config),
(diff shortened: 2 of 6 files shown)


waisbrot:
The rebar2 builds are failing in Travis:

```
WARN:  Missing plugins: [rebar_covertool]
==> covertool (compile)
Compiling /home/travis/build/erlcloud/erlcloud/deps/covertool/src/rebar_covertool.erl failed:
/home/travis/build/erlcloud/erlcloud/deps/covertool/include/covertool.hrl:1: record config has field(s) without type information
ERROR: compile failed while processing /home/travis/build/erlcloud/erlcloud/deps/covertool: rebar_abort
make: *** [eunit] Error 1
```

fp:
I've looked into this and covertool, which is pulled in by the new eini, has a record without types for the fields, which for some reason is treated as errors in this build. I'll fix it upstream, so that this works as is when merged.

Please review the change for anything else, since this is just a warning treated as error in a build only package (which you get with rebar2... sigh)

rebar.config
```diff
@@ -17,9 +17,19 @@
 
 {deps, [
         {jsx, "2.8.0"},
-        {lhttpc, "1.4.0"}
+        {lhttpc, "1.4.0"},
+        {eini, ".*",
+         {git, "https://github.com/accense/eini.git",
+          {branch, "master"}}}
```
waisbrot: I think you should use a tag or a hash, not "master".

Having a non-Hex dependency makes publishing to Hex problematic.

src/erlcloud_aws.erl
```diff
@@ -494,3 +501,189 @@ get_filtered_statuses(ServiceNames, Statuses) ->
...
+%%  role_arn=arn:aws:iam::892406118791:role/centralized-users
+%%  source_profile = default
+%% </pre></code>
+%%
+%% When using the the <em>role_arn</em> specification, you may supply the
+%% following two options to control the way in which the assume_role request
+%% is made via AWS STS service:
+%%
+%% <ul>
+%%  <li><code>'rold_session_name'</code>
```
waisbrot:
```
role_session_name
   ^
```

waisbrot:
I think this is a nice change. Should `erlcloud_aws:default_config/0` call `erlcloud_aws:profile/0`? That's how the other SDKs behave.

rebar.config
smilerlee: Agree, stick to a stable tag is reasonable.

src/erlcloud_aws.erl
```diff
@@ -494,3 +501,189 @@ get_filtered_statuses(ServiceNames, Statuses) ->
...
+%%
+%% @see profile/2
+%%
+profile() ->
+    profile( default, [] ).
+
+
+%%%---------------------------------------------------------------------------
+-spec profile( Name :: atom() ) -> {ok, aws_config()} | {error, string()}.
+%%%---------------------------------------------------------------------------
```
smilerlee: According to your implementation, `Name` is either an atom or a binary. It could be better if string also be supported, since binary is not that writing-friendly.

smilerlee:
👍  for this change. @waisbrot I think the call is necessary.

src/erlcloud_aws.erl
```diff
@@ -494,3 +501,189 @@ get_filtered_statuses(ServiceNames, Statuses) ->
...
+
+profiles_assume( Id, Secret, undefined, _Options ) ->
+    Config = #aws_config{ access_key_id = Id, secret_access_key = Secret },
+    {ok, Config};
+profiles_assume( Id, Secret, Role,
+                 #profile_options{ session_name = Name,
+                                   session_secs = Duration } ) ->
+    Config = #aws_config{ access_key_id = Id, secret_access_key = Secret },
+    {AssumedConfig, _Creds} =
+        erlcloud_sts:assume_role( Config, Role, Name, Duration, undefined ),
```
motobob: need to attempt to extract `external_id ` from profile too instead of passing `undefined`  right away

src/erlcloud_aws.erl
```diff
@@ -494,3 +501,189 @@ get_filtered_statuses(ServiceNames, Statuses) ->
...
+%% </pre></code>
+%%
+%% as well as indirection using <em>source_profile</em>:
+%%
+%% <code><pre>
+%%  [foo]
+%%  source_profile = default
+%% </pre></code>
+%%
+%% and finally, will supports the <em>role_arn</em> specification, and will
```
motobob: role based credential will expire in `session_secs`. Profile loading assumes role only once. 
As a library consumer its unclear (since expiration is not stored in `#aws_config`) when and how I should re-assume role before expiration before execution next AWS call.

I see two options:
- do not assume the role while loading profile and let consumers do it the way they wish
- provide a facility which will hide the token renewal from the library consumer. Preferable option but one may argue it's library complication.

src/erlcloud_aws.erl
motobob: same applies to `role_session_name`. this parameter is optional. Since we are trying to be on par with CLI I think we should supprt that too. 
see http://docs.aws.amazon.com/cli/latest/topic/config-vars.html

src/erlcloud_aws.erl
motobob: another "compromise" alternative is to propagate STS returns `expiration` into `aws_config` and let library consumers handle it.

src/erlcloud_aws.erl
fp: ok, i'll merge in support for the other options that can appear in the profile.

src/erlcloud_aws.erl
fp: I can add the expiration time to the `#aws_config{}` record so that callers can refer to it.

@motobob On the rest, I think what you are really asking for is to have `erlcloud_sts:assume_role` produce a config that will auto-renew? I wouldn't think that this should apply only to profile based configs, but all such configs.  If you agree, I'll be happy to add that support to the `erlcloud_sts` and the auto-renew support to `erlcloud_aws`.

BTW, this generally applies to metadata based credentials, which probably would benefit from having a single mechanism used for renewal, since there would be at least two different config credential generators at that point.

rebar.config
fp: good catch, there does appear to be a tag to use, so I'll fix that.

src/erlcloud_aws.erl
motobob: @fp you're right. the completeness of this will be to apply assume role re-newal for all assume calls and not just profiles, which is a separate feature to be added. 
So let's stop at adding `expiration` to config within this PR so that library consumer could re-new themselves

fp:
@motobob @waisbrot please check that I have addressed all your comments

motobob:
@fp 👍 looks like all the comment were addressed. 
let's give @waisbrot some time to respond and i'll merge it.

waisbrot:
👍 

I'll make a PR on eini asking for a Hex package..

PR #275, opened 2016-05-03
Fix sts expiration processing

fp:
Mix up with `erlcloud_xml:decode` and converting the `expiration` timestamp. This has been live verified!


src/erlcloud_sts.erl
```diff
@@ -117,13 +117,14 @@ sts_query(AwsConfig, Action, Params, ApiVersion) ->
     ).
 
 
+expiration_tosecs( Datetime ) when is_tuple(Datetime) ->
+    GregorianSeconds = calendar:datetime_to_gregorian_seconds( Datetime ),
+    (GregorianSeconds - ?UTC_TO_GREGORIAN);
 expiration_tosecs( Timestamp ) ->
+    io:format( "timestamp: ~s~n", [Timestamp] ),
```
motobob: no need for this

fp:
@motobob removed!

motobob:
tnx for promt fix.

Issue #278, opened 2016-05-16
Erlang/OTP 19.0 requires switching from random:uniform/1 to rand:uniform/1


okeuday:
The random module is deprecated, for removal in 20.0, so when warnings are turned on the compilation will fail with 19.0.  If an erlcloud_random.erl module is added to switch between random:uniform/1 and rand:uniform/1 based on preprocessor version macros, it can be easy to support older versions too.  An example of the macros which might be used is at https://github.com/okeuday/uuid/blob/master/rebar.config#L4-L22 .  Is there any reason to not pursue this approach?

okeuday:
`erlcloud_util.erl` appears to be meant for compatibility purposes, but due to its runtime checking it still creates warnings.  That appears to be the proper place, based on the hash functions there.

smilerlee:
We've been working on the obsolete usages and backward compatibilities problems, like what we did [here](https://github.com/erlcloud/erlcloud/commit/36b00f81c8ff38f0ab0b5b954fc7e8e1e4e87804).

A PR for this is welcome.

okeuday:
I added changes for this at https://github.com/erlcloud/erlcloud/pull/287

motobob:
adressed

PR #279, opened 2016-05-16
Fixes for Erlang/OTP 19.0

okeuday:
* -spec syntax is fixed to not cause compilation syntax errors
* -include_lib for erlcloud is changed to -include to work with rebar2

Other changes that would be nice, but are not included in this pull request are mentioned in the issue  #278

### DIFF
--- a/src/erlcloud_as.erl
+++ b/src/erlcloud_as.erl
@@ -1,7 +1,7 @@
 -module(erlcloud_as).
 
--include_lib("erlcloud/include/erlcloud_aws.hrl").
--include_lib("erlcloud/include/erlcloud_as.hrl").
+-include("erlcloud_aws.hrl").
+-include("erlcloud_as.hrl").
 
 %% AWS Autoscaling functions
 -export([describe_groups/0, describe_groups/1, describe_groups/2, describe_groups/4,
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -21,7 +21,7 @@
 ]).
 
 -include("erlcloud.hrl").
--include_lib("erlcloud/include/erlcloud_aws.hrl").
+-include("erlcloud_aws.hrl").
 
 -define(ERLCLOUD_RETRY_TIMEOUT, 10000).
 
--- a/src/erlcloud_cloudformation.erl
+++ b/src/erlcloud_cloudformation.erl
@@ -1,7 +1,7 @@
 -module (erlcloud_cloudformation).
 
--include_lib("erlcloud/include/erlcloud.hrl").
--include_lib("erlcloud/include/erlcloud_aws.hrl").
+-include("erlcloud.hrl").
+-include("erlcloud_aws.hrl").
 
 -define(API_VERSION, "2010-05-15").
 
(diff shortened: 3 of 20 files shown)


motobob:
this is great to hear a PR for prelim support for R19 after that few days of announce!
Before we start this venture amy I ask to include R19 into travis targets(within this PR) as current auto checks are irrelevant to the scope changes here.

okeuday:
@motobob 19.0 has only had a rc1 release so far, so no official release has happened yet.  This means that it can't be added to travis yet, since travis won't be including it in images yet.

motobob:
I agree that it's a little early...
However, `kerl` has it so that way it be patched. 
I'm just not yet comfortable merging such a big change in specs all over the place. especially given that dialyzer was not that happy even with the original number of specs. 

Can you comment on:
>-include_lib for erlcloud is changed to -include to work with rebar2

majority of consumer of this library uses rebar2 and it works all fine with then.  what specific issue are we fixing here. I'm both hands up for the change  (probably with a separate PR) but I would like here a specific problem we are addressing.

okeuday:
@motobob The include_lib usage with erlcloud will not work if the erlcloud repo is cloned into a directory name that does not match "erlcloud" (at least with rebar2 at tag `2.1.0-pre`, I have used an old tag to avoid instability that was common in the past with the rebar2 development and that tag proved to provide all the critical functionality without new problems, I haven't tried the new tags with this specific compilation issue but would expect the same result).

A separate concern is the unfortunate use of {vsn, git} in the .app.src file, which does not match the http://erlang.org/doc/man/app.html format as defined by Erlang/OTP (an atom rather than a string).  While you might not think this matters, since the typical use of the repo is to depend on rebar to substitute the tag name, that depends on usage within this repo, and ignores using the source code as a directory elsewhere.  It is much more dependable to have a string version that is updated manually, and viewable in the git history, even if commits are deleted or changed within the commit history.

smilerlee:
Using `-include` instead of `-include_lib` seems much more reasonable, as far as I can tell most of the official libraries are using this approach.
It would be great to have a separate PR.

okeuday:
@smilerlee The reason these two changes are together is due to both being required for getting erlcloud back to a state where it compiles with Erlang/OTP 19.0.  You will notice the older source code does use `-include` instead of `-include_lib`.  This is just fixing a regression with the newer source code.  The `-spec` changes are changing the syntax to the accepted format.  I find it strange that we are still talking about the changes, due to none of the changes affecting the functionality or the compilation result.

smilerlee:
@okeuday I vote for both two changes.

* `-include_lib` should only be used when a external library file being included;
* `-spec Function(ArgType1, ..., ArgTypeN) -> ReturnType.` is what I've been used from the beginning, as stated in the [official documentation](http://erlang.org/doc/reference_manual/typespec.html#id79031)

I thought there are no needs to discuss the changes effects any further since these changes are self-explanatory.

I know they both address compile problems in Erlang/OTP 19, but different problems however, at least in separate commits?

okeuday:
@smilerlee Now the changes are in separate commits.

smilerlee:
@okeuday thanks.
@motobob what do you think?

motobob:
thnx for splitting commits. 
I'm okay as it not feasible for me to check if there are any typos. So I will trust travis/dialyzer here that at least current consumers are not broken. 

@okeuday thank you for this great effort of preparation for R19 and making dialyzer more happy.

i'll submit a separate Enhancement here to just not forget to add R19 to the auto builds.

PR #280, opened 2016-05-17
Add eini application dependency.

okeuday:
The eini dependency is currently missing from the .app.src file, but is present in the rebar.config.  Due to the fetch of the dep being required and the dependency being used in the source code, it is best to make it a dependency in the .app file that is generated.


motobob:
@okeuday good catch. i missed that during original PR for adding eini

PR #285, opened 2016-05-24
Make associate_public_ip optional in ec2 instance creation

timofey-barmin:
If more than one interface is specified server returns
"The associatePublicIPAddress parameter cannot be specified when
launching with multiple network interfaces"

### DIFF
--- a/src/erlcloud_ec2.erl
+++ b/src/erlcloud_ec2.erl
@@ -2751,21 +2751,23 @@ run_instances(InstanceSpec, Config)
             Error
     end.
 
-net_if_params(#ec2_net_if{private_ip=undefined}=X) ->
-    [
-        {"DeviceIndex", X#ec2_net_if.device_index},
-        {"SubnetId",    X#ec2_net_if.subnet_id},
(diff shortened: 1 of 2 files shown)


src/erlcloud_ec2.erl
```diff
@@ -2751,21 +2751,23 @@ run_instances(InstanceSpec, Config)
             Error
     end.
 
-net_if_params(#ec2_net_if{private_ip=undefined}=X) ->
-    [
-        {"DeviceIndex", X#ec2_net_if.device_index},
-        {"SubnetId",    X#ec2_net_if.subnet_id},
-        {"AssociatePublicIpAddress", X#ec2_net_if.associate_public_ip}
-    ] ++ erlcloud_aws:param_list(X#ec2_net_if.security_group, "SecurityGroupId");
+private_ip_params(undefined) -> [];
```
motobob: record definition does not mention `undefined` so it has to be `[]` too.
or let's fix the record define.

timofey-barmin: check it out

motobob:
👍

PR #286, opened 2016-05-24
Add support for elb attributes

posilva:
Add support to get attributes from elb 

- In my case this is useful to evaluate the connection draining attribute and have a automated decommissioned of instances when the number of connections get 0.


src/erlcloud_elb.erl
```diff
@@ -22,23 +24,26 @@
 
 -import(erlcloud_xml, [get_text/2]).
 
--spec new(string(), string()) -> aws_config().
+%%%===================================================================
+%%% API
+%%%===================================================================
+-spec(new/2 :: (string(), string()) -> aws_config()).
```
motobob: please return the specs back to the format used. that was fixed by #279

motobob:
👍

PR #287, opened 2016-06-14
Add compatibility source code for Erlang/OTP 19.0

okeuday:
* Switch to rand:uniform/1 instead of random:uniform/1 when using 19.x or higher

Please note:
* The proper dependency tag that is being used creates a hard dependency on Erlang/OTP 18.x and higher, but that can be handled by manually specifying rebar dependencies at a higher level
* This commit makes the supported Erlang/OTP versions explicit so it is easier to understand the versions supported by this dependency


motobob:
LGTM

PR #288, opened 2016-06-14
Use an explicit version

okeuday:
There are a few reasons for this change (mentioned earlier at https://github.com/erlcloud/erlcloud/pull/279#issuecomment-219836003):

1.  The use of the atom `git` is not valid .app file syntax (http://erlang.org/doc/man/app.html).
1.  Using the atom `git` instead of an explicit version string makes it more difficult to understand the version within the commit history, when examining commits.  While you can argue that as long as you know how to use git, you should be able to see what tag relates to what source code, the problem becomes more complex when the source code exists outside of this source code repository, when the tags may not be present.
1.  Having a direct relationship between the version a hex package represents and the source code repository represents is important.  So, updating the .app.src file at the same time as a hex package is created can help make sure both copies of the source code are used interchangeably at the same version, with no risk of extra changes that would defeat the purpose of having a dependable release process.
1.  Using explicit version numbers makes it easier to track the version of erlcloud being used, without random characters as a suffix.  That makes the erlcloud application look more dependable, since it should obviously have a dependable release process, if it has a dependable version number that can be referred back to within its commit history, within any copy of the source code.
1.  Requiring the .app.src file to change for a release encourages a release process, which will help to avoid incomplete changes and will promote cooperation to avoid causing errors or bugs.

Versioning is very important to make sure anything is reliable, whether it be in source code, a protocol, a schema, documentation, etc., so lets make sure the version is done in a dependable way.

### DIFF
--- a/src/erlcloud.app.src
+++ b/src/erlcloud.app.src
@@ -3,7 +3,7 @@
 
 {application, erlcloud,
  [{description, "Erlang cloud computing library"},
-  {vsn, git},
+  {vsn, "0.13.4"},
   {registered, []},
   {applications, [stdlib,
                   kernel,


waisbrot:
But this is not correct. 0.13.4 is [here](https://github.com/erlcloud/erlcloud/releases/tag/0.13.4) and you can see that there have been many commits since that release. The correct name for the head of the master branch is the Git ref. If you want to be friendly, it's the most recent tag + the Git ref.

The file you edited is the .app **.src** file. I don't believe it's covered by Erlang/OTP, but the various build tools all consume it and produce the (valid) .app file in the ebin directory.

okeuday:
@waisbrot Yes, 0.13.4 is the last actual release.  Pretending each commit is its own release is not helpful and is just ignoring the problem.  The .app.src file is the source of the .app file and it should have a valid format for the .app file.  Do all build tools for Erlang support having a git atom in the .app.src file? (no, because it is not standard, since there is no standard for a .app.src file except the format for a .app file)

Claiming each commit is a release would only be saying that no releases occur and each commit is as good/bad as the other.

waisbrot:
Well let's say that there's a huge change coming in. This will be a major, backwards-breaking change and there will be a release soon afterward. Under your scheme, before that new release is cut, the change is labeled as 0.13.4. That completely defeats the point of versioning, because now I have two different sets of code called "0.13.4" and they don't even present a compatible API.

When rebar3 builds, it detects when the current commit has a tag and it uses that tag as the release version. This is the correct behavior -- the only thing that's called 0.13.4 is the exact code that was present when then 0.13.4 release was determined.

I think that what you're suggesting would work under the Git-flow model of commits, where changes go into a "develop" branch and then there's a big merge into the "stable" branch and that's a release.

okeuday:
@waisbrot The old release version happens to stay in the .app.src file until a release actually happens, so it could be incorrect while more changes are committed, but that should be ok since the branch is a development branch, even if it is also the release branch.  As long as a release updates the .app.src file with a new commit and does a tag with the same version name (then uses the same version name with a hex package), everything is dependable when people use the tag or package.

If someone chooses to not use a tag, then they get the last version number despite the extra changes, but they would know that they are doing that (out of necessity, due to needing extra changes).  Their separate commit reference would require a separate release process, if it needed a separate version number that was meant to be regarded as "valid", but that is outside the scope of this repository.  It is best to focus on the release that is relevant to the changes in this repository and how that impacts these release versions.

I agree the situation is simpler when using more than one branch for releases, but I don't think it is required to have a separate "stable" or "release" branch (usually this is just "master" to make sure it is the easiest branch to get) since the person getting the erlcloud dependency should always be referencing a commit or tag to get a repeatable result of their build.  If they don't, then that is their development problem, which isn't necessarily relevant to this repository.

motobob:
@okeuday this change breaks the key principal which @waisbrot refers to - there should not be two different versions of code with the same version revision `0.13.4` or `foo`

nobody is pretending that every merge to master is a separate release:
- not every change deserves a dedicated tag/release. there are a lot of changes which are super minor.
- release tag is added often in the library. I'm keeping it simpler.
- if app.src has a hardcoded version string and not git it will block parallel development of PRs as each merge will produce a conflicts in other PRs.

There are two different SW configuration managements strategies:
- pinning to master to get all the latest changes every time your rebuild your app. This lets and relays on the dependancy be managed properly with no breaking changes. We are maintaining backward compatibility here. 
- pinning to a specify tag. and handling upgrades to new version manually, like:
```erlang
https://github.com/alertlogic/erlcloud/blob/master/rebar.config.script#L23
```

If your use case dictates to pin to a very specific version, please do it via rebar.config/ mix.esx, etc and not via the changes in app.src.

okeuday:
@motobob I believe your comment doesn't accurately represent the situation.  A version is the product of a release.  If you agree that a version is the product of a release, then there are never two versions of the code with the same version, since that statement is nonsensical.  If you do not agree that a version is the product of a release, then you are pretending that every commit is a separate release with `git` atom usage for a version in the .app.src file, despite the syntax is invalid based on the standard .app file format.

The list of different SW configuration management strategies is incorrect, because ideally the usage should be either pinning to a commit hash (not the master branch, since that randomly accepts any changes) or pinning to a tag to get a specific version.  Either way, you are agreeing to a specific commit which may or may not be part of a release.  The version only represents the release that occurred in the past.

The current situation does not provide any benefit and only provides the negative aspects mentioned previously.  Faking a release only makes this repository look less trustworthy, since that would indicate a release isn't really a release and a version isn't really a version.

motobob:
instead of having a semantical argument about `version`, `release`, `tag` and `product`  let's come back to the point. 
please clarify what exactly is the problem you are having with current versioning, what does it break in your use case and how this PR fixes that.

okeuday:
@motobob If you refer to the items in the description, that is an accurate representation of the problems.  I have switched from bullet points to a numbered list to make this easier to discuss.  The current status on these items, based on the discussion so far is below:

1.  So far, it appears you don't care about nonstandard .app.src syntax not matching the standard .app syntax
1.  The git commit history not being as useful is related to the ability to rely on this repository directly
1.  Having a dependable release is required to make sure this repository is dependable.  So far, the discussion shows a lack of clarity on what a release is for this repository, since there is a lack of clarity on what a real version is.
1.  This relates to runtime usage of the application version (hasn't yet been discussed).
1.  This describes how this source code change (single line) is directly related to having a dependable release process (hasn't yet been discussed).

nalundgaard:
+1 to @ebob's comments.

@okeuday are you familiar with how the `vsn` `git` in rebar (or `"git"` in rebar3) works? It kind of sounds like you haven't [read about it](https://github.com/rebar/rebar/wiki/Processing-versions-in-configuration-files) ().

This is pretty standard operating procedure in Erlang OTP applications, including, say, [rebar3 itself](https://github.com/erlang/rebar3/blob/master/src/rebar.app.src#L6).

Releases are defined via tags. Rebar builds the `erlcloud.app` using `git` in the `erlcloud.app.src`, respecting the release set by the git tag. If you failed to checkout a release, and instead checked out master, in order to build, you will get a dirty version tag like `"v1.2.1-5-gc4ccc0a"`. These are clearly identifiable as non-releases by their divergence from the well-defined version format. This is awesome for us as developers.

I am strongly against this change.

nalundgaard:
I forgot to mention, [here](https://github.com/erlcloud/erlcloud/releases) are the erlcloud releases. The latest one is clearly displayed on the top of the page. My understanding is that this is very standard behavior for a GitHub project.

waisbrot:
@okeuday all your numbers read to me like opinions about how releases and versions and tagging ought to work. Can you phrase them in terms of bugs caused by the current behavior (rather than just "the current behavior is a bug"). 

For example, if you could provide an example of a build that fails because the .app.src has an atom for the version string (1), that might make things clearer.

nalundgaard:
@okeuday one more thing. I see one of your projects that has a [tag/release/whatever](https://github.com/okeuday/pqueue/blob/v1.5.1.1/src/pqueue.app.src#L3) where the git tag is v1.5.1.1, but you didn't update the `vsn` in the .app.src, so it's still claiming the last version, v1.5.1. In my opinion, this is broken, and a common result of the change that you are proposing. Much less trustworthy for developers.

This is what my erlcloud.app file looks like when I build erlcloud from master right now:

```erlang
{application,erlcloud,
             [{description,"Erlang cloud computing library"},
              {vsn,"0.13.4-22-ga9bf20b"},
              {registered,[]},
              {applications,[stdlib,kernel,crypto,public_key,ssl,xmerl,inets,
                             eini,jsx,lhttpc]},
              {modules,[erlcloud,erlcloud_as,erlcloud_autoscaling,
                        erlcloud_aws,erlcloud_cloudformation,
                        erlcloud_cloudtrail,erlcloud_ddb,erlcloud_ddb1,
                        erlcloud_ddb2,erlcloud_ddb_impl,erlcloud_ddb_streams,
                        erlcloud_ddb_util,erlcloud_directconnect,erlcloud_ec2,
                        erlcloud_elb,erlcloud_http,erlcloud_httpc,
                        erlcloud_iam,erlcloud_inspector,erlcloud_json,
                        erlcloud_kinesis,erlcloud_kinesis_impl,erlcloud_kms,
                        erlcloud_lambda,erlcloud_mon,erlcloud_mturk,
                        erlcloud_rds,erlcloud_redshift,erlcloud_retry,
                        erlcloud_route53,erlcloud_s3,erlcloud_sdb,
                        erlcloud_ses,erlcloud_sns,erlcloud_sqs,erlcloud_sts,
                        erlcloud_util,erlcloud_waf,erlcloud_xml]},
              {env,[]},
              {licenses,["MIT"]},
              {links,[{"Github","https://github.com/erlcloud/erlcloud"}]},
              {maintainers,[]}]}.
```

Just what I want. 22 commits ahead of the last release, and in a tag format that indicates it's ahead of the `0.13.4` release, but will always be behind a future `0.13.5`, `0.14.0`, etc. release.

okeuday:
@nalundgaard Yes, I am aware of `git` working in the .app.src file with rebar, which is the reason for this discussion and pull request.  rebar3 is not standard enough to be included in Erlang/OTP, though its repository has been moved into the Erlang organization, so it is still effectively beta software (if not alpha).  I do not see a dirty version as helpful and I also am a developer.  The releases link you provided just shows what github does with tags and I am not questioning the usage of tags.  The `v1.5.1.1` tag you mentioned relates to semantic versioning (http://semver.org/) and the lack of any release, but the need elsewhere to use the tag due to recent additions or simply just to make sure future changes don't impact other software.  That is common usage of tags, tags are not specifically used only for releases since they can also be used inbetween releases.

@waisbrot A basic example is copying erlcloud into https://github.com/CloudI/CloudI/tree/develop/src/external/cloudi_x_erlcloud .  If I did not change the version from the `git` atom to a valid string, it would be unable to function properly, due to being a separate repository.  Your next question would be why would I copy the erlcloud repository?  I replicate all the dependencies to make sure CloudI is dependable, isolated from this repository disappearing, moving, having its commit history rewritten and other problems (if you are wondering about the cloudi_x_ prefix, I have a script to add a prefix to source code during build-time, to simulate a namespace in Erlang, due to it not being a programming language feature).  For fault-tolerance, it is important to make sure your build is not impacted by random behaviour of random people on the internet.  Typically, private usage of the erlcloud repo would be a fork that needs to replicate the tags, which shouldn't be a problem as long as it remains a git repository.  You may not value the fault-tolerance aspect of this problem, but it is important, since Erlang's main advantage is the fault-tolerance it is able to provide.

Again, just to be clear, a version should be the product of a release.  The dirty version that is currently being created is not dependable since it depends on the tags in this specific git repository.  Having dependable releases and dependable versions is important enough that the version needs to be explicit.

nalundgaard:
@okeuday okay, fair enough. If you don't use rebar to package your software, I propose that an ideal solution is that you maintain a separate fork of this repo. I reiterate my vote against this PR.

okeuday:
@nalundgaard I invite you to re-read the comment.  I actually use rebar for my development.

nalundgaard:
My original comment [linked to rebar](https://github.com/rebar/rebar/wiki/Processing-versions-in-configuration-files). As in, not rebar3—rebar 2.x. replacing `{vsn, 'git'}` to the result of `git describe --always --tags` is standard rebar behavior. In fact, they changed it in rebar3 (to use the string `"git"`, so this is specific to rebar 2.x. I realize you called out rebar3 in your comment above, but I don't understand your point, because, again, both versions of rebar use the same (slightly different in terms of atom vs. string) behavior for standard and fully OTP-compliant versioning in the `.app` file.

okeuday:
@nalundgaard The atom `git` required for rebar2 support is not valid .app syntax as mentioned/linked in item 1 within the description.  Making it a string `git` only avoids the syntax problem while risking it not breaking with rebar2 or some other build situation, such that dependency versions become "git" which should at least cause breakage during a release, hopefully.

waisbrot:
@okeuday Thanks for the explanation. I see how the proposed change would make things easier for you. However, as another user of Erlcloud it would make things harder for me: I generally stick to releases but I sometimes use HEAD because there's a new bug-fix or feature which I desperately need. In these cases, it's important to me that I have a simple way to check what version I'm running: is it a release? The latest release? Not a release at all?

The way Erlcloud does releases (with the dynamic `git` value) matches my expectations for a library and I think it matches the expectations of most users.

I don't know if you've tried Hex, but the rebar3 Hex plugin appears to perform `git`-version replacement before publishing the code, so that the .app.src on the Hex servers have the version tag you're wishing for. Maybe you could request/write a rebar3 "publish" plugin that would publish to a tarball or directory for you.

okeuday:
@waisbrot While using the `git` automatic version substitution that rebar provides for a git repository may be what you want, in-line with expectations for Alert Logic usage of erlcloud (based on the comments), I don't believe you speak for all users of erlcloud.  The change to use the automatic version substitution is a recent change that occurred when Alert Logic took control of this repository.  I find this direction disappointing, due to not meeting my expectations for release quality.  I have noticed the recent changes required in the erlcloud repository to get it back to a functional state are also a reduction in quality and these source code changes are concrete things we can talk about.  How is avoiding the dangers of an automatic release version a real problem?

okeuday:
Also, usage of hex is not relevant here.  The focus is on the version and usage for a release.

smilerlee:
@okeuday Imagine this situation:
I'm using erlcloud master branch as a dependency in a project, the version in `.app.src` was `0.13.4`.
Now there are some minor bug fixes newly added to the master branch, but `0.13.5` is not released yet.
If we use hard-coded `0.13.4`, how could I hot-upgrade my running server by generating appups between two identical version number?

okeuday:
@smilerlee No responsible developer is going to point their build at any random commit of the master branch for a dependable release.  A release done that way will always be undependable due to it depending on the time when the release is created.  All dependencies must be locked down to have a dependable build, so if a new change was required, a responsible developer would fork this repository to add the new change that is required separately, in a repository they control, which then allows them to create tags if they need to due to other necessary changes before a release of this repository.

Doing something problematic to solve a problem that doesn't exist, with an imagination about random commits producing dependable builds is not productive.  The main focus of a repository should be releases that are dependable, not automatically incrementing a version count with each new commit of dubious quality.  That is just giving up on release management and any pursuit of quality within a repository.

motobob:
The proposed change is already 2 releases behind. 
See https://github.com/erlcloud/erlcloud/releases. 
As I said earlier, I would not like to add this maintenance burden on every contributor to update the version and thus block parallel development as all the PR will become conflicting to each other at this line. 
I'm not okay bumping version manually all the time and have useless commits over and over again. 
We use GIT tags/releases and `rebar`/`rebar3` tooling for that. 

Thank you for your proposal about versioning GIT projects but majority of repo members would like to stay with the current approach.
The PR is now closed.

PR #289, opened 2016-06-17
Implement put method in aws_request4

x0id:
This patch helps to implement Elasticsearch API.

### DIFF
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -144,6 +144,11 @@ aws_request4(Method, Protocol, Host, Port, Path, Params, Service, Config) ->
             {error, Reason}
     end.
 
+aws_request4_no_update(put, Protocol, Host, Port, Path, Params, Service, #aws_config{} = Config) ->
+    Region = aws_region_from_host(Host),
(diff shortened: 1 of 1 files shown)


src/erlcloud_aws.erl
```diff
@@ -144,6 +144,11 @@ aws_request4(Method, Protocol, Host, Port, Path, Params, Service, Config) ->
             {error, Reason}
     end.
 
+aws_request4_no_update(put, Protocol, Host, Port, Path, Params, Service, #aws_config{} = Config) ->
```
motobob: i would rather ask to add a `put` case clause at https://github.com/erlcloud/erlcloud/pull/289/files#diff-b6a88197b95cfa7d84db56a756107eadR156 five lines below

or be simpler and use alike: 
https://github.com/alertlogic/erlcloud/blob/master/src/erlcloud_aws.erl#L153-L161

x0id: Adding **put** to case clause didn't help much, I tried. Actually, this is "dirty" patch. The problem was not only the put was not supported, but also narrow semantic of Params, which in my case would be binary (json). I will try to rework this.

x0id:
Revoked as questionable and not solving any problem in general.

PR #291, opened 2016-06-30
Implemented erlcloud_mon:get_metric_statistics for AWS CloudWatch

samble:


samble:
This is the first time I did a Erlang, let me know if I did it wrong please

smilerlee:
@samble I recommend you to squash all these commits into a single one.

Not familiar with CloudWatch API, @motobob you call.

samble:
@smilerlee Hm, I thought you would do that upon merge. I'll close this PR and do a new one squashed.

nalundgaard:
@samble You don't need to close this PR. You just need to squash your commits and force-push them over this PR branch (samble:master). The open PR will automagically update. You can also amend and force-push your single commit to address people's concerns on your PR without adding additional commits.

src/erlcloud_mon.erl
```diff
@@ -216,10 +217,51 @@ put_metric_data(Namespace, MetricName, Value, Unit, Timestamp) ->
...
 
+
+%%------------------------------------------------------------------------------
+%% @doc CloudWatch API - GetMetricStatistics - Easy average version
+%% Gets average stats at 60 second intervals for the given metric on the given instance for the given interval
+%% @end
+%%------------------------------------------------------------------------------
+-spec get_metric_statistics(
+        MetricName  ::string(),
+        StartTime   ::string(),
```
nalundgaard: The example doesn't show this as being a string? One or the other should be fixed.

samble:
@nalundgaard Automagic indeed. The function ran with the -spec having string()s and passing in datetime()s, but I updated the -spec to take datetime()s and this seems to work too.

waisbrot:
@samble `-spec`s are not used at run-time, they're just advisory. You need to run dialyzer (`rebar3 dialyzer`) to check function specs.

samble:
@waisbrot Understood on specs. I ran that command and got 646 warnings and no errors. Do I need to do anything there to get this PR merged?

src/erlcloud_mon.erl
```diff
@@ -216,33 +217,111 @@ put_metric_data(Namespace, MetricName, Value, Unit, Timestamp) ->
...
-    todo.
+    Config = default_config(),
+    Params =
+          [
+           {"Namespace",  Namespace},
+           {"MetricName", MetricName},
+           {"StartTime",  format_timestamp(StartTime)},
+           {"EndTime",    format_timestamp(EndTime)},
+           {"Period",     Period},
+           [{"Unit",      Unit} || Unit=/=undefined, Unit=/=""]
```
nalundgaard: doesn't this need to be flattened down?

samble: None of those param values are lists (those are further down) so I didn't think flattening would do anything.

nalundgaard: `[{"Unit", Unit} || Unit=/=undefined, Unit=/=""]` is a list comprehension. I think if you don't flatten the containing list, you will get bad formatting. see:

```erlang
1> [{"k1", "v1"}, [{"Unit", ""} || Unit=/=undefined, Unit=/=""]].
* 1: variable 'Unit' is unbound
2> [{"k1", "v1"}, [{"Unit", Unit} || Unit=/=undefined, Unit=/=""]].
* 1: variable 'Unit' is unbound
3> Unit = "".
[]
4> [{"k1", "v1"}, [{"Unit", Unit} || Unit=/=undefined, Unit=/=""]].
[{"k1","v1"},[]]
5> f(Unit).
ok
6> Unit = "test".
"test"
7> [{"k1", "v1"}, [{"Unit", Unit} || Unit=/=undefined, Unit=/=""]].
[{"k1","v1"},[{"Unit","test"}]]
```

Whereas, I think you want this:

```erlang
8> lists:flatten([{"k1", "v1"}, [{"Unit", Unit} || Unit=/=undefined, Unit=/=""]]).
[{"k1","v1"},{"Unit","test"}]
```

src/erlcloud_mon.erl
```diff
@@ -216,33 +217,111 @@ put_metric_data(Namespace, MetricName, Value, Unit, Timestamp) ->
...
 
+
+%%------------------------------------------------------------------------------
+%% @doc CloudWatch API - GetMetricStatistics - Easy average version
+%% Gets average stats at 60 second intervals for the given metric on the given instance for the given interval
+%% @end
+%%------------------------------------------------------------------------------
+-spec get_metric_statistics(
+        MetricName  ::string(),
+        StartTime   ::datetime(),
```
nalundgaard: It looks like these can be `datetime() | string()`, as the local module's `format_datetime/1` function has a head for both the datetime tuple and a string. So as long as you pass an AWS-compliant datetime string, that works too. Probably worth noting in the specs for `StartTime`/`EndTime`.

src/erlcloud_mon.erl
samble: Ah I get it, thanks for the explanation. It didn't matter if the param values were lists anyway, since flatten doesn't go inside tuples it looks like.

It worked the way I had it for some reason. It seems like it shouldn't. Instead of tracing down why I just added the lists:flatten.

src/erlcloud_mon.erl
samble: Added string() to the type union in the specs for both get_metric_statistics/4 and /8

src/erlcloud_mon.erl
samble: @nalundgaard It worked anyway because Unit's not a required field, and when Unit was passed inside the unflattened list, it was silently dropped from the request. I didn't test that properly, so you caught a bug.

samble:
@nalundgaard @motobob anything else to do here?

src/erlcloud_mon.erl
```diff
@@ -216,33 +217,111 @@ put_metric_data(Namespace, MetricName, Value, Unit, Timestamp) ->
...
+          lists:flatten(
+            [begin
+                 {Name, Value} = lists:nth(N, Dimensions),
+                 [{?FMT("Dimensions.member.~b.Name", [N]), Name},
+                  {?FMT("Dimensions.member.~b.Value", [N]), Value}]
+             end
+             || N<-lists:seq(1, length(Dimensions))]
+           ),
+    Doc = mon_query(Config, "GetMetricStatistics", Params),
+    Members = xmerl_xpath:string("/GetMetricStatisticsResponse/GetMetricStatisticsResult/Datapoints/member", Doc),
```
motobob: I do not see `Label` getting extracted. http://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_GetMetricStatistics.html
it might be useful for you application and needed from an library perspective

src/erlcloud_mon.erl
motobob: let's fix this @samble

motobob:
@samble 2 tiny updates ^ and i'll merge right after. Thank you.

src/erlcloud_mon.erl
samble: @motobob it looks like Label is just what we send as MetricName, and we must send exactly MetricName. Including that will put the useful datapoint results one level deeper into the returned object, since there has to be the top level Label string and then the DataPoint list.

Anyway if you think it's the right move, I did it, so I'll push it if you say the word. The results look like this:
`{{'label', 'CPUUtilization'}, {'datapoints', [[{:timestamp, {{2016, 6, 29}, {0, 3, 0}}}, {:unit, 'Percent'}, {'average', '0.0'}, {'maximum', '0.0'}], [{:timestamp, {{2016, 6, 29}, {0, 1, 0}}}, {:unit, 'Percent'}, {'average', '0.0'}, {'maximum', '0.0'}], [{:timestamp, {{2016, 6, 29}, {0, 0, 0}}}, {:unit, 'Percent'}, {'average', '0.0'}, {'maximum', '0.0'}], [{:timestamp, {{2016, 6, 29}, {0, 11, 0}}}, {:unit, 'Percent'}, {'average', '0.0'}, {'maximum', '0.0'}], [{:timestamp, {{2016, 6, 29}, {0, 6, 0}}}, {:unit, 'Percent'}, {'average', '0.0'}, {'maximum', '0.0'}], [{:timestamp, {{2016, 6, 29}, {0, 16, 0}}}, {:unit, 'Percent'}, {'average', '0.0'}, ...`

Also, should each datapoint have a list around it like this? It looks redundant, but it also looks like this is what dict:from_list wants.

src/erlcloud_mon.erl
samble: @moto I did fix it here: https://github.com/samble/erlcloud-aws-get-metric-statistics/blob/master/src/erlcloud_mon.erl#L289

I don't know why it didn't mark @nalundgaard 's comment as out of date.

motobob: sorry...did not notice.

src/erlcloud_mon.erl
motobob: yes, i would prefer 
```erlang
[{'label', 'CPUUtilization'}, {'datapoints', [...]}]
```
as it better matches the API

samble: @motobob OK this is done. I put it like you had it, with the whole thing in a list rather than a tuple.

src/erlcloud_mon.erl
```diff
@@ -216,33 +217,112 @@ put_metric_data(Namespace, MetricName, Value, Unit, Timestamp) ->
...
+  InstanceId) ->
+          get_metric_statistics(
+              "AWS/EC2",
+              MetricName,
+              StartTime,
+              EndTime,
+              60,
+              "",
+              ["Average","Maximum"],
+              [{'InstanceId', InstanceId}]).
```
waisbrot: Is `InstanceId` supposed to be an atom or a string? Looking at the spec for `get_metric_statistics/8`, it should be a string so either this line or that spec should change.

samble: Right you are @waisbrot , fixed in newest version

motobob:
@samble thank you for the PR and sorry for a lengthy review. Welcome to Erlang universe ;)

PR #293, opened 2016-07-17
add security token to config

fogfish:
_WHY_

The access to AWS resource might be managed through [third-party identity provider](http://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-idp.html). The access is managed using [temporary security credentials](http://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html). 

`erlcloud_aws:update_config(...)` is capable to update `access_key_id`, `secret_access_key` and `security_token` using instance meta-data. However, instance meta-data might not be available at development environments (outside of AWS).

We need to support API to define `AccessKeyID`, `SecretAccessKey` and  `SecurityToken` programmatically. 

_WHAT_

1. The support of `AWS_SECURITY_TOKEN` environment variable is added.

2. The `#aws_config{}` declaration is defined at `erlcloud` that provides a "service" agnostic definition of keys. This api allows also rotation of keys stored at process dictionary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,25 @@ For usage information, consult the source code and refer to the API reference at
 - http://docs.aws.amazon.com/AmazonS3/latest/API/Welcome.html
 - http://docs.amazonwebservices.com/AmazonSimpleDB/latest/DeveloperGuide/
 
+### Using Temporary Security Credentials
+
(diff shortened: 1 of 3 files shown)


motobob:
sorry for a late response as i was out on holidays. 
thanks for the contribution 👍 
I also recommend you to check out: 
https://github.com/alertlogic/erlcloud/blob/master/src/erlcloud_aws.erl#L291 
and 
https://github.com/alertlogic/erlcloud/blob/master/src/erlcloud_aws.erl#L812-L868

fogfish:
Thanks for hints!
Indeed  `auto_config()` is capable to get security token from ENV but `profile` requires improvements. It do not read security token.

Grrr,  This features is exactly what I was looking for. It kinds of made my PR redundant. Let me fix it.

PR #299, opened 2016-07-27
Fix a bug in erlcloud_ec2:register_image/2

mtaylor91:
Fixes https://github.com/erlcloud/erlcloud/issues/298


nalundgaard:
Looks like this was a copy-paste error in the implementation. Thanks for your fix.

PR #301, opened 2016-08-02
EC2 `describe_instances`: Corrects spec of return value

ephe-meral:
Discovered this while reading through the code.


motobob:
Thank you for contribution

PR #305, opened 2016-09-09
Erlang 19 dialyzer fixes for S3

waisbrot:
Dialyzer no longer adds an implicit `|undefined` to types inside records, so it rejects most of erlcloud.

I'd like to do this in multiple PRs, one per module (except for tiny modules) so that it doesn't feel hopeless to review. If you'd rather one mega-PR, I can do that instead.

This takes care of `erlcloud_s3` (starting here because this is the module that bit me), though the header file changes solve a large number of R19 dialyzer complaints across the board.

### DIFF
--- a/include/erlcloud_aws.hrl
+++ b/include/erlcloud_aws.hrl
@@ -42,7 +42,7 @@
           secret_access_key::string()|undefined|false,
           security_token=undefined::string()|undefined,
           %% epoch seconds when temporary credentials will expire
-          expiration :: pos_integer(),
+          expiration=undefined :: pos_integer()|undefined,
           %% Network request timeout; if not specifed, the default timeout will be used:
           %% ddb: 1s for initial call, 10s for subsequence;
           %% s3:delete_objects_batch/{2,3}, cloudtrail: 1s;
--- a/include/erlcloud_aws.hrl
+++ b/include/erlcloud_aws.hrl
@@ -72,16 +72,16 @@
 
           %% Read from response
           attempt = 0 :: integer(),
-          response_type :: ok | error,
-          error_type :: aws | httpc,
-          httpc_error_reason :: term(),
-          response_status :: pos_integer(),
-          response_status_line :: string(),
-          response_headers :: [{string(), string()}],
-          response_body :: binary(),
+          response_type :: ok | error | undefined,
+          error_type :: aws | httpc | undefined,
+          httpc_error_reason :: term() | undefined,
+          response_status :: pos_integer() | undefined,
+          response_status_line :: string() | undefined,
+          response_headers :: [{string(), string()}] | undefined,
+          response_body :: binary() | undefined,
 
           %% Service specific error information
-          should_retry :: boolean()
+          should_retry :: boolean() | undefined
         }).
 
 -type(aws_request() :: #aws_request{}).
--- a/src/erlcloud_s3.erl
+++ b/src/erlcloud_s3.erl
@@ -1126,25 +1126,21 @@ s3_request2_no_update(Config, Method, Host, Path, Subresource, Params, Body, Hea
                                 end
                                ]),
 
-    Request = #aws_request{service = s3, uri = RequestURI, method = Method},
-    Request2 = case Method of
-                   M when M =:= get orelse M =:= head orelse M =:= delete ->
-                       Request#aws_request{
-                         request_headers = RequestHeaders,
-                         request_body = <<>>};
-                   _ ->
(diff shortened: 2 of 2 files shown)


include/erlcloud_aws.hrl
```diff
@@ -42,7 +42,7 @@
           secret_access_key::string()|undefined|false,
           security_token=undefined::string()|undefined,
           %% epoch seconds when temporary credentials will expire
-          expiration :: pos_integer(),
+          expiration=undefined :: pos_integer()|undefined,
```
motobob: i wonder why it's defaulted only here. is this necessary? 
i guess we default it to `undefined` everywhere or nowhere, agree?

waisbrot: It's not necessary; I did it because it's in a block of others that all have an explicit default value and `security_token` right above this already had an explicit-undefined default.

I think that most of the values that get an implicit `undefined` do that from laziness, but some actually mean it. For example, all the response-related values below this in the record are definitely `undefined` until the response comes back and they get filled in.

My preference would be to make defaults explicit, because then `x=undefined::integer()|undefined` means "this can actually be undefined (because it's filled in conditionally, or there's a special meaning when it's not defined)" and `x::integer()|undefined` means "this is lazy-code and we don't actually know if `x` needs to accept a value of `undefined`.

Odds of someone carefully refactoring the whole code-base to eliminate unnecessary implicit `undefined`s are real low, though.

motobob: I agree with you that non explicit setting is more laziness and just making dializer happy without making code cleaner and thus also prefer explicit setting. 

let's leave it here as is.

motobob:
@nalundgaard any feedback?

include/erlcloud_aws.hrl
nalundgaard: I have no strong feeling on this. As long as the specs correctly say it can be undefined, knowing that this is the default value for records where the field is not set seems fine. Explicitly calling out the default is helpful but doesn't seem needed to me.

src/erlcloud_s3.erl
```diff
@@ -1126,25 +1126,21 @@ s3_request2_no_update(Config, Method, Host, Path, Subresource, Params, Body, Hea
...
-                                          RequestHeaders
-                                  end,
-                       Request#aws_request{
-                         request_headers = Headers2,
-                         request_body = Body}
-               end,
-    Request3 = erlcloud_retry:request(Config, Request2, fun s3_result_fun/1),
-    erlcloud_aws:request_to_return(Request3).
+    {RequestHeaders2, RequestBody} = case Method of
+                                         M when M =:= get orelse M =:= head orelse M =:= delete ->
```
nalundgaard: any reason not to use `;` over `orelse`?

src/erlcloud_s3.erl
```diff
@@ -1126,25 +1126,21 @@ s3_request2_no_update(Config, Method, Host, Path, Subresource, Params, Body, Hea
...
-                                      _ ->
-                                          RequestHeaders
-                                  end,
-                       Request#aws_request{
-                         request_headers = Headers2,
-                         request_body = Body}
-               end,
-    Request3 = erlcloud_retry:request(Config, Request2, fun s3_result_fun/1),
-    erlcloud_aws:request_to_return(Request3).
+    {RequestHeaders2, RequestBody} = case Method of
```
nalundgaard: Does this refactor fix a dialyzer issue, or is it just miscellaneous 'cleanup' included in the PR?

nalundgaard:
I think I am 👍 on the change. I asked a few questions above, but neither is really an issue.

src/erlcloud_s3.erl
waisbrot: The purpose of this change is to move `Request = #aws_request{...` down to the bottom of the function. 

This allows `#aws_request.request_headers` and `#aws_request.request_body` to by typed as never undefined. I may not be able to keep them not-undefined when I get to later files, so I'm happy to roll this back if it creates any doubt for you.

nalundgaard: No, this makes sense to me. 👍

src/erlcloud_s3.erl
motobob: it was there before. this is just code move.

motobob:
okay...i hear no serious arguments so 👍

src/erlcloud_s3.erl
nalundgaard: Yep. Let's merge it.

PR #306, opened 2016-09-09
Erlang 19 dialyzer fixes for WAF and SNS

waisbrot:
As with #305

### DIFF
--- a/include/erlcloud_aws.hrl
+++ b/include/erlcloud_aws.hrl
@@ -64,7 +64,7 @@
 -record(aws_request,
         {
           %% Provided by requesting service
-          service :: s3,
+          service :: s3|waf,
           uri :: string() | binary(),
           method :: atom(),
           request_headers :: [{string(), string()}],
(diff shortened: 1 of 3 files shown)


src/erlcloud_sns.erl
```diff
@@ -624,8 +624,8 @@ parse_key("EventEndpointUpdated") -> event_endpoint_updated;
 parse_key("EVentDeliveryFailure") -> event_delivery_failure;
 parse_key(OtherKey) -> list_to_atom(string:to_lower(OtherKey)).
 
-scheme_to_protocol(S) when is_list(S) -> s2p(string:to_lower(S));
-scheme_to_protocol(_)                 -> erlang:error({sns_error, badarg}).
+-spec scheme_to_protocol(string()) -> string().
+scheme_to_protocol(S) when is_list(S) -> s2p(string:to_lower(S)).
```
motobob: this change breaks the tests. 
I know that sns/s3 API with use `erlang:error` is awful but that is subject to further major revision change and not to be addressed within this PR

waisbrot:
Now I'm only touching `-spec`, `-type`, and moving the `-include`. Getting rid of _all_ warnings turned out to be impossible without some scarier refactoring in later files, so I'm going to limit my plan to just getting rid of the "no local return" complaints.

PR #309, opened 2016-09-19
cleanup couple gaps of previous merges

motobob:
Problem:
- rebar3 uses our fork , rebar2 still tries to use talk. they are identical but we need to be consistent.
- include_lib from one PRs

Solution:
just fix it. 

@nalundgaard @waisbrot fyi.  this is cosmetics but still


nalundgaard:
👍

PR #311, opened 2016-09-20
Improve on "ec2 describe_instances - correct spec of return val" (from PR #301)

paulo-ferraz-oliveira:
The {ok, _} result is actually a mapping (by means of a list comprehension) of calls to extract_reservation, which is not a proplist, but a list of proplists


motobob:
thank you for contribution.

Issue #313, opened 2016-10-06
Bad publication of 0.13.8


waisbrot:
erlcloud 0.13.8 never got published to Hex. I published it, but it turns out that my Rebar3/Hex plug-in combo was broken and so the metadata for the package is incorrect (it doesn't list any dependencies).

I sent a note to Hex asking them to take it down but maybe it's easier to just cut 0.13.9.

motobob:
i'll fix up #314 as i merged it to fast, sorry

motobob:
hm, seem to have same issue with :
```
Dependencies:
    
```
so did not publish 0.13.9 manually yet
and it's too late to revert 0.13.8
and travis still not publishing but at least building :)

continue digging...

motobob:
managed to publish manually.

waisbrot:
Motobob's publication works. 
Hex won't unpublish the bad version. They'll eventually have a feature to mark versions as deprecated for cases like this, but that doesn't exist today.

Closing this, since the current latest version is good.

PR #314, opened 2016-10-06
Travis configuration for automatically publishing releases

waisbrot:
When a build is tagged, after tests pass for the "main" build (Erlang 19 + Rebar3), this attmpts to publish the tag to Hex.

Hopefully this makes #313 not happen again.

This is configured to use my credentials. We can swap in someone else's or a bot's pretty easily.

I mostly copied this from another project of mine where it's working, but it's painful to debug if it doesn't work the first time (because you have to keep publishing releases).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,3 +107,22 @@ else
 	wget https://s3.amazonaws.com/rebar3/rebar3
 	chmod a+x rebar3
 endif
+
+travis-publish:
+	@echo Create directories
+	mkdir -p ~/.hex
+	mkdir -p ~/.config/rebar3
+
+	@echo Decrypt secrets
+	@openssl aes-256-cbc -K $encrypted_9abc06b32f03_key -iv $encrypted_9abc06b32f03_iv -in hex.config.enc -out ~/.hex/hex.config -d
+
+	@echo Create global config
+	echo '{plugins, [rebar3_hex]}.' > ~/.config/rebar3/rebar.config
+
+	@echo Edit version tag in app.src
+	vi -e -c '%s/{vsn, *.*}/{vsn, "'${TRAVIS_TAG}'"}/g|w|q' src/dogstatsd.app.src
+
+	@echo Publish to Hex
+	echo 'Y' | ./vendor/rebar3 hex publish
+
+	@echo Done
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ otp_release:
   - 18.1
   - 19.0
 env:
+  global:
+  - MAIN_OTP = 19.0
+  matrix:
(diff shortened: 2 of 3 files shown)


motobob:
i knew it's possible to do it via travis. 👍 @waisbrot 
merging this and going to tag it right away to 0.13.9

Makefile
```diff
@@ -107,3 +107,22 @@ else
...
+	mkdir -p ~/.config/rebar3
+
+	@echo Decrypt secrets
+	@openssl aes-256-cbc -K $encrypted_9abc06b32f03_key -iv $encrypted_9abc06b32f03_iv -in hex.config.enc -out ~/.hex/hex.config -d
+
+	@echo Create global config
+	echo '{plugins, [rebar3_hex]}.' > ~/.config/rebar3/rebar.config
+
+	@echo Edit version tag in app.src
+	vi -e -c '%s/{vsn, *.*}/{vsn, "'${TRAVIS_TAG}'"}/g|w|q' src/dogstatsd.app.src
```
motobob: @waisbrot oops...merged too fast as  i guess that should be `erlcloud.app.src`

Makefile
```diff
@@ -107,3 +107,22 @@ else
...
+travis-publish:
+	@echo Create directories
+	mkdir -p ~/.hex
+	mkdir -p ~/.config/rebar3
+
+	@echo Decrypt secrets
+	@openssl aes-256-cbc -K $encrypted_9abc06b32f03_key -iv $encrypted_9abc06b32f03_iv -in hex.config.enc -out ~/.hex/hex.config -d
+
+	@echo Create global config
+	echo '{plugins, [rebar3_hex]}.' > ~/.config/rebar3/rebar.config
```
motobob: should we may be just update rebar.config right away?

Makefile
```diff
@@ -107,3 +107,22 @@ else
...
+
+	@echo Edit version tag in app.src
+	vi -e -c '%s/{vsn, *.*}/{vsn, "'${TRAVIS_TAG}'"}/g|w|q' src/dogstatsd.app.src
+
+	@echo Publish to Hex
+	echo 'Y' | ./vendor/rebar3 hex publish
```
motobob: should be ./rebar3

Makefile
waisbrot: Yup. Sorry, too much copy/paste from me.

Makefile
waisbrot: I'm actually putting this into the _global_ rebar3 config, because I don't want ercloud's dependencies to include it. I may be overthinking things, though; I'm not sure what rebar actually does if there are plugin dependencies listed in a Hex dependency.

PR #315, opened 2016-10-06
Add put_records support

arkdro:
This PR adds AWS put_records api support.

### DIFF
--- a/src/erlcloud_kinesis.erl
+++ b/src/erlcloud_kinesis.erl
@@ -12,6 +12,7 @@
          get_shard_iterator/3, get_shard_iterator/4, get_shard_iterator/5,
          get_records/1, get_records/2, get_records/3,
          put_record/3, put_record/4, put_record/5, put_record/6,
+         put_records/2, put_records/3,
          merge_shards/3, merge_shards/4,
          split_shards/3, split_shards/4
         ]).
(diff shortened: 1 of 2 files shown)


src/erlcloud_kinesis.erl
```diff
@@ -406,6 +407,57 @@ put_record(StreamName, PartitionKey, Data, ExplicitHashKey, Ordering, Config) wh
...
+%%------------------------------------------------------------------------------
+%% Kinesis API:
+%% [https://docs.aws.amazon.com/kinesis/latest/APIReference/API_PutRecords.html]
+%%
+%% ===Example===
+%%
+%% This operation puts a data record into an Amazon Kinesis stream from a producer.
+%%
+%% `
+%% erlcloud_kinesis:put_records(<<"test">>, [{<<"asdasd">>, <<"asdasd">>, <<"asdasd">>}]).
```
nalundgaard: This doesn't seem like an acceptable example. Please follow the style of the existing examples in this module.

src/erlcloud_kinesis.erl
```diff
@@ -406,6 +407,57 @@ put_record(StreamName, PartitionKey, Data, ExplicitHashKey, Ordering, Config) wh
...
+-spec put_records(string(), Items, Config) -> proplist() when
+      Items :: [Item],
+      Item :: {string(), string()}
+            | {string(), string(), string()},
+      Config :: aws_config().
+
+put_records(StreamName, Items, Config) ->
+    put_records_common(StreamName, Items, Config).
+
+put_records_common(StreamName, Items, Config) ->
```
nalundgaard: Why? Why not just do this in put_records/3, and make put_records/2 call it? Seems like useless abstraction.

arkdro:
are you sure about squashing?

nalundgaard:
I think squashing makes the git history a lot easier to understand, but I may be alone on that. @motobob any feelings on that?

arkdro:
about squashing - is it on your side or on mine?

src/erlcloud_kinesis.erl
```diff
@@ -406,6 +407,73 @@ put_record(StreamName, PartitionKey, Data, ExplicitHashKey, Ordering, Config) wh
...
+%%     ]
+%% }
+%% '
+%%
+%% @end
+%%------------------------------------------------------------------------------
+
+-spec put_records(string(), Items) -> proplist() when
+      Items :: [Item],
+      Item :: {string(), string()}
```
nalundgaard: My issue with the example is maybe a microcosm of the fact that you have to read the code for prepare_put_records_data, etc. to understand the form of the input. It's not clear what the order is supposed to be here—what element of the tuple is data, hash key, partition key?

Might be more helpful to make types, although I guess it's not done in general in this module. e.g.:

``` erlang
-type put_records_item() :: {Data :: string(),  PartitionKey :: string()} |
                            {Data :: string(), ExplicitHashKey :: string(), PartitionKey :: string()}.
-type put_records_items() :: [put_records_item()].
```

Then you can just have

```erlang
spec put_records(string(), put_records_items()) -> proplist().
...
spec put_records(string(), put_records_items(), aws_config()) -> proplist().
```

seems more succinct and easy to read.

As an aside, I notice that your spec calls for strings, but your examples are all binary—which are *not* strings. I am also not 100% sure that it would even work with strings. It seems like this might be a larger problem with the authorship of this module... but shouldn't all these instances of `string()` in the specs actually be `binary()`?

nalundgaard:
On squashing, there are [lots of ways](https://www.google.com/search?q=git+squash+commits) for you to do it. You would need to squash all your commits into one locally, then force-push your commit up to `arkdro:put_records_api`. This would let you make a commit with a succinct commit message that would reflect the feature you have added. We can also squash + merge this on our side, but it will just roll up your many commit messages into a multiline summary.

waisbrot:
At work (on private repos) I've been loving the "Squash and merge" button on pull requests. I'm not thrilled that it strips my GPG signature off, but no longer needing to show my coworkers how to squash a commit is a big win.

If you Alert Logic folks don't feel strongly about it, I think using the GitHub interface to perform the squash ourselves would be a nice way to make outside contributions easier. (And if you want to assign me to write a better summary for the final commit, I don't mind doing that as a regular chore.)

nalundgaard:
@waisbrot Good thought. That sounds fine to me. @arkdro Up to you whether you want to learn to squash yourself. I made some comments about improving type specs above, and I also still think that your use of `<<"asdasd">>` in your documented examples for the Data, HashKey, and PartitionKey elements of the example item is unhelpful.

arkdro:
squashing in its current form is error prone. Nevertheless.

nalundgaard:
👍 from me. The `string()` input spec is totally wrong, but at least it matches the larger repo. That should really be addressed as a separate PR.

src/erlcloud_kinesis.erl
```diff
@@ -406,6 +407,71 @@ put_record(StreamName, PartitionKey, Data, ExplicitHashKey, Ordering, Config) wh
...
+put_records(StreamName, Items, Config) ->
+    Operation = put_records_operation(),
+    Json = prepare_put_records_data(StreamName, Items),
+    erlcloud_kinesis_impl:request(Config, Operation, Json).
+
+put_records_operation() ->
+    "Kinesis_20131202.PutRecords".
+
+prepare_put_records_data(StreamName, Items) ->
+    Records = [prepare_put_records_item(X) || X <- Items],
```
motobob: @arkdro @nalundgaard API allows to put only 500 records in the batch? 
should we protect naive user by any means or let it fail with extremely descriptive `400 - InvalidArgumentException` 
in general in the library, we tent to protect, including this module.

arkdro:
I'd prefer go without the limit here. For example: in this very module there are 100 items guards for get_records api. But the documentation at the amazon site says that current limit is 10000.

src/erlcloud_kinesis.erl
nalundgaard: I prefer to let the API do input validation like that, generally. If they ever change it, then we're broken.

motobob:
@arkdro okay I fine putting it as as with no limit on number of records. 
And I'll send follow up PRs for fixing currently broken limits.

Also, it would be probably nice to have smth like `put_records_batch(StreamName, Items, Config)` where `Items` can be bigger than 500 and library will magically split it into batches or max size.

motobob:
on the subject of rotten limits in kinesis module we have it [fixed](https://github.com/alertlogic/erlcloud/blob/master/src/erlcloud_kinesis.erl#L89) 
one more call for a single upstream merge which is now going to happen really soon.

PR #316, opened 2016-10-06
fix publish

motobob:
fix forward of too fast merge of #314.


motobob:
and #313

waisbrot:
Thanks for fixing that!

PR #317, opened 2016-10-06
Adjust the branches/only property to include anything that looks like a semver

waisbrot:
#314 didn't produce a Travis build when @motobob tagged it. I think this is why: the "branches" property is really refs and so Travis doesn't see the tag as a build candidate. On my other project, I don't limit to building one branch and that's the only difference I'm seeing.

Unless you want to tag another release to try again, we should just manually push 0.13.9.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ env:
 branches:
   only:
     - master
+    - /^[0-9]+\.[0-9]+\.[0-9]+/
 install:
   - make travis-install
 script: make eunit


motobob:
i don't think it's that 
```sh
mbr13_ebob:erlcloud_upstream ebob$ make travis-publish
Create directories
mkdir -p ~/.hex
mkdir -p ~/.config/rebar3
Decrypt secrets
non-hex digit
invalid hex iv value
make: *** [travis-publish] Error 1
```

let me play a little deeper before further fast merges

waisbrot:
If you're running this locally, it should not work because you don't have the magic Travis variables `$encrypted_9abc06b32f03_key` and `$encrypted_9abc06b32f03_iv` set.

motobob:
okay looks sane and according to the docs

motobob:
just linking #313 .

okay this time it have built it but no publish...hm...
screen shot 2016-10-06 at 21 57 48

waisbrot:
From the logs, it doesn't look like it tried to run the deploy script. I say 

```
condition: $FORCE_REBAR2 = false && $TRAVIS_OTP_RELEASE = $MAIN_OTP
```

so my guess is that one of those two is not what I expected it to be. I can bump around using pull requests, so we won't have to do a release for _every_ debug cycle :-P

PR #318, opened 2016-10-06
Another try at automatic Hex deploys

waisbrot:
It was close to working already, but the spaces around the `=` in environment variable assignment made Travis skip deploying on _all_ branches when I'd only meant to skip everything but 19.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ otp_release:
   - 19.0
 env:
   global:
-  - MAIN_OTP = 19.0
+  - MAIN_OTP=19.0
   matrix:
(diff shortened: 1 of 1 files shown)


waisbrot:
@motobob 

```
Decrypt secrets
non-hex digit
invalid hex iv value
```

:'(

That's a pretty puzzling one to me. I think it means the openssl command failed, but I have no idea why it would.

motobob:
yeah...i'm lost too. lovely "works on my machine" 
unfortunately, i can't focus on this during a week due to other stuff on me and I really need to finish review of ECS. 
I'll come back to that later

PR #319, opened 2016-10-07
IMPORTANT: update on proposed library roadmap

motobob:
Hello Erlcloud community, 

The library has significantly diverged between the forks and we would like to fix this situation especially given the fact that upstream is missing quite a number or major functionality.

It's not a secret that upstream and Alert Logic forks are mostly maintained by the same set of people. Therefore, we would like to propose a library major revision update which will be basically almost as FF merge to https://github.com/alertlogic/erlcloud 
Please refer to Roadmap section of Readme for details.

This will bring lots of stuff (Initial attempt was back at #215 ):
- Full support for SigV4 across all modules. According to AWS, all services support SigV4 in all regions and only a few still have SigV2 therefore SigV2 funs were not removed if smb still needs them
- S3 follow redirect option was added
- KMS update, basically reworked #182
- DirectConnect described APIs added
- RDS APIs added
- Redshift API
- Route53 added
- erlcloud_autoscalling added along with some edits erlcloud_as to have backward compatibility
- Large number of maintenance defect fixes here and there
- enhancements to DDB2, EC2 (adding all the gateways) and Kinesis
- enriching ELB APIs
- Adding AWS Inspector API
- adding more testing

**Note**: 
- No API modules or functions were removed, our fork only adds or updates the APIs. 
- Several APIs, like ELB have minor incompatibilities.

@smilerlee @waisbrot @nalundgaard @ptrakhtman @okeuday @ddosia @fogfish please weigh in.

Unless I hear back extremely negative feedback with detailed explanation of blockers, I'll perform this in by Oct20, IMO two weeks warning should be enough. 

P.S: "my application is pinned to master branch" is not accepted as a reasonable excuse for not progressing the library further.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,35 +1,75 @@
-# erlcloud: Cloud Computing APIs For Erlang #
+# erlcloud: AWS APIs library for Erlang #
 
 [![Build Status](https://secure.travis-ci.org/erlcloud/erlcloud.png?branch=master)](http://travis-ci.org/erlcloud/erlcloud)
 
-Service APIs implemented:
+This library is not developed or maintained by AWS thus lots of functionality is still missing comparing to [aws-cli](https://aws.amazon.com/cli/) or [boto](https://github.com/boto/boto).
+Required functionality is being added upon request.
 
+Service APIs implemented:
 - Amazon Elastic Compute Cloud (EC2)
 - Amazon Simple Storage Service (S3)
 - Amazon Simple Queue Service (SQS)
-- Amazon SimpleDB
-- Amazon Mechanical Turk
-- Amazon CloudWatch (MON)
-- Amazon DynamoDB (ddb2) 
+- Amazon DynamoDB & DDB streams (ddb2)
+- Amazon Autoscalling(as)
+- Amazon CloudTrail (CT)
+- Cloud Formation (CFN)
+- ElasticLoadBalancing (ELB)
+- Identity and Access Management (IAM)
+- Kinesis
+- CloudWatch
+- MechanicalTurk
+- Simple DB (SDB)
+- Relational Data Service (RDS)
+- Simple Email Service (SES)
+- Short Token Service (STS)
+- Simple Notification Service (SNS)
+- Web Application Firewall (WAF)
+- and more to come
 
-All API functions have been implemented.  Not all functions have been thoroughly tested, so exercise care when integrating this library into production code.  Please send bug reports and patches.
+Majority of API functions have been implemented.
+Not all functions have been thoroughly tested, so exercise care when integrating this library into production code.
+Please send issues and patches.
 
-The libraries can be used two ways: either you can specify configuration parameters in the process dictionary, or you can create a configuration object and pass that to each request as the final parameter.
+The libraries can be used two ways:
+- either you can specify configuration parameters in the process dictionary. Useful for simple tasks
+- you can create a configuration object and pass that to each request as the final parameter. Useful for Cross AWS Account access
+
+## Roadmap ##
+
+Below is the proposed library roadmap update along with regular features and fixes.
+
+- 0.13.10
+ * Existing code
+
+- 2.0.X
+ * Going further we would like to merge in [Alert Logic](https://github.com/alertlogic/erlcloud/tree/v1.2.3) fork into upstream.
+ This is a major version bump which contains lots of new features and functionality.
+ Unfortunately, it also contains quite a number of low level APIs incompatibilities since the fork diverged for a long while.
+ Making it backward compatible does not seem feasible and valuable at the moment.
+ * No APIs have been removed and it's on branched of 0.13.10 at the moment. Any minor version delta added during notice time will be compensated before the merge.
+ * intentionally jumping to 2.0.X as AL fork has v1.0.X
+
+- 2.1.X
+ * fix dialyzer findings and make it mandatory for the library
+ * make full support of Mix/HEX
+
+- 2.2.X
+  * Further deprecation of legacy functionality
+  * remove R14/R15 support.
+  * Only SigV4 signing and generalised in one module. Keep SigV2 in SBD section only
+  * TBD: no more `erlang:error()` use and use of regular tuples as error API.
+
+### Major API compatability changes between 0.13.X and 2.0.x
+ - ELB APIs
+ - ... list to be filled shortly
 
 ## Getting started ##
-You need to clone the repository and download rebar (if it's not already available in your path).
-
+You need to clone the repository and download rebar/rebar3 (if it's not already available in your path).
 ```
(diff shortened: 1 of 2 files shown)


motobob:
@waisbrot note updated. sorry  I meant s/reacher/richer :)

README.md
```diff
@@ -1,58 +1,95 @@
...
-You need to clone the repository and download rebar (if it's not already available in your path).
+- v0.13.9
+ * Existing code
 
-```
-git clone https://github.com/erlcloud/erlcloud.git
-cd erlcloud
-wget http://cloud.github.com/downloads/basho/rebar/rebar && chmod u+x rebar
-```
+- v2.0.X
```
nalundgaard: Why 2.0.X and not 1.0.X? Any reason for skipping the incrementation here?

README.md
```diff
@@ -1,58 +1,95 @@
...
+ * Going further we would like to merge in [Alert Logic](https://github.com/alertlogic/erlcloud/tree/v1.2.3) fork.
+ This is a major version bump which contains lots of new features and functionality.
+ Unfortunately, it also contains quite a number of low level APIs incompatibilities since the fork diverged for a long while.
+ No APIs have been removed and it's on top or 0.13.8 at the moment.
```
nalundgaard: "it's on top of" or "it's branched from"

README.md
nalundgaard: For that matter, if the alertlogic/erlcloud tags are compatible with (ahead of) the erlcloud/erlcloud tags, is there a good reason not to import them?

README.md
```diff
@@ -1,58 +1,95 @@
...
 
-You can provide your amazon credentials in environmental variables.
+- v2.1.X
+ * fix dialyzer findings and make it mandatory for the library
+ * make full support of Mix/HEX
 
+- v2.2.X
+ * Further deprecation of legacy functionality
+  * Only SigV4 signing and generalised in one module. Keep SigV2 in SBD section only
+  * no more `erlang:error()` use and use of regular tuples as error API.
```
nalundgaard: This sounds like this should entail another major version increment, as it is a substantial behavior change that will again break people's code. I'd make this 3.0.X (or 2.0.X if we decrement the other instances of 2.*.X).

README.md
```diff
@@ -1,58 +1,95 @@
...
-If you're using erlcloud in your application, add it as a dependency in your application's configuration file.  To use erlcloud in the shell, you can start it by calling:
+If you're using erlcloud in your application, add it as a dependency in your application's configuration file.
+To use erlcloud in the shell, you can start it by calling:
 
 ```
-ssl:start().
-erlcloud:start().
+application:ensure_all_started(erlcloud).
 ```
+You can provide your amazon credentials in environmental variables.
```
nalundgaard: Is it worth also mentioning the support for profiles added by @fp here? That stuff is super cool.

README.md
```diff
@@ -84,43 +121,32 @@ start_instance(Ami, KeyPair, UserData, Type, Zone) ->
...
+You can provide your amazon credentials in environmental variables
 
 ```
 export AWS_ACCESS_KEY_ID=<Your AWS Access Key>
 export AWS_SECRET_ACCESS_KEY=<Your AWS Secret Access Key>
 export AWS_SECURITY_TOKEN=<Your AWS Security Token>
 ```
-
-
 If you did not provide your amazon credentials in the environmental variables, then you need to provide configuration read from your profile:
```
nalundgaard: Per my other comment, it seems like this is worth leading with. I doubt anyone wants to use shell variables given there is support for profiles.

fogfish:
Existence of single branch definetly healthy for community. I am not going to discuss why such fragmentation appears. You have to consider couple of things as part of upstream movement to alertlogic.

1. erlcloud upstream has more starts (it means that it is bookmarked by higher volume of developers).  You shall transfer project from erlcloud account to alertlogic account. You cannot just announce that alertlogic is upstream

2. What is your exit plan? I saw many project maintained by biz entity (e.g. rebar) who dropped them later and moved it to community. They started development in biz account (e.g. basho) and later moved it to community account (e.g. rebar). Do you see such risks here?  I've seen already 3 ownership transfers for this project. Should we just keep and maintain stable community?

README.md
motobob: the reason is we have v1.X.X in AL fork.
I was planning to import laster AL tag into here and even publish it to HEX once.

README.md
motobob: it is mentioned below in Temp credentials section.

README.md
motobob: I'll update the section to give preference to profile() and new() Config options instead of evn variable. however i feel that is still valuable functionality.

motobob:
@fogfish 
>You cannot just announce that alertlogic is upstream

this is **not** the proposal here, sorry for not being super verbose. It's actually the opposite. 
https://github.com/erlcloud/erlcloud/ is the upstream and we are **not** going to change it. It stay the upstream. 
I plan to git merge the changes from AL fork to here. Also, I plan actually work inside the company to not divert AL fork any further and use upstream for further development. Probably, i'll even close for write AL fork.

We do plan to continue maintaining https://github.com/erlcloud/erlcloud/ as a single and stable community upstream. 
hope that clarifies the concern.

smilerlee:
> which will be basically almost as FF merge to https://github.com/alertlogic/erlcloud

I suppose `--no-ff` merging is preferable.

> Several APIs, like ELB have minor incompatibilities.

If possible, we could add detailed notes in README to reveal the incompatibilities between 2.x and 0.9.x.

fogfish:
@motobob 👍 

No objections then! As I said this is an excellent idea to align.

motobob:
@smilerlee i'll do my best to capture the API delta.

I'll keep this open for a while until i actually perform the necessary merges (Oct17-20) so that rest of library consumers could be prepared.

waisbrot:
Copying my comment over to here: I'd like to see the roadmap include plans to support/deprecate older versions of Erlang/OTP. 

R16 was updated in 2014 and the introduction of maps was a pretty substantial change for the code I work with (that is, maps have been a nice improvement and we upgraded to Erlang/OTP 18 as soon as it was available to get full support for maps).

I don't have a good sense of how many Erlangs back Erlcloud users expect to be able to use, but I think it would be nice to say "we support the most recent $n$ revisions" or "we support the most recent $n$ years".

For the users I represent (a couple of us at WHOOP), I'd say "any Erlang released in the past 2 calendar years", which would include 17.5 until 2017 rolls around. But we have no legacy code so that may be too aggressive for most users.

motobob:
i'm happy to deprecate R14/15. Unfortunately, we still use R16 to somewhat heavily atm.
nevertheless, I'll add that to roadmap.

motobob:
merging this and will list of breaking possible breaking changes right before performing actual merge.

PR #326, opened 2016-10-19
adhere to zero warn policy

motobob:
Problem:
there are warnings in many places

Solution:
adhere to zero warning policy

notes: 
- there is lot of copy paste in tests to be removed, separately.
- one `[export all]` left in erlcloud_sdb_tests as that requires special care and to be done separately 

@nalundgaard  @ddosia


src/erlcloud_ddb_util.erl
```diff
@@ -25,13 +25,10 @@
...
          q_all/2, q_all/3, q_all/4,
          scan_all/1, scan_all/2, scan_all/3,
-         write_all/2, write_all/3, write_all/4
+         write_all/2, write_all/3, write_all/4,
+         set_out_opt/1
         ]).
 
-%% Export all functions for unit tests
--ifdef(TEST).
--compile(export_all).
```
nalundgaard: I would suggest

```erlang
%% Export private functions for unit testing
-ifdef(TEST).
-export([set_out_opt/1]).
-endif.
```

src/erlcloud_ddb_util.erl
```diff
@@ -455,9 +452,6 @@ write_all_result([{error, Reason} | _]) ->
 write_all_result([]) ->
     ok.
 
-%%%------------------------------------------------------------------------------
-%%% Internal Functions
```
nalundgaard: which would make this unneeded.

motobob:
agree, that is nicer. 
PR amended.

PR #327, opened 2016-10-20
Check warnings as part of the Travis tests

waisbrot:
Run a build under a profile with `warnings_as_errors` on during tests

Related to #326

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,9 @@ branches:
     - /^[0-9]+\.[0-9]+\.[0-9]+/
 install:
   - make travis-install
-script: make eunit
+script:
+  - make check_warnings
+  - make eunit
 deploy:
   skip_cleanup: true
   provider: script
--- a/Makefile
+++ b/Makefile
@@ -63,6 +63,13 @@ else
 	$(REBAR) shell
 endif
 
+check_warnings:
+ifeq ($(REBAR_VSN),2)
+	@echo skip checking warnings
+else
+	@$(REBAR) as warnings compile
+endif
+
 eunit:
 ifeq ($(REBAR_VSN),2)
 	@$(REBAR) compile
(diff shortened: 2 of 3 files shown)


nalundgaard:
I dismissed my review, looking closer. There are several warnings, and the build still passes. I am curious why?

waisbrot:
Can you be more specific? I see a few things:

- There are some warnings from Rebar, but those aren't Erlang compiler warnings so they aren't part of the warnings_as_errors directive
- There is a warning under Erlang 19 when compiling lhttpc. The compiler options are not propagated down to dependences (nor, I think, would we want them to be)
- There is a warning when running eunit because `erlcloud_sdb_tests.erl` uses `export_all`. I'd wanted to attach warnings_as_errors to the `test` profile, but I don't know how to override the setting just for that file (seems OK to me to use export_all in tests)

nalundgaard:
I think the latter one was confusing me. I guess we don't fail warnings for the test steps. I must have misread the output. I agree, that should be fine.

waisbrot:
Yeah, rebar3 does a totally separate build for each profile. So the current tests build everything under `_build/test` and this PR means we do a _second_ build under `_build/warnings`. If you run `rebar3 dialyzer` it'll make a third build in `_build/default`.

If you put some warning-generating code into a regular module with an `-ifdef(TEST).`, it would not fail the build because under the `warnings` profile the code is skipped and under the `test` profile it'll produce a warning but not have warnings_as_errors

I could have the Makefile write a banner or something to separate the "are there any warnings?" check from the "does it pass tests?" check more clearly. Would that be useful?

nalundgaard:
I mean, as long as "real" warnings are caught, I am 👍 on it.

motobob:
@waisbrot thanks, that was literally my next thing :)

PR #329, opened 2016-11-01
handle another http error codes

michalzajda:
it will nicely format bunch of other errors, some covered by amazon business semantic like 404 - no target ARN


motobob:
AWS API is awesome and HTTP code does not necessarily match Code inside XML
a typical case here `400 Bad Request "InstanceID.NotFound"` which could be a more friendly 404.
however it's will be a API breaking change to add HttpErr into the output. Before we change SNS module to not to throw `erlang:error()` this change seems fine to me

michalzajda:
Thanks. I hit a problem using SNS API. When calling GetEndpointAttributes it gives 404 and its perfectly fine - resource not found.

PR #331, opened 2016-11-06
Export service_host/2 function

fp:


waisbrot:
Any particular reason why you'd like it exported?

motobob:
i suspect to gain access to https://github.com/fp/erlcloud/blob/ac1435e2ad8039b8c57f890031469c84b06a6ee4/src/erlcloud_aws.erl#L494-L502 from `erlcloud_s3` side. 

This should be no harm but a brief description would be valuable.

PR #332, opened 2016-11-14
Feature/support sts use

posilva:
- This feature allow to configure an assume_role setting to use when update_config
- We just need to fill `aws_config` record with `assume_role` options: 
```erlang
#aws_config{
...
assume_role = [
       {role_arn, "arn:aws:iam::123456789012:role/marketingadmin"}, %% Required
       {session_name, "asessionname"}, %% Optional with default: "erlcloud"
       {session_expiration, 900}, %% Optional with default:900
]}
`

### DIFF
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -613,12 +613,14 @@ get_credentials_from_role(#aws_config{assume_role = AssumeRoleOptions} = Config)
                                       "erlcloud"),
     SessionExpiration = proplists:get_value(session_expiration,
                                             AssumeRoleOptions, 900),
+    ExternalId = proplists:get_value(external_id, AssumeRoleOptions,
+                                      undefined),
     %% We have to remove the assume role options to make sure we do not
     %% enter in a infinite loop because erlcloud_sts:assume_role also calls
     %% update_config deep inside when makes the request
     case catch erlcloud_sts:assume_role(Config#aws_config{assume_role = []},
                                         RoleArn, SessionName,
-                                        SessionExpiration) of
+                                        SessionExpiration, ExternalId) of
         {#aws_config{}=_NewConfig, Creds} ->
             ExpireAt = calendar:datetime_to_gregorian_seconds(
                 proplists:get_value(expiration, Creds)),
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -381,14 +381,8 @@ config_metadata() ->
 
 
 -spec update_config(aws_config()) -> {ok, aws_config()} | {error, term()}.
-update_config(#aws_config{access_key_id = KeyId} = Config)
-  when is_list(KeyId), KeyId /= [] ->
-    %% In order to support caching of the aws_config, we could store the expiration_time
-    %% and check it here. If it is about to expire (within 5 minutes is what boto uses)
-    %% then we should get the new config.
-    {ok, Config};
-update_config(#aws_config{assume_role = AssumeRoleOptions} = Config)
-    when  is_list(AssumeRoleOptions), AssumeRoleOptions /= [] ->
+update_config(#aws_config{assume_role = [{_,_}|_] = AssumeRoleOptions} = Config)
+    when is_list(AssumeRoleOptions) ->
     %% The assume role options are defined lets try to assume a role
     case get_role_credentials(Config) of
         {error, Reason} ->
@@ -399,6 +393,12 @@ update_config(#aws_config{assume_role = AssumeRoleOptions} = Config)
                 secret_access_key = Credentials#role_credentials.secret_access_key,
                 security_token = Credentials#role_credentials.session_token}}
     end;
+update_config(#aws_config{access_key_id = KeyId} = Config)
+  when is_list(KeyId), KeyId /= [] ->
+    %% In order to support caching of the aws_config, we could store the expiration_time
+    %% and check it here. If it is about to expire (within 5 minutes is what boto uses)
+    %% then we should get the new config.
+    {ok, Config};
 update_config(#aws_config{} = Config) ->
     %% AccessKey is not set. Try to read from role metadata.
     case get_metadata_credentials(Config) of
@@ -591,12 +591,13 @@ prop_to_list_defined( Name, Props ) ->
 
 
 -spec get_role_credentials(aws_config()) -> {ok, #role_credentials{}} | {error, term()}.
-get_role_credentials(Config) ->
-    case application:get_env(erlcloud, role_credentials) of
+get_role_credentials(#aws_config{assume_role = AssumeRoleOptions} = Config) ->
+    RoleArn = proplists:get_value(role_arn, AssumeRoleOptions, undefined),
+    case application:get_env(erlcloud, {role_credentials, RoleArn}) of
         {ok, #role_credentials{expiration_gregorian_seconds = Expiration} = Credentials} ->
             Now = calendar:datetime_to_gregorian_seconds(calendar:universal_time()),
-            %% Get new credentials if these will expire in less than 5 minutes
-            case Expiration - Now < 300 of
+            %% Get new credentials if these will expire in less than 2 minutes
+            case Expiration - Now < 120 of
                 true -> get_credentials_from_role(Config);
                 false -> {ok, Credentials}
             end;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by motobob** [comment]
it seems this supports only one level of assume role but there could be nested assumes, like Account1->Account2->Account3->.. and this will fail on renewal.

File: src/erlcloud_aws.erl [hunk 1]
```diff
@@ -574,6 +590,42 @@ prop_to_list_defined( Name, Props ) ->
...
+        {#aws_config{}=_NewConfig, Creds} ->
+            Record = #role_credentials{
+                access_key_id =  proplists:get_value(access_key_id, Creds),
+                secret_access_key = proplists:get_value(secret_access_key, Creds),
+                session_token = proplists:get_value(session_token, Creds),
+                expiration_gregorian_seconds = timestamp_to_gregorian_seconds(
+                    proplists:get_value(expiration, Creds))},
+            application:set_env(erlcloud, role_credentials, Record),
+            {ok, Record};
+        {'EXIT', Reason} ->
```
motobob: i would rather protect it from being called with `undefined`

motobob:
populating role details when loading from aws profiles is missing an update along with this change.

src/erlcloud_aws.erl
```diff
@@ -574,6 +590,42 @@ prop_to_list_defined( Name, Props ) ->
...
+    case catch erlcloud_sts:assume_role(Config, RoleArn, SessionName,
+                                        SessionExpiration) of
+        {#aws_config{}=_NewConfig, Creds} ->
+            Record = #role_credentials{
+                access_key_id =  proplists:get_value(access_key_id, Creds),
+                secret_access_key = proplists:get_value(secret_access_key, Creds),
+                session_token = proplists:get_value(session_token, Creds),
+                expiration_gregorian_seconds = timestamp_to_gregorian_seconds(
+                    proplists:get_value(expiration, Creds))},
+            application:set_env(erlcloud, role_credentials, Record),
```
posilva: Library already uses env to cache metadata_credentials. I see your point of multiple roles being assumed,  we can cache in a "dictionary" based in the role_arn.  The idea is to allow to a assume a role  per request and once you assume a role the library will cache it for an amount of time per assumed role. This is different from the role_arn in the profile file.

This use case is  for applications that run in ec2 instances and need to assume a role and do not have the profile file, just the policy attached to the IAM Role of the instance. It's true that you may have a cascade of assume role but you do not have a good way of express that to be used in the library. My Idea is to keep it simple and do it in a per request base where you can change the assume_role (not implemented yet)

motobob:
@posilva to make this be merged let's at least fix the current use case of using multiple Creds per application in different processes (not use `set_env`  and only `#aws_config{}`?) and i would agree to deal with nested assume later in a separate PRs

posilva:
@motobob I think we should use set_env just for cache sts_credentials as used in the `metadata_credentials` cache, otherwise we will pay for the sts_assume call in every request which is costly in terms of latency. 
We can use `application:set_env(erlcloud, {sts_credentials, RoleArn}, STSCredentialsRec).` Of course if we have several RoleARN per request will pollute the application_env, but the common use case should not be that one.

This approach should fix this:
> more over is used heavily in the mode where multiple roles being assumed from a given application. this will break things.

because each Role will have is cache key and will "refresh" independently from each other.

What do you think?

posilva:
I updated the PR with a multiple RoleARN solution. I hope this will be enough. I tested locally in my machine making calls to a DDB table in a different AWS Account and works pretty good. I can use one RoleARN to call DDB tables in one account and other Role ARN to get objects from an S3 bucket.

No support to nested AssumeRoles, but I have some ideas :), just need to check better the docs for that cases (hmmm, if these cases exists or make sense when we are using an AWS SDK)

posilva:
If you are willing to merge the PR I will squash the branch in one commit to have a cleaner history :)

motobob:
@posilva i'll review new version in couple days.

src/erlcloud_aws.erl
```diff
@@ -374,6 +381,18 @@ config_metadata() ->
 
 
 -spec update_config(aws_config()) -> {ok, aws_config()} | {error, term()}.
+update_config(#aws_config{assume_role = [{_,_}|_] = AssumeRoleOptions} = Config)
+    when is_list(AssumeRoleOptions) ->
```
motobob: guard is unnecessary given the match for at least one element

src/erlcloud_aws.erl
```diff
@@ -574,6 +590,51 @@ prop_to_list_defined( Name, Props ) ->
...
+            ExpireAt = calendar:datetime_to_gregorian_seconds(
+                proplists:get_value(expiration, Creds)),
+            Record = #role_credentials{
+                access_key_id =  proplists:get_value(access_key_id, Creds),
+                secret_access_key = proplists:get_value(secret_access_key, Creds),
+                session_token = proplists:get_value(session_token, Creds),
+                expiration_gregorian_seconds = ExpireAt},
+            application:set_env(erlcloud, {role_credentials, RoleArn}, Record),
+            {ok, Record};
+        {'EXIT', Reason} ->
```
motobob: previous comment not addressed, let's protect for calling the fun with `undefined` and not do `catch`

src/erlcloud_aws.erl
```diff
@@ -574,6 +590,51 @@ prop_to_list_defined( Name, Props ) ->
...
+                false -> {ok, Credentials}
+            end;
+        undefined ->
+            get_credentials_from_role(Config)
+    end.
+
+-spec get_credentials_from_role(aws_config()) -> {ok, #role_credentials{}} |
+                                                 {error, term()}.
+get_credentials_from_role(#aws_config{assume_role = AssumeRoleOptions} = Config) ->
+    RoleArn = proplists:get_value(role_arn, AssumeRoleOptions, undefined),
```
motobob: it seems to me we would better use a record here for `assume_role `, no?

src/erlcloud_aws.erl
```diff
@@ -574,6 +590,51 @@ prop_to_list_defined( Name, Props ) ->
...
 
 
+-spec get_role_credentials(aws_config()) -> {ok, #role_credentials{}} | {error, term()}.
+get_role_credentials(#aws_config{assume_role = AssumeRoleOptions} = Config) ->
+    RoleArn = proplists:get_value(role_arn, AssumeRoleOptions, undefined),
+    case application:get_env(erlcloud, {role_credentials, RoleArn}) of
+        {ok, #role_credentials{expiration_gregorian_seconds = Expiration} = Credentials} ->
+            Now = calendar:datetime_to_gregorian_seconds(calendar:universal_time()),
+            %% Get new credentials if these will expire in less than 2 minutes
+            case Expiration - Now < 120 of
```
motobob: shouldn't this be configurable via `#role_credentials{}`? 
SigV4 does not tolerate drift of 5 min. 
consumers may chose diff refresh guard.

src/erlcloud_aws.erl
```diff
@@ -574,6 +590,51 @@ prop_to_list_defined( Name, Props ) ->
...
+
+-spec get_credentials_from_role(aws_config()) -> {ok, #role_credentials{}} |
+                                                 {error, term()}.
+get_credentials_from_role(#aws_config{assume_role = AssumeRoleOptions} = Config) ->
+    RoleArn = proplists:get_value(role_arn, AssumeRoleOptions, undefined),
+    SessionName = proplists:get_value(session_name, AssumeRoleOptions,
+                                      "erlcloud"),
+    SessionExpiration = proplists:get_value(session_expiration,
+                                            AssumeRoleOptions, 900),
+    ExternalId = proplists:get_value(external_id, AssumeRoleOptions,
```
motobob: well, `external_id can be changed` absolutelly separately from the Role ARN in AWS. 
library requires a way to (invalidate) previous record for consumers.

src/erlcloud_aws.erl
posilva: I will change to 300 secs again but i do not see any advantage of put this a consumer config options.

src/erlcloud_aws.erl
posilva: 👍

src/erlcloud_aws.erl
posilva: I have tried to avoid nested records, but i can create a record if you think that will be more convenient.

src/erlcloud_aws.erl
posilva: 👍

src/erlcloud_aws.erl
posilva: We have two options:
1. Add external ID to the cache key {role_metadata, RoleARN, Externald}
2. We keep the external ID cached and if it changes we invalidate the record and get new credentials from sts with the new ExternalID

**Option 1:**
- Cons: Increase the pollution of the set_env cache system if we have many different external_ids
- Optimize the calls of sts:assume role  

**Option 2:**
- Cons: If most of the operations are between two different externalID we will be calling sts:assume_role many times to swap between external ids
- Pro: Less cache pollution

I prefer Option 1

posilva: Just a small benchmark, from my laptop, with erlcloud_ddb2:get_item(...) the first time I make the call, it takes avg **~1500ms** and the next ones take avg **~190ms**. The first options is definitely better.

motobob:
LGTM. @nalundgaard @fp any comments?

posilva:
> I'll check out how it all behaves in one of our use cases where an app operates with >1000 roles at the same time and come back to you.

@motobob did you check this ^? How the app operates?

motobob:
Thanks you @posilva for contribution and addressing all the comments

motobob:
@posilva look like I missed one comment which was not addressed, namely - library api to clean up the application env from outdated/ deleted role credentials (by ARN and all expired). 
Problem: if you dynamically add/delete role ARNs `erlcloud`, env table grows infinitely and effectively pollutes it until OOM.
If you got some time to have that implemented - it would be great. If not that's fine - i'll pick it up in a week.

posilva:
@motobob I can implement that. Just some sync questions to avoid to have to change later after the review: 

1. The api will be available in the erlcloud_aws module? 
2. Interface: 
- `clear_sts_credentials_cache(RoleArn)` and `clear_sts_credentials_cache(RoleArn, ExternalId) `
- `clear_sts_credentials_cache_outdated()`
- I can arrange to have a on the fly cleaning mechanism using for example `orddict` and/or a bounded cache (with LRU replacement)
- I will create a new PR branching from master

Please let me know what you think.

PR #333, opened 2016-11-16
implement EC2 pagination

motobob:
this is copy of leftover from AL fork under https://github.com/alertlogic/erlcloud/pull/134 

implemented by @SergeyOvchinnik

### DIFF
--- a/src/erlcloud_ec2.erl
+++ b/src/erlcloud_ec2.erl
@@ -31,7 +31,7 @@
          delete_volume/1, delete_volume/2,
          describe_snapshot_attribute/2, describe_snapshot_attribute/3,
          describe_snapshots/0, describe_snapshots/1, describe_snapshots/2,
-         describe_snapshots/3, describe_snapshots/4,
+         describe_snapshots/3, describe_snapshots/4, describe_snapshots/5,
          describe_volumes/0, describe_volumes/1, describe_volumes/2,
          detach_volume/1, detach_volume/2,
          modify_snapshot_attribute/3, modify_snapshot_attribute/4,
@@ -54,7 +54,7 @@
          %% Instances
          describe_instance_attribute/2, describe_instance_attribute/3,
          describe_instances/0, describe_instances/1, describe_instances/2,
-         describe_instances/3,
+         describe_instances/3, describe_instances/4,
          modify_instance_attribute/3, modify_instance_attribute/4,
          modify_instance_attribute/5,
          reboot_instances/1, reboot_instances/2,
@@ -64,7 +64,8 @@
          stop_instances/1, stop_instances/2, stop_instances/3,
          terminate_instances/1, terminate_instances/2,
          describe_instance_status/1, describe_instance_status/2,
-         describe_instance_status/3,
+         describe_instance_status/3, describe_instance_status/4,
+         describe_instance_status/5,
 
          %% Key Pairs
          create_key_pair/1, create_key_pair/2,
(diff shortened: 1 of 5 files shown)


src/erlcloud_ec2.erl
```diff
@@ -1348,37 +1374,66 @@ attribute_xpath(_, AttributeName) ->
...
 
+-type ec2_instances_ids() :: [string()].
+
 %%
+%% Function for making calls to DescribeInstances action
+%% DescribeInstances Documentation: docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstances.html
 %%
--spec describe_instances() -> {ok, [proplist()]} | {error, any()}.
-describe_instances() -> describe_instances([]).
+-spec describe_instances() -> ok_error(proplist()).
```
waisbrot: This is changing from yielding a list of proplists to a single proplist. It looks like the original spec was incorrect and this is a fix.

src/erlcloud_ec2.erl
```diff
@@ -1348,37 +1374,66 @@ attribute_xpath(_, AttributeName) ->
...
-    describe_instances(InstanceIDs, Filter, default_config()).
+    when is_list(InstanceIDs), is_record(Config, aws_config) ->
+    describe_instances(InstanceIDs, [], Config).
 
--spec describe_instances([string()], filter_list(), aws_config()) -> {ok, [proplist()]} | {error, any()}.
+- spec describe_instances(ec2_instances_ids(), filter_list(), aws_config()) -> ok_error(proplist());
+                         (filter_list(), ec2_max_result(), ec2_token()) -> ok_error(proplist(), ec2_token()).
 describe_instances(InstanceIDs, Filter, Config)
-  when is_list(InstanceIDs) ->
+    when is_list(InstanceIDs), is_list(Filter), is_record(Config, aws_config) ->
```
waisbrot: This code is very `when`-heavy. Are there any written style-guides over at AlertLogic for Erlang code? If so, it would be nice to copy/link them from this project. (My own preference in absence of any agreed style is towards picking _one_ `describe_instances/3` and users who want to filter and paginate without providing a custom Config can pass `default_config()` themselves.)

src/erlcloud_ec2.erl
```diff
@@ -1941,44 +2057,84 @@ describe_snapshot_attribute(SnapshotID, create_volume_permission, Config)
             Error
     end.
 
-%%
-%%
--spec describe_snapshots() -> [proplist()].
-describe_snapshots() -> describe_snapshots([], "self").
+-type ec2_snapshot_ids() :: [string()].
+-type ec2_snapshot_owner() :: string().
+-type ec2_snapshot_restorable_by() :: string() | none.
```
waisbrot: Similar style-question. I'm used to putting all my type declarations at the top of the file, even when some are specific to a few private functions.

test/erlcloud_ec2_tests.erl
```diff
@@ -18,6 +18,10 @@
 %% The _f macro is a terse way to wrap code in a fun. Similar to _test but doesn't annotate with a line number
 -define(_f(F), fun() -> F end).
 
+-ifdef(TEST).
+-compile(export_all).
+-endif.
```
waisbrot: The `ifdef` guard seems unnecessary here, since there's no reason to compile this file when not testing.

src/erlcloud_ec2.erl
motobob: original spec was broken, correct

src/erlcloud_ec2.erl
motobob: well, this given example is actually to conform the rest of the library module and not per AL code-style per-se. I guess it just went to extreme. 
my preference would be also to simply all this Java-style /1/2/3/4/5 wrappers. but one thing is for sure used heavily is `default_config()` passing. we usually explicitly set the config since most of the time is cross account

motobob:
i'll update the two minor requested changes and merge this.

PR #337, opened 2016-11-22
Allow the `x-amz-content-sha256` header to be specified

fp:
Since the sign_v4 processing generates the internal header
`x-amz-content-sha256`, it should be possible to specify this value
in case the caller already has a `sha256` of the payload available.
This typically can happen with s3 and represents a nice alternative
to also generating a weaker md5 digest to ensure PutObject
integrity, since sign_v4 does that already.


motobob:
👍

PR #312, opened 2016-09-24
Added AWS ECS support

ptrakhtman:
@nalundgaard and/or @motobob can you please review and merge this PR.
Thank you.

PR #338, opened 2016-11-23
ECS support revised 

motobob:
slightly reworked https://github.com/erlcloud/erlcloud/pull/312 

@ptrakhtman please take a look if you're fine with my updates.

@nalundgaard fyi


include/erlcloud_ecs.hrl
```diff
@@ -0,0 +1,306 @@
...
+}).
+
+-record(ecs_describe_tasks, {
+    tasks :: [#ecs_task{}],
+    failures :: [#ecs_failure{}]
+}).
+
+-record(ecs_list_clusters, {
+    cluster_arns :: [erlcloud_ecs:arn()],
+    next_token :: token()
```
motobob: I'm still not sure `next_token` belongs to the record at all. While at the same time having `{ok, Record}|{ok, Record, next_token}` is also not the best but at least inline with rest of the library. IMO consistency is better choice here.

@nalundgaard @zfox-alertlogic @kkuzmin thought?
if you agree I'll purge this too.

kkuzmin: Agree. Token doesn't belong to a record. Token should be returned separately and passed as a parameter in order to obtain the next record/page.

nalundgaard: Totally disagree. The record is a record form of the response body, which clearly contains these 2 fields. If you are going to obscure the response body, then why not ditch records altogether and just return n-tuples that had position places for each possible response values?

README.md
```diff
@@ -7,15 +7,18 @@ Required functionality is being added upon request.
...
 - Amazon SimpleDB
 - Amazon Mechanical Turk
 - Amazon CloudWatch (MON)
+- Amazon CloudSearch
+- Amazon Inspector
 - Amazon Key Management Service (KMS)
 - Amazon DirectConnect
 - Amazon DynamoDB & DDB streams (ddb2)
-- Amazon Autoscalling(as)
+- Amazon Autoscalling (as)
```
nalundgaard: this should be Autoscaling

include/erlcloud_ecs.hrl
nalundgaard: Maybe you should undo this PR and reinstate the `simple` output that offers stripped down responses that only include the thing(s) most people are likely to care about. Again, the record formats are supposed to model the responses from the API, and to the extend that you guys arbitrarily change that because it sounds more convenient to you, you're going to make this API less intuitive.

nalundgaard: @motobob I think you meant to mention @zacharyfox, BTW.

motobob: @nalundgaard I guess the message was not clear enough. 
- I agree that from json to record conversion perspective `next_token` should be in. and that's why it's here and we will not get enough input - they stay as is right now.
- WRT implementing pagination on consumer side I see no preference of use of `next_token` inside the record or outside in the tuple. *convenience* is irrelevant here.
- the only question on the table was should this new module be consistent with rest or not.

As for `simple` - here is totally disagree and see no value in one more weird way of doing records. I have the same opinion about ddb2 `simple`, but that existed before we took over.

motobob: to close the loop. 
http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/ECS.html#listServices-property
and bot has the same.
so `next_token` stays in.

motobob:
so. 
i'm 👍  as i only simplified #312 which was original ok with me. 
token question self resolved.

any more comments/concerns?

motobob:
i also got offline 👍  from @ptrakhtman

include/erlcloud_ecs.hrl
zacharyfox: 👍 for keeping next_token in the record - it's part of the response

nalundgaard: @motobob Fair enough, maybe I misunderstood what you were proposing. I was trying to say, I prefer functions like this:

```erlang
list_services_all(MaxResults, Token) ->
    list_services_all(MaxResults, Token, []).

list_clusters_all(MaxResults, Token, Acc) -<
    case list_clusters(MaxResults, Token) of
        {ok, #ecs_list_clusters{cluster_arns=Arns, next_token=undefined}} ->
            {ok, flatreverse([Arns| Acc]};
        {ok, #ecs_list_clusters{cluster_arns=Arns, next_token=NextToken}} ->
            list_clusters_all(MaxResults, NextToken, [Arns|Acc]);
        {error, _} = Error -> Error
    end.
```

to this:

```erlang
   case list_clusters(MaxResults, Token) of
        {ok, #ecs_list_clusters{cluster_arns=Arns}} ->
            {ok, flatreverse([Arns| Acc]};
        {ok, #ecs_list_clusters{cluster_arns=Arns}, NextToken} ->
            list_clusters_all(MaxResults, NextToken, [Arns|Acc]);
        {error, _} = Error -> Error
    end.
```

I thought your initial comment was proposing the latter.

Anyway, I guess we are in agreement. And, I suppose on reflection, with the `simple` output as well. If I take stock of my experience with `erlcloud_dd2`, simple output is a convenience that is nice, but probably not worth it.

kkuzmin: The point of separating token out is to have records for effective data and not having them for entire request/responses. I'm as a user of this library no longer interested in tokens inside records during further processing after list all is done.
Anyway if we'd like to keep records as returns let's rename `ecs_list_clusters` to `ecs_list_clusters_responce`, otherwise it sounds misleading to me because at some point we may also want to have records for input, for example `ecs_list_clusters_request`.

nalundgaard:
This looks like it should be good to me. I raised my eyebrows at the abstraction of the outer request layer on update/delete_service. If AWS ever adds any other fields to the response body from those things, we will have no good way to expose that to clients. I'd prefer the explicitness of the outer record, personally, but I am open to being overruled.

motobob:
well.. if they ever do it that must be accompanied by API version bump and thus a change on our side anyway https://github.com/motobob/erlcloud/blob/f471fa3fea391b157d6775bda6bfb5eb0534f57f/src/erlcloud_ecs.erl#L49

include/erlcloud_ecs.hrl
motobob: @nalundgaard  see, it's all about preferences as they are almost equivalent. 

@kkuzmin I see you rationale and that was the reason why I wanted to discuss it all with the community before merging in.

nalundgaard: I agree that we would want to name the records according to `_response` if we were to also have `_request` records, but if that is not the case, we should save ourselves the trouble.

nalundgaard:
@motobob I believe it's possible that they can add (non-breaking) things without version incrementing, but it seems improbable.

I haven't had time to review this with a fine-toothed comb, but in general I like it very much, and what I have been able to delve into in detail has proved correct and looks well-tested. Given that I like the library structure overall, I am 👍 to merge and address any bugs experienced in future use by clients as issues with fixes as usual. Thanks for the contribution @ptrakhtman!

kkuzmin:
@motobob In general I'm fine to ~~bring more chaos~~ elevate erlcloud interface, but I think we need some external feedback from the community.
BTW, this module doesn't comply with existing best practices described in read me:

>## Best Practices ##
>- Public interfaces with paging logic should prefer {ok, Results, Marker} style to the {{paged, Marker}, >Results} found in some modules
>- Public interfaces should normally expose proplists over records

So we should either rethink/enhance the best practices or make our best to unify interfaces with existing ones.

include/erlcloud_ecs.hrl
motobob: @kkuzmin 
I hear your point of "tokens are unnecessary in further processing". but what what is wrong with the record manipulations described by @nalundgaard above? Does it fall under "further processing or not"? keep in mind that proplist format is supported by the module
From pure API response perspective JSON to record conversion is done properly to my standards. 
Please propose your alternative record structure as I was convinced that `{ok, Rec, token}` is bad idea and also looking into other AWS client libs.

motobob:
@kkuzmin tnx for bringing best practices topic. The new module is aligned with ddb2. best practices do not really cover records user and i fix it separatelly in the readme.

src/erlcloud_ecs.erl
```diff
@@ -0,0 +1,2467 @@
...
+-spec describe_services_opts() -> opt_table().
+describe_services_opts() ->
+    [
+        {cluster, <<"cluster">>, fun to_binary/1}
+    ].
+
+-spec describe_services_record() -> record_desc().
+describe_services_record() ->
+    {#ecs_describe_services{},
+     [{<<"services">>, #ecs_
```
motobob: comment does not make sense to me. please clarify where are the booleans? http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Service.html
http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DescribeServices.html

kkuzmin: Sorry. Unfortunately github fails to display comments in right paces when the review is too big.
This comment is actually related to `erlcloud_ecs.erl:2461`

include/erlcloud_ecs.hrl
kkuzmin: @motobob as you previously mentioned: "it's all about preferences".
Nothing is wrong with @nalundgaard suggestion but I'd prefer the following interface:
```
case list_clusters(MaxResults, Token) of
        {ok, #ecs_list_clusters{cluster_arns=Arns}} ->
            {ok, flatreverse([Arns| Acc]};
        {ok, #ecs_list_clusters{cluster_arns=Arns}, NextToken} ->
            list_clusters_all(MaxResults, NextToken, [Arns|Acc]);
        {error, _} = Error -> Error
    end.
```
 because it complies with existing best practices. My only request is not to introduce more interface inconsistencies as there are lots of them across existing modules already: some of them return tokens inside a tuple `{ok, Results, Token}` and some inside a resulting proplist, [for example](https://github.com/alertlogic/erlcloud/blob/master/src/erlcloud_sns.erl#L303-L309).

kkuzmin:
@motobob I'm both hands up for supporting records in output, but frankly I missed the moment when ddb2 became the library best practice. Let's state it clear in the read me.

motobob:
readme to be addressed separatelly as agreed ^^^

PR #340, opened 2016-11-25
Fix handling of 'count' select in ddb query/scan all

arthurnovak:
`q_all`/`scan_all` should be able to return count value in case of setting `{select, count}` in `Opts`. At the moment it fails because there is no specific case clause for that.
AWS always returns `undefined` items for `count` request, so we'll match on that response.

@motobob @danilagamma

### DIFF
--- a/src/erlcloud_ddb_util.erl
+++ b/src/erlcloud_ddb_util.erl
@@ -49,7 +49,9 @@
 -type range_key_name() :: erlcloud_ddb2:range_key_name().
 -type table_name() :: erlcloud_ddb2:table_name().
 
--type items_return() :: {ok, [out_item()]} | {error, term()}.
+-type items_return() :: {ok, [out_item()]}
+                      | {ok, non_neg_integer()}
+                      | {error, term()}.
 
 -export_type(
    [batch_read_ddb_opt/0,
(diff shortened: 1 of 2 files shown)


motobob:
apparently this never worked so please add a unit test for this.

src/erlcloud_ddb_util.erl
```diff
@@ -309,22 +311,39 @@ q_all(Table, KeyConditionsOrExpression, Opts) ->
...
+                     items              = undefined,
+                     count              = Count}} ->
+            {ok, lists:sum([Count|Acc])};
+        {ok, #ddb2_q{last_evaluated_key = undefined,
+                     items              = Items}} ->
+            {ok, flatreverse([Items|Acc])};
+        {ok, #ddb2_q{last_evaluated_key = LastKey,
+                     items              = Items,
+                     count              = Count}} ->
+            UpdAcc = case Items of
```
motobob: seems unnecessary nesting to me. 
make it outer case clauses please by `items = undefined`

arthurnovak: Actually it looks better for me having one case clause for case when `last_evaluated_key /= undefined`. I think it makes sense to call `q_all(Tbl, KeyConditionsOrExpression, Opts, Config, UpdAcc, LastKey)` just once in one clause.

nalundgaard:
@arthurnovak It looks like this `applies to `scan_all`](http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Scan.html#DDB-Scan-request-Select) as well. Any reason not to also fix that?

arthurnovak:
@nalundgaard I need just `q_all` 😄. But yeah, it makes sense to fix that behavior in `scan_all` as well. Will do

src/erlcloud_ddb_util.erl
arthurnovak: ok

PR #342, opened 2016-11-28
Use V4 signature for signing S3 urls

kovyl2404:
Hi, guys. Some time ago we encountered an issue with temporary S3 links. Some datacenters (such as eu-central-1) doesn't support old signature algorithm. This PR enables use of V4 signature for S3 temporary urls.

### DIFF
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -15,6 +15,7 @@
          http_headers_body/1,
          http_body/1,
          request_to_return/1,
+         sign_v4_request/9,
          sign_v4_headers/5,
          sign_v4/8,
          get_service_status/1,
@@ -627,6 +628,17 @@ request_to_return(#aws_request{response_type = error,
                                response_headers = Headers}) ->
     {error, {http_error, Status, StatusLine, Body, Headers}}.
 
+
+-spec sign_v4_request(atom(), list(), aws_config(), headers(), list(), string(), string(), list(), string()) -> string().
+sign_v4_request(Method, Path, Config, Headers, Payload, Region, Service, QueryParams, DateTime) ->
+  {CanonicalRequest, _} = canonical_request(Method, Path, QueryParams, Headers, Payload),
+  io:format("--------------\n~s\n--------------------\n", [CanonicalRequest]),
+  Scope = credential_scope(DateTime, Region, Service),
+  StringToSign = to_sign(DateTime, Scope, CanonicalRequest),
+  SigningKey = signing_key(Config, DateTime, Region, Service),
+  base16(erlcloud_util:sha256_mac(SigningKey, StringToSign)).
+
+    
 %% http://docs.aws.amazon.com/general/latest/gr/signature-version-4.html
 -spec sign_v4_headers(aws_config(), headers(), binary(), string(), string()) -> headers().
 sign_v4_headers(Config, Headers, Payload, Region, Service) ->
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -632,7 +632,6 @@ request_to_return(#aws_request{response_type = error,
 -spec sign_v4_request(atom(), list(), aws_config(), headers(), list(), string(), string(), list(), string()) -> string().
 sign_v4_request(Method, Path, Config, Headers, Payload, Region, Service, QueryParams, DateTime) ->
   {CanonicalRequest, _} = canonical_request(Method, Path, QueryParams, Headers, Payload),
-  io:format("--------------\n~s\n--------------------\n", [CanonicalRequest]),
   Scope = credential_scope(DateTime, Region, Service),
   StringToSign = to_sign(DateTime, Scope, CanonicalRequest),
   SigningKey = signing_key(Config, DateTime, Region, Service),
--- a/src/erlcloud_s3.erl
+++ b/src/erlcloud_s3.erl
@@ -30,6 +30,7 @@
          set_object_acl/3, set_object_acl/4,
          make_link/3, make_link/4,
          make_get_url/3, make_get_url/4,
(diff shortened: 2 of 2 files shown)


kkuzmin:
Hi Viacheslav.

I think the issue with request V4 signing has been fixed already. Please see https://github.com/erlcloud/erlcloud/pull/218
Signature V2 is no longer used inside erlcloud.

Could you please verify the issue is still valid with the latest library version?
If it still the case could you please provide more details about the problem you faced with?
Thanks in advance.

Regards,
Konstantin

kovyl2404:
Yup, I've seen this PR. But anyway, some functions like [make_get_url](https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_s3.erl#L1025) still use old [signing algorithm](https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_s3.erl#L971).
In this PR we just enable use of V4 signing for these functions.

kkuzmin:
Thank you for the explanation.
In this case I'd suggest to fix existing `make_get_url/3`and `make_get_url/4` to use V4 signature only rather than introducing the new ones. It is safe to do so because all AWS S3 regions support sign v4 now.

src/erlcloud_aws.erl
```diff
@@ -627,6 +628,16 @@ request_to_return(#aws_request{response_type = error,
                                response_headers = Headers}) ->
     {error, {http_error, Status, StatusLine, Body, Headers}}.
 
+
+-spec sign_v4_request(atom(), list(), aws_config(), headers(), list(), string(), string(), list(), string()) -> string().
+sign_v4_request(Method, Path, Config, Headers, Payload, Region, Service, QueryParams, DateTime) ->
+  {CanonicalRequest, _} = canonical_request(Method, Path, QueryParams, Headers, Payload),
```
kkuzmin: Please fix indention within this function.

src/erlcloud_s3.erl
```diff
@@ -1014,6 +1015,47 @@ make_link(Expire_time, BucketName, Key, Config) ->
...
 
+ -spec make_get_url_v4(integer(), string(), string()) -> iolist().
+
+make_get_url_v4(ExpireTime, BucketName, Key) ->
+    make_get_url_v4(ExpireTime, BucketName, Key, default_config()).
+  
+-spec make_get_url_v4(integer(), string(), string(), aws_config()) -> iolist().
+
+make_get_url_v4(ExpireTime, BucketName, Key, Config) ->
+  Region = aws_region_from_host(Config#aws_config.s3_host),
```
kkuzmin: Please fix indention within this function.

src/erlcloud_s3.erl
```diff
@@ -1014,6 +1015,47 @@ make_link(Expire_time, BucketName, Key, Config) ->
...
+  Host = "s3-" ++ Region ++ ".amazonaws.com",
```
kkuzmin: This is not correct because there is no `s3-us-east-1.amazonaws.com` [endpoint](http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region).  
`Config#aws_config.s3_host` should be used as a `Host`.

kovyl2404:
Konstantin, I need your advice. What kind of tests should I implement? What exactly do we want to cover?

src/erlcloud_s3.erl
```diff
@@ -1014,24 +1014,47 @@ make_link(Expire_time, BucketName, Key, Config) ->
...
-    lists:flatten([get_object_url(BucketName, Key, Config),
-     "?AWSAccessKeyId=", erlcloud_http:url_encode(Config#aws_config.access_key_id),
-     "&Signature=", erlcloud_http:url_encode(Sig),
-     "&Expires=", Expires,
-     SecurityTokenQS]).
+make_get_url(ExpireTime, BucketName, Key, Config) ->
+    Region = aws_region_from_host(Config#aws_config.s3_host),
+    Host = "s3-" ++ Region ++ ".amazonaws.com",
+    Path = erlcloud_http:url_encode_loose("/" ++ BucketName ++ "/" ++ Key),
+    EndpointUrl = "https://" ++ Host ++ Path,
```
kkuzmin: This is not correct because there is no `s3-us-east-1.amazonaws.com` [endpoint](http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region).  
`Config#aws_config.s3_host` should be used as a `Host` instead.

kkuzmin:
I think basic unit tests should be enough to check how URL is contracted from the given `#aws_config{}` and other parameters. Also tests always make a review easier.

Sorry for copy-pasting the comments  as they appeared to be inside outdated diff.

src/erlcloud_s3.erl
```diff
@@ -1014,24 +1014,48 @@ make_link(Expire_time, BucketName, Key, Config) ->
...
-     SecurityTokenQS]).
+make_get_url(ExpireTime, BucketName, Key, Config) ->
+    Host = Config#aws_config.s3_host,
+    Region = aws_region_from_host(Host),
+    Path = erlcloud_http:url_encode_loose("/" ++ BucketName ++ "/" ++ Key),
+    EndpointUrl  = lists:flatten([
+      Config#aws_config.s3_scheme,
+      Host, port_spec(Config), Path]),
+    Headers = [{"Host", Host}],
+    Service = "s3",
```
motobob: let's not just duplicate `erlcloud_aws:sign_v4` one more time

andrey-kulikov: I can not using erlcloud_aws:sign_v4 result for signing temporary url. We need erlcloud_aws function that just sign request but not do and return anysing else. I thnk sign_v4 - good name for this function but everything else not related to signature can be moved to other function to prevent copypaste. But sign_v4 - public function and we can not change its interface. I added sign_v4_request and can move there all related to signature.

PR #345, opened 2016-11-30
Fixed dialyzer warnings in cloudformation, cloudtrail and elb

SergeyOvchinnik:
Problem:

- Modules erlcloud_cloudformation, erlcloud_cloudtrail and erlcloud_elb had 16 dialyzer warnings between them
- Module erlcloud_cloudformation had several broken functions eg. `default_config` atom was used instead of `default_config()` call

Solution:

- Fixed all dialyzer warnings in those 3 modules by correcting specs and fixing broken functions

### DIFF
--- a/src/erlcloud_cloudformation.erl
+++ b/src/erlcloud_cloudformation.erl
@@ -338,7 +338,7 @@ get_template(StackName, Config = #aws_config{}) ->
 -spec get_template_summary(params(), string()) ->
     {ok, cloudformation_list()} | {error, error_reason()}.
 get_template_summary(Params, StackName) ->
-    get_template_summary(Params, StackName, default_config).
+    get_template_summary(Params, StackName, default_config()).
 
 -spec get_template_summary(params(), string(), aws_config()) ->
     {ok, cloudformation_list()} | {error, error_reason()}.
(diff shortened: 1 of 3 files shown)


src/erlcloud_elb.erl
```diff
@@ -151,7 +151,7 @@ configure_health_check(LB, Target, Config) when is_list(LB) ->
 %% specific configuration and specific balancer name. 
 %% @end
 %% --------------------------------------------------------------------
--spec describe_load_balancer(string()) -> proplist().
+-spec describe_load_balancer(string()) -> {{paged, string()}, proplist()} | {error, term()}.
```
motobob: this one does seem off. what if i just have 1 ELB?

src/erlcloud_cloudformation.erl
```diff
@@ -360,20 +360,20 @@ get_template_summary(Params, StackName, Config = #aws_config{}) ->
 -spec describe_account_limits_all() ->
     {ok, cloudformation_list()} | {error, error_reason()}.
 describe_account_limits_all() ->
-    describe_account_limits_all(default_config).
+    describe_account_limits_all(default_config()).
```
motobob: this one is just awesome :)

src/erlcloud_cloudtrail.erl
```diff
@@ -172,9 +173,7 @@ update_trail(Trail, S3BucketName, S3KeyPrefix, SnsTopicName, IncludeGlobalServic
 
 % Json parameter must be a list of binary key/value tuples.
 ct_request(Operation, [], Config) ->
-    #aws_config{cloudtrail_scheme = Scheme, 
-                cloudtrail_host = Host} = Config,
-    request_impl(post, Scheme, Host, port_spec(Config), "/", Operation, [], "{}", Config);
+    ct_request(Operation, "{}", Config);
```
motobob: can you elaborate on the change?

src/erlcloud_elb.erl
SergeyOvchinnik: Yep, forgot about the {ok, Results} case here. Fixing...

src/erlcloud_cloudtrail.erl
SergeyOvchinnik: This clause defined for empty `Body` param was doing exactly the same thing as the clause with non-empty `Body` except it was passing `Body` as a string instead of binary. The functions this `Body` param was getting passed to are expecting a binary so dialyzer was complaining. Seems like the purpose of this clause was to define a special case for empty body and instead of calling `jsx:encode([])` it just has the result of this call hard coded. The problem is: the result of `jsx:encode([])` is `<<"[]">>` and not `"{}"`. I suppose the correct solution would be to either remove the clause completely or have this clause same as it was but with `"{}"` replaced with `<<"[]">>`

src/erlcloud_cloudformation.erl
SergeyOvchinnik: Kinda wish all of these warnings were as simple as that

SergeyOvchinnik: I know right :-D

src/erlcloud_cloudtrail.erl
```diff
@@ -171,11 +172,6 @@ update_trail(Trail, S3BucketName, S3KeyPrefix, SnsTopicName, IncludeGlobalServic
     ct_request("UpdateTrail", Json, Config).
 
 % Json parameter must be a list of binary key/value tuples.
-ct_request(Operation, [], Config) ->
-    #aws_config{cloudtrail_scheme = Scheme, 
-                cloudtrail_host = Host} = Config,
-    request_impl(post, Scheme, Host, port_spec(Config), "/", Operation, [], "{}", Config);
```
motobob:
```erlang
(cloud_explorer@mbr13_ebob)2> jsx:encode([]).
<<"[]">>
```
and not "{}"
this will break describe trails

motobob:
fix api back and reopen

PR #348, opened 2016-12-02
Begin supporting ElasticMapReduce

jonathn-al:
Only a few api functions so far, but probably better than nothing.


motobob:
closing in favour of already merged one.

PR #351, opened 2016-12-04
add clear_config api

motobob:
Problem: 
#332 added function to add credentials into application config but no clear api was added 

Solution:
Add unset API for one config and all expired ones

### DIFF
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -9,7 +9,8 @@
          aws_region_from_host/1,
          aws_request_form/8,
          aws_request_form_raw/8,
-         param_list/2, default_config/0, auto_config/0, update_config/1,
+         param_list/2, default_config/0, auto_config/0,
+         update_config/1,clear_config/1, clear_expired_configs/0,
          service_config/3, service_host/2,
          configure/1, format_timestamp/1,
          http_headers_body/1,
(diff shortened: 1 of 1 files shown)


motobob:
@posilva here is the cleanup api added. please let me know if that matched your use case.

posilva:
@motobob Yes, it ok for my use case. I would do it more conservative, with a bounded number (configurable) of allowed  cache entries, eventually with some LRU strategy. But it works for me.

motobob:
@posilva I intentionally do not want to complicate the library with any caches and additional processes running. But it might change in future.

motobob:
@kkuzmin any feedback?

kkuzmin:
@motobob could you please also add cleanup call [here](https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_aws.erl#L606)

src/erlcloud_aws.erl
```diff
@@ -411,6 +412,21 @@ update_config(#aws_config{} = Config) ->
                    security_token = Credentials#metadata_credentials.security_token}}
     end.
 
+-spec clear_config(aws_config()) -> ok.
+clear_config(#aws_config{assume_role = #aws_assume_role{role_arn = Arn, external_id = ExtId}}) ->
+    application:unset_env(erlcloud, {role_credentials, Arn, ExtId}).
+
+-spec clear_expired_configs() -> ok.
+clear_expired_configs() ->
```
kkuzmin: I'd rather name this function `clear_expired_credentials()`

kkuzmin:
@motobob other than the comments above 👍 .

motobob:
@kkuzmin 
>@motobob could you please also add cleanup call here

there is set_env call right at the end of that fun at https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_aws.erl#L632-L636

kkuzmin:
@motobob thanks for the explanation. 👍

PR #350, opened 2016-12-03
Added support for AWS Config in erlcloud_s3:explore_dirstructure

SergeyOvchinnik:
Problem:

- erlcloud_s3:explore_dirstructure function did not have support for AWS Config
- Issue [#346](https://github.com/erlcloud/erlcloud/issues/346) has been raised

Solution:

- Added a new clause for erlcloud_s3:explor_dirstructure which supports AWS Config

Issue #346, opened 2016-12-02
erlcloud_s3:explore_dirstructure() doesn't take AwsConfig argument


jonathn-al:
erlcloud_s3:explore_dirstructure() doesn't have a variant taking an AwsConfig argument similar to most other functions.

SergeyOvchinnik:
Doesn't look like erlcloud_s3:explore_dirstructure uses Config anywhere anyway, why would passing Config to it be required?

jonathn-al:
It would be passed on to erlcloud_s3:list_objects(..., Config) here:
  https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_s3.erl#L327

SergeyOvchinnik:
Oh, I see what you mean. Yeah, can implement that

motobob:
addressed

PR #353, opened 2016-12-05
Fix erlcloud_s3:explore_dirstructure/4

jonathn-al:
The previous PR (https://github.com/erlcloud/erlcloud/pull/350) addressing AwsConfig parameter issue (https://github.com/erlcloud/erlcloud/issues/346), didn't work because it didn't pass the Config back into recursive calls.


motobob:
tnx. sorry for an overlook while merging previous PR

PR #354, opened 2016-12-06
read AWS region from env/appenv

massemanet:
This patch automatically changes the default AWS config to reflect the AWS region. It does this by checking for the existence of the OS environment variable "AWS_REGION" and the 'erlcloud' application env 'aws_region'; if neither exists this patch does nothing.

### DIFF
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -319,7 +319,7 @@ default_config_wrap() ->
 %% return undefined
 default_config_get(OsVar, EnvVar) ->
     case {os:getenv(OsVar), application:get_env(erlcloud, EnvVar)} of
-        {OsVal,  _} when OsVal /= undefined -> OsVal;
+        {OsVal,  _} when OsVal /= false -> OsVal;
         {_, {ok, EnvVal}} -> EnvVal;
         _ -> undefined
     end.
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -309,44 +309,27 @@ default_config() ->
     end.
 
 default_config_wrap() ->
-    {Id, Key, Token, Region} = default_config_get(),
+    Id = default_config_get("AWS_ACCESS_KEY_ID", aws_access_key_id),
+    Key = default_config_get("AWS_SECRET_ACCESS_KEY", aws_secret_access_key),
+    Token = default_config_get("AWS_SECURITY_TOKEN", aws_security_token),
+    Region = default_config_get("AWS_REGION", aws_region),
     default_config_region(default_config_assert(Id, Key, Token), Region).
-
-default_config_get() ->
-    {default_config_get("AWS_ACCESS_KEY_ID", aws_access_key_id),
-     default_config_get("AWS_SECRET_ACCESS_KEY", aws_secret_access_key),
-     default_config_get("AWS_SECURITY_TOKEN", aws_security_token),
-     default_config_get("AWS_REGION", aws_region)}.
 
 %% try to get config value from OS env, failing that try app env, else
 %% return undefined
-default_config_get(OSENV, APPENV) ->
-    case default_config_osenv(OSENV) of
-        undefined -> default_config_appenv(APPENV);
-        Val -> Val
+default_config_get(OsVar, EnvVar) ->
+    case {os:getenv(OsVar), application:get_env(erlcloud, EnvVar)} of
+        {OsVal,  _} when OsVal /= undefined -> OsVal;
+        {_, {ok, EnvVal}} -> EnvVal;
+        _ -> undefined
     end.
 
-default_config_osenv(ENV) ->
-    case os:getenv(ENV) of
-        Value when is_list(Value) -> Value;
-        false -> undefined
-    end.
-
-default_config_appenv(ENV) ->
-    case application:get_env(erlcloud, ENV) of
-        {ok, Value} -> Value;
-        undefined -> undefined
-    end.
-
+default_config_assert(undefined, _Key, _Token) -> #aws_config{};
+default_config_assert(_Id, undefined, _Token) -> #aws_config{};
 default_config_assert(Id, Key, Token) ->
-    case Id==undefined orelse Key==undefined of
-        false ->
-            #aws_config{access_key_id = Id,
-                        secret_access_key = Key,
-                        security_token = Token};
-        true ->
-            #aws_config{}
-    end.
+    #aws_config{access_key_id = Id,
+                secret_access_key = Key,
+                security_token = Token}.
 
 %% call service_config/3 with our Region for all services
 default_config_region(AwsConfig, undefined) ->
@@ -510,7 +493,7 @@ clear_expired_configs() ->
 
 %%%---------------------------------------------------------------------------
 -spec service_config( Service :: atom() | string() | binary(),
-                      Region :: string() | binary(),
+                      Region :: atom() | string() | binary(),
                       Config :: #aws_config{} ) -> #aws_config{}.
 %%%---------------------------------------------------------------------------
 %% @doc Generate config updated to work with specified AWS service and region
@@ -529,6 +512,8 @@ service_config( Service, Region, Config ) when is_atom(Service) ->
     service_config( atom_to_binary(Service, latin1), Region, Config );
 service_config( Service, Region, Config ) when is_list(Service) ->
     service_config( list_to_binary(Service), Region, Config );
+service_config( Service, Region, Config ) when is_atom(Region) ->
+    service_config( Service, atom_to_binary(Region, latin1), Config );
 service_config( Service, Region, Config ) when is_list(Region) ->
     service_config( Service, list_to_binary(Region), Config );
 service_config( <<"as">>, Region, Config ) ->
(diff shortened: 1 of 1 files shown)


src/erlcloud_aws.erl
```diff
@@ -363,6 +370,39 @@ config_env() ->
...
+
+config_with_region(Region, Config) ->
+    Services =
+        [autoscaling,
+         cloudformation,
+         cloudsearch,
+         cloudtrail,
+         dynamodb,
+         'streams.dynamodb',
+         ec2,
```
motobob: `ecs` is already missing.

motobob:
https://github.com/erlcloud/erlcloud/pull/354/files#diff-b6a88197b95cfa7d84db56a756107eadR362 should be changed too probably

src/erlcloud_aws.erl
```diff
@@ -363,6 +370,39 @@ config_env() ->
         _ -> {error, environment_config_unavailable}
     end.
 
+config_with_region(Config) ->
+    case {os:getenv("AWS_REGION"), application:get_env(erlcloud, aws_region)} of
```
motobob: we do not do aws_access_key_id/aws_secret_access_key/aws_security_token through application env. so may be we should not do it for region either?

src/erlcloud_aws.erl
massemanet: yes, this list is bad. Perhaps code that traverses the #aws_config{} record and looks for tags of the form *_host would be more future proof?

src/erlcloud_aws.erl
massemanet: Or the other way around; allow aws_access_key_id/aws_secret_access_key/aws_security_token to be set through application env.

src/erlcloud_aws.erl
massemanet: like this abomination;
```[list_to_atom(H)||[H,"host"]<-[string:tokens(atom_to_list(F),"_")||F<-record_info(fields,aws_config)]].```

Comment (motobob) [inline src/erlcloud_aws.erl]: seems nice to me

File: src/erlcloud_aws.erl (hunk 2 above)
Comment (motobob) [inline src/erlcloud_aws.erl]: so we already have OS env, AWS profile, and config params. 
making it via application env is okay way but i'm not sure we need one more way.

Comment (motobob) [inline src/erlcloud_aws.erl]: @nalundgaard @waisbrot what are your thoughts?

Comment (nalundgaard) [inline src/erlcloud_aws.erl]: I'm not opposed to it, as long as it's backwards-compatible and well-documented.

Comment (motobob) [inline src/erlcloud_aws.erl]: okay, let's do it with app env too. and update the readme

**Comment by massemanet** [comment]
Seems like github blew away the conversation.
Anyway, this PR makes erlcloud_aws:default_config check the environment (first the OS env, if that fails the erlcloud application env), for the configs AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SECURITY_TOKEN, and AWS_REGION. It then modifies the returned #aws_config{} appropriately.

If AWS_REGION is unset and the ercloud application env is unused, this patch does nothing.

If AWS_REGION is set, we run erlcloud_aws:service_config on all services.
The list of services is arrived at by looking for keys in #aws_config{} that ends in "_host".
For this to work, I had to add some aliases in service_config;

 as -> autoscaling
 ddb_streams -> streams.dynamodb
 mon -> monitoring
 mturk -> mechanicalturk
 mms -> metering.marketplace

I added entries in service_config for the following services:
 directconnect
 inspector
 monitoring
 redshift
 route53

where monitoring and route53 does not get regions set.

Offtopic; in service_config, both "as" and "autoscaling" set "#aws_config.as_host" field, while "#aws_config.autoscaling_host" is not settable. Bug or feature?

**Comment by motobob** [comment]
tnx for the update. i'll take a look in couple day on the update.
re: `as` vs `autoscalling` - is a legacy feature, as these modules implement same service in two different forms (records vs proplists) and they should be merged

File: src/erlcloud_aws.erl [hunk 3]
```diff
@@ -289,19 +289,81 @@ format_timestamp({{Yr, Mo, Da}, {H, M, S}}) ->
...
+
+default_config_get() ->
+    {default_config_get("AWS_ACCESS_KEY_ID", aws_access_key_id),
+     default_config_get("AWS_SECRET_ACCESS_KEY", aws_secret_access_key),
+     default_config_get("AWS_SECURITY_TOKEN", aws_security_token),
+     default_config_get("AWS_REGION", aws_region)}.
+
+%% try to get config value from OS env, failing that try app env, else
+%% return undefined
+default_config_get(OSENV, APPENV) ->
```
motobob: sorry, probably it's more of a code style but i do not like this function spiral.
let's have it
```erlang
default_config_get(OsVar, EnvVar) -> 
   case {os:getenv(OsVar), application:get_env(erlcloud,EnvVar} of
       {OsVal,  _} when OsVal /= undefined -> OsVal;
       {_, EnvVal} when EnvVal /= undefined -> EnvVal;
       _ -> undefined
   end.
```

src/erlcloud_aws.erl
```diff
@@ -289,19 +289,81 @@ format_timestamp({{Yr, Mo, Da}, {H, M, S}}) ->
...
+        false -> undefined
+    end.
+
+default_config_appenv(ENV) ->
+    case application:get_env(erlcloud, ENV) of
+        {ok, Value} -> Value;
+        undefined -> undefined
+    end.
+
+default_config_assert(Id, Key, Token) ->
```
motobob: fun with a single case.  
```erlang
default_config_assert(undefined, undefined, _) ->
    #aws_config{};
default_config_assert(Id, Key, Token) ->
    #aws_config{access_key_id = Id,
                         secret_access_key = Key,
                         security_token = Token}.
```

src/erlcloud_aws.erl
```diff
@@ -289,19 +289,81 @@ format_timestamp({{Yr, Mo, Da}, {H, M, S}}) ->
...
     end.
 
+%% call service_config/3 with our Region for all services
+default_config_region(AwsConfig, undefined) ->
+    AwsConfig;
+default_config_region(AwsConfig, Region) ->
+    ConfF = fun(Service, C0) -> service_config(Service, Region, C0) end,
+    lists:foldl(ConfF, AwsConfig, default_config_region_services()).
+
+%% horrible hack to get a list of service names, assuming that if a key in
```
motobob: adding one line comment to aws_config records will not make it so horrible. 
but the idea is okay to have as discussed before

src/erlcloud_aws.erl
```diff
@@ -289,19 +289,81 @@ format_timestamp({{Yr, Mo, Da}, {H, M, S}}) ->
...
+                        secret_access_key = Key,
+                        security_token = Token};
+        true ->
+            #aws_config{}
     end.
 
+%% call service_config/3 with our Region for all services
+default_config_region(AwsConfig, undefined) ->
+    AwsConfig;
+default_config_region(AwsConfig, Region) ->
```
motobob: `service_config` looks for Region to be binary. lets protect it as we have two inputs, OS and Env here.

src/erlcloud_aws.erl
```diff
@@ -289,19 +289,81 @@ format_timestamp({{Yr, Mo, Da}, {H, M, S}}) ->
...
 
-default_config_env() ->
-    case config_env() of
-        {ok, Config} -> Config;
-        {error, _} -> #aws_config{}
+default_config_wrap() ->
+    {Id, Key, Token, Region} = default_config_get(),
+    default_config_region(default_config_assert(Id, Key, Token), Region).
+
+default_config_get() ->
```
motobob: i feel this fun is redundant and we can have it ^ directly

src/erlcloud_aws.erl
massemanet: your code does not have the same semantics as the original (which required that neither Id nor Key was undefined).
Is that a change you want to make?

src/erlcloud_aws.erl
massemanet: OK

src/erlcloud_aws.erl
massemanet: OK

src/erlcloud_aws.erl
massemanet: Not sure what you mean here. ```service_config``` is speced with ```Region :: string() | binary()```.
I guess it could be an atom though...

src/erlcloud_aws.erl
massemanet: Unsure again. You want me to add a comment? If so, where?

src/erlcloud_aws.erl
massemanet: Changed ```service_config``` to accept atom(Region), since there was already a bunch of such checks.

src/erlcloud_aws.erl
massemanet: rewrote to get rid of the case

motobob:
@massemanet thx for all the updates. i have no more comments code-wise.
as a last change, may i ask you to update https://github.com/erlcloud/erlcloud#using-temporary-security-credentials and we :shipit:

motobob:
i'll update the readme

PR #357, opened 2016-12-12
Make erlcloud_mms respect the aws_config assume_role value

carlisom:
Update `erlcloud_mms` to allow users to assume_roles via `#aws_config`. This will prevent users from needing to update the config manually before calling `erlcloud_mms`.


nalundgaard:
👍

PR #359, opened 2016-12-22
bump things Forward

motobob:
- remove R14/15 and simplify util module
- move travis to 19.1
- move lhttpc to 1.4.1 (larger default pool)
- prepare for total dyalizer in make file

`2.1.0` this to become.


Makefile
```diff
@@ -17,31 +17,14 @@ endif
 # eventually this should be just ebin/*.beam, but there are a number
 # of warnings in other files. Just check the clean files for now.
```
waisbrot: Remove the "this should just be ebin/*.beam" now that it is? :-)

motobob: good catch

PR #360, opened 2016-12-22
add london region to S3

motobob:


src/erlcloud_s3.erl
```diff
@@ -228,7 +229,8 @@ encode_location_constraint('us-east-1') -> undefined;
 encode_location_constraint('us-east-2') -> "us-east-2";
 encode_location_constraint('us-west-1') -> "us-west-1";
 encode_location_constraint('us-west-2') -> "us-west-2";
-encode_location_constraint('eu-west-1') -> "EU";
+encode_location_constraint('eu-west-1') -> "eu-west-1";
```
motobob: eu -> "EU" but `eu-west-1`is `"eu-west-1"`. this change is just cosmetic and fine  with http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUT.html

PR #362, opened 2016-12-27
bump to lhttpc 1.5.0 (proper chuked handling)

motobob:
@fp FYI

@nalundgaard @waisbrot when you have a sec.


waisbrot:
👍 
Longer-term, though, I'd be interested in switching Erlcloud over to Hackney (or some other client) rather than staying on lhttpc as the erlcoud-specific http client.

motobob:
@waisbrot totally support, `hackney` or `gun` are in my consideration.

PR #363, opened 2016-12-28
Add application dep on new base16 component

fp:
Without this relx packaging does not pick up the base16 application
and therefore erlcloud is not packaged completely.


nalundgaard:
👍. @carlisom OTP shame!

carlisom:
The need the Game of Thrones shame bell for situations like this.

PR #364, opened 2016-12-29
Allow `profile_option()` list for `auto_config/1`

fp:
Add new `erlcloud_aws:auto_config/1` function that allows the caller
to specify a list of `profile_option()` tuples used if the source
of the credentials is the credentials file.

### DIFF
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -9,7 +9,7 @@
          aws_region_from_host/1,
          aws_request_form/8,
          aws_request_form_raw/8,
-         param_list/2, default_config/0, auto_config/0,
+         param_list/2, default_config/0, auto_config/0, auto_config/1,
          update_config/1,clear_config/1, clear_expired_configs/0,
          service_config/3, service_host/2,
          configure/1, format_timestamp/1,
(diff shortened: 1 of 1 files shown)


src/erlcloud_aws.erl
```diff
@@ -336,13 +336,30 @@ default_config_env() ->
...
+
+
+%%%---------------------------------------------------------------------------
+-spec auto_config( ProfileOptions :: [profile_option()]) ->
+                         {ok, aws_config()} | undefined.
+%%%---------------------------------------------------------------------------
+%% @doc Generate config using the best available credentials
+%%
+%% This function works the same as {@link auto_config/0}, but if credentials
+%% are developed from <em>User Profile</code> as the source, the
```
nalundgaard: looks like mismatched tags here? s/em/code?

fp: fak, just a sec

fp: i mean, i could just not write doc like the rest of this code, and then i wouldn't have this problem...

nalundgaard: :trollface:

fp:
@motobob have a look, please

nalundgaard:
Makes sense to me. 👍

PR #365, opened 2016-12-30
bump lhttpc to 1.5.1 where canonical headers reverted

motobob:
fix previous version where lhttpc 1.5.0 removed canonical headers and broke erlcloud.

repin to a fixed lhttpc 1.5.1

hopefully, this is the last lhttpC venture and next will be hackney/gun switch.


motobob:
restore justice and fix a bug introduced in 2.2.0

PR #366, opened 2017-01-03
Added support for flow logs

ptrakhtman:


src/erlcloud_ec2.erl
```diff
@@ -2936,6 +2946,168 @@ net_if_params(#ec2_net_if{}=X) ->
...
+        {"TrafficType", TT},
+        {"DeliverLogsPermissionArn", DeliverLogsPermissionArn}
+    ] ++ Resources,
+    Params = case ClientToken of
+                none -> P;
+                _ -> P ++ {"ClientToken", ClientToken}
+            end,
+    case ec2_query(Config, "CreateFlowLogs", Params, ?NEW_API_VERSION) of
+        {ok, Doc} ->
+            {ok, [get_text(Item) || Item <- xmerl_xpath:string("/CreateFlowLogsResponse/flowLogIdSet/item", Doc)]};
```
motobob: it seems the API supports partial success as it also has `unsuccessful` item in response. please add that. 
same applies to `clientToken`

src/erlcloud_ec2.erl
```diff
@@ -2936,6 +2946,168 @@ net_if_params(#ec2_net_if{}=X) ->
...
+delete_flow_logs(FlowLogIDs, Config) when is_record(Config, aws_config) ->
+    {Resources, _} = lists:foldl(fun(ID, {Acc, Index}) ->
+                                    I = integer_to_list(Index),
+                                    FlowLogID = "FlowLogId."++I,
+                                    {[{FlowLogID, ID} | Acc], Index+1}
+                            end, {[], 1}, FlowLogIDs),
+
+    case ec2_query(Config, "DeleteFlowLogs", Resources, ?NEW_API_VERSION) of
+        {ok, Doc} ->
+            {ok,[{unsuccessful, get_text("/DeleteFlowLogsResponse/unsuccessful", Doc)}]};
```
motobob: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_UnsuccessfulItem.html is an object so it probably should be parsed accordingly.

src/erlcloud_ec2.erl
```diff
@@ -2936,6 +2946,168 @@ net_if_params(#ec2_net_if{}=X) ->
...
+                        (ec2_flow_ids()) -> ok_error(proplist()).
+describe_flow_logs(Config)
+    when is_record(Config, aws_config) ->
+    describe_flow_logs([], Config);
+describe_flow_logs(FlowIDs) ->
+    describe_flow_logs(FlowIDs, [], default_config()).
+
+-spec describe_flow_logs(ec2_flow_ids(), filter_list()) -> ok_error(proplist());
+                        (ec2_flow_ids(), aws_config()) -> ok_error(proplist());
+                        (ec2_max_result(), ec2_token()) -> ok_error(proplist(), ec2_token()).
```
motobob: implementation for this spec is missing.

motobob: delete spec or add fun

src/erlcloud_ec2.erl
```diff
@@ -2936,6 +2946,168 @@ net_if_params(#ec2_net_if{}=X) ->
...
+                        (ec2_flow_ids(), aws_config()) -> ok_error(proplist());
+                        (ec2_max_result(), ec2_token()) -> ok_error(proplist(), ec2_token()). 
+describe_flow_logs(FlowIDs, Filter)
+    when is_list(FlowIDs), is_list(Filter) orelse Filter =:= none ->
+    describe_flow_logs(FlowIDs, Filter, default_config());
+describe_flow_logs(FlowIDs, Config)
+    when is_list(FlowIDs), is_record(Config, aws_config) ->
+    describe_flow_logs(FlowIDs, [], Config).
+
+-spec describe_flow_logs(ec2_flow_ids(), filter_list(), aws_config()) -> ok_error(proplist());
```
motobob: this /3 is tricky as why not having
```erlang
(ec2_flow_ids(), filter_list(), ec2_token()) -> ok_error(proplist());
or
(ec2_flow_ids(), ec2_max_result(), ec2_token()) -> ok_error(proplist());
```
may be we simplify out /3?

src/erlcloud_ec2.erl
```diff
@@ -2936,6 +2946,168 @@ net_if_params(#ec2_net_if{}=X) ->
...
+extract_flow(Node) ->
+    [
+        {deliver_logs_error_message, get_text("deliverLogsErrorMessage", Node)},
+        {resource_id, get_text("resourceId", Node)},
+        {deliver_logs_permission_arn, get_text("deliverLogsPermissionArn", Node)},
+        {flow_log_status, get_text("flowLogStatus", Node)},
+        {creation_time, erlcloud_xml:get_time("creationTime", Node)},
+        {log_group_name, get_text("logGroupName", Node)},
+        {traffic_type, get_text("trafficType", Node)},
+        {flow_log_id, get_text("flowLogId", Node)}
```
motobob: deliverLogsStatus is missing, no? 
http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_FlowLog.html

src/erlcloud_ec2.erl
```diff
@@ -2936,6 +2946,168 @@ net_if_params(#ec2_net_if{}=X) ->
...
+    end;
+describe_flow_logs(Filter, MaxResults, NextToken)
+    when is_list(Filter) orelse Filter =:= none,
+         is_integer(MaxResults),
+         is_list(NextToken) orelse NextToken =:= undefined ->
+    describe_flow_logs(Filter, MaxResults, NextToken, default_config()).
+
+-spec describe_flow_logs(filter_list(), ec2_max_result(), ec2_token(), aws_config())
+    -> ok_error(proplist(), ec2_token()).
+describe_flow_logs(Filter, MaxResults, NextToken, Config)
```
motobob: so, there will be no `/6`, right?
```erlang
(ec2_flow_ids() , filter_list(), ec2_max_result(), ec2_token(), aws_config()) -> ok_error(proplist(), ec2_token()).
```
asking for 100 known Ids with only 5 returned while filtered by sucessful delivery only sounds intriguingly desired.

motobob:
@nalundgaard welcome to review too

src/erlcloud_ec2.erl
```diff
@@ -185,7 +190,8 @@
 % -define(NEW_API_VERSION, "2013-10-15").
 % -define(NEW_API_VERSION, "2014-02-01").
 % -define(NEW_API_VERSION, "2014-06-15").
--define(NEW_API_VERSION, "2015-10-01").
+% -define(NEW_API_VERSION, "2015-10-01").
```
nalundgaard: does this need to be preserved?

src/erlcloud_ec2.erl
```diff
@@ -2936,6 +2946,168 @@ net_if_params(#ec2_net_if{}=X) ->
...
+        end,
+    P = [
+        {"LogGroupName", LogGroupName},
+        {"ResourceType", RT},
+        {"TrafficType", TT},
+        {"DeliverLogsPermissionArn", DeliverLogsPermissionArn}
+    ] ++ Resources,
+    Params = case ClientToken of
+                none -> P;
+                _ -> P ++ {"ClientToken", ClientToken}
```
nalundgaard: I don't think this does what you want. I would suggest either

```erlang
[{"ClientToken", ClientToken}|P]
```
or
```erlang
P ++ [{"ClientToken", ClientToken}]

File: src/erlcloud_ec2.erl [hunk 9]
```diff
@@ -2936,6 +2946,168 @@ net_if_params(#ec2_net_if{}=X) ->
...
+        {error, _} = Error ->
+            Error
+    end.
+
+%%
+%% Function for making calls to DeleteFlowLogs action
+%% DescribeFlowLogs Documentation: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DeleteFlowLogs.html
+%%
+-spec delete_flow_logs([string()]) -> ok_error(proplist()).
+delete_flow_logs(FlowIDs) ->
```
nalundgaard: `FlowLogIDs`?

src/erlcloud_ec2.erl
```diff
@@ -2936,6 +2946,168 @@ net_if_params(#ec2_net_if{}=X) ->
...
+    describe_flow_logs(FlowIDs, Filter, default_config());
+describe_flow_logs(FlowIDs, Config)
+    when is_list(FlowIDs), is_record(Config, aws_config) ->
+    describe_flow_logs(FlowIDs, [], Config).
+
+-spec describe_flow_logs(ec2_flow_ids(), filter_list(), aws_config()) -> ok_error(proplist());
+                        (filter_list(), ec2_max_result(), ec2_token()) -> ok_error(proplist(), ec2_token()).
+describe_flow_logs(FlowIDs, Filter, Config)
+    when is_list(FlowIDs), is_list(Filter) orelse Filter =:= none , is_record(Config, aws_config) ->
+   Params = erlcloud_aws:param_list(FlowIDs, "FlowLogId") ++ list_to_ec2_filter(Filter),
```
nalundgaard: 3 space indentation?

src/erlcloud_ec2.erl
```diff
@@ -2936,6 +2946,168 @@ net_if_params(#ec2_net_if{}=X) ->
...
+    describe_flow_logs(Filter, MaxResults, NextToken, default_config()).
+
+-spec describe_flow_logs(filter_list(), ec2_max_result(), ec2_token(), aws_config())
+    -> ok_error(proplist(), ec2_token()).
+describe_flow_logs(Filter, MaxResults, NextToken, Config)
+    when is_list(Filter) orelse Filter =:= none,
+         is_integer(MaxResults) andalso MaxResults >= ?FLOWS_MR_MIN andalso MaxResults =< ?FLOWS_MR_MAX,
+         is_list(NextToken) orelse NextToken =:= undefined,
+         is_record(Config, aws_config) ->
+    Params = list_to_ec2_filter(Filter) ++ [{"NextToken", NextToken}, {"MaxResults", MaxResults}],
```
nalundgaard: Should `NextToken` be in the params if it's `undefined`?

src/erlcloud_ec2.erl
```diff
@@ -2936,6 +2946,168 @@ net_if_params(#ec2_net_if{}=X) ->
...
+    when is_list(Filter) orelse Filter =:= none,
+         is_integer(MaxResults),
+         is_list(NextToken) orelse NextToken =:= undefined ->
+    describe_flow_logs(Filter, MaxResults, NextToken, default_config()).
+
+-spec describe_flow_logs(filter_list(), ec2_max_result(), ec2_token(), aws_config())
+    -> ok_error(proplist(), ec2_token()).
+describe_flow_logs(Filter, MaxResults, NextToken, Config)
+    when is_list(Filter) orelse Filter =:= none,
+         is_integer(MaxResults) andalso MaxResults >= ?FLOWS_MR_MIN andalso MaxResults =< ?FLOWS_MR_MAX,
```
nalundgaard: Do we really need to constrain this? What's wrong with letting AWS enforce the range constraints?

src/erlcloud_ec2.erl
ptrakhtman: Not sure how this is useful?

src/erlcloud_ec2.erl
motobob: it will be stripped out down the line so it's fine

src/erlcloud_ec2.erl
motobob: this follows the module structure and why not protect user form oversite?

src/erlcloud_ec2.erl
ptrakhtman: @motobob This follows the same style as describe_instances. Are you proposing completely removing /3 function?

src/erlcloud_ec2.erl
nalundgaard: That seems fine to me, given that it fits the structure. I dislike this for the same reason I don't like putting business logic in clients that get deployed all over the place: if the service properly returns a 400 bad request with the reason your input is out of range, all you've done is disperse redundant validation on every instance of the client. And this isn't for a web UI where you get a tangible benefit from doing local input validation (responsiveness, more client-friendly error output)—instead, you get a `function_clause` error when out of range, and you have to fish through the code, with your inputs in hand, to see what you messed up. I don't see the value, if Amazon's error response is OK for it.

But that's my general approach to these APIs, and given that it's not the approach of this module... it's no big deal to me.

ptrakhtman:
@nalundgaard and @motobob Please review.

src/erlcloud_ec2.erl
ptrakhtman: Addressed

src/erlcloud_ec2.erl
ptrakhtman: I was just trying to follow the existing style. If you prefer to remove older version api, I am happy to do that.

src/erlcloud_ec2.erl
ptrakhtman: Well, yes it is missing because this value isn't never returned by the API.

ptrakhtman:
@nalundgaard and @motobob are I believe I addressed all items from your feedback.

src/erlcloud_ec2.erl
motobob: are you saying AWS doc at ^ is incorrect?

src/erlcloud_ec2.erl
nalundgaard: Seems fine, given that this can be tracked with #369.

src/erlcloud_ec2.erl
```diff
@@ -2936,6 +2946,221 @@ net_if_params(#ec2_net_if{}=X) ->
...
+    case ec2_query(Config, "DescribeFlowLogs", Params, ?NEW_API_VERSION) of
+        {ok, Doc} ->
+            Flows = extract_results("DescribeFlowLogsResponse", "flowLogSet", fun extract_flow/1, Doc),
+            NewNextToken = extract_next_token("DescribeFlowLogsResponse", Doc),
+            {ok, Flows, NewNextToken};
+        {error,  _} = E -> E
+    end.
+
+-spec describe_flow_logs(ec2_flow_ids(), filter_list(), ec2_max_result(), aws_config()) -> ok_error(proplist(), ec2_token()).
+describe_flow_logs(FlowIDs, Filter, MaxResults, Config)
```
SergeyOvchinnik: According to [AWS API](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeFlowLogs.html), `FlowIDs` and `MaxResults` cannot be specified in the same request, so erlcloud should not have an interface for that. 

Instead I recommend you implement the interface `(Filter, MaxResults, NextToken, Config)` so that pagination can be used by making a call `(Filter, MaxResult, undefined, Config)` and making consequent calls using `(Filter, MaxResults, NextToken, Config)` where `NextToken` is the token retrieved from previous calls. Same as it's implemented in `describe_instances/4`.

src/erlcloud_ec2.erl
```diff
@@ -2936,6 +2946,221 @@ net_if_params(#ec2_net_if{}=X) ->
...
+    describe_flow_logs([], MaxResults, Config);
+describe_flow_logs(MaxResults, NextToken)
+    when is_integer(MaxResults) andalso MaxResults >= ?FLOWS_MR_MIN andalso MaxResults =< ?FLOWS_MR_MAX,
+    is_list(NextToken) ->
+    describe_flow_logs(MaxResults, NextToken, default_config()).
+
+-spec describe_flow_logs(ec2_flow_ids(), filter_list(), aws_config()) -> ok_error(proplist());
+                        (ec2_flow_ids(), filter_list(), ec2_max_result()) -> ok_error(proplist());
+                        (filter_list(), ec2_max_result(), aws_config()) -> ok_error(proplist(), ec2_token());
+                        (ec2_max_result(), ec2_token(), aws_config()) -> ok_error(proplist(), ec2_token()).
```
SergeyOvchinnik: I would also like to see interface `(Filter, MaxResults, NextToken)`, so that filtered paginated calls can be made using default config.

src/erlcloud_ec2.erl
```diff
@@ -2936,6 +2946,221 @@ net_if_params(#ec2_net_if{}=X) ->
...
+describe_flow_logs(Config)
+    when is_record(Config, aws_config) ->
+    describe_flow_logs([], Config);
+describe_flow_logs([{_, _} | _] = Filter) ->
+    describe_flow_logs([], Filter, default_config());
+describe_flow_logs(FlowIDs)
+    when is_list(FlowIDs) ->
+    describe_flow_logs(FlowIDs, [], default_config());
+describe_flow_logs(MaxResults)
+    when is_integer(MaxResults) ->
```
ptrakhtman: describe_flow_logs/3 already is checking for the boundary conditions... Wouldn't it be redundant to check here as well?

src/erlcloud_ec2.erl
```diff
@@ -2936,6 +2946,221 @@ net_if_params(#ec2_net_if{}=X) ->
...
+
+-spec describe_flow_logs(ec2_flow_ids(), filter_list()) -> ok_error(proplist());
+                        (ec2_flow_ids(), aws_config()) -> ok_error(proplist());
+                        (filter_list(), aws_config()) -> ok_error(proplist());
+                        (filter_list(), ec2_max_result()) -> ok_error(proplist());
+                        (ec2_max_result(), aws_config()) -> ok_error(proplist(), ec2_token());
+                        (ec2_max_result(), ec2_token()) -> ok_error(proplist(), ec2_token()). 
+describe_flow_logs(FlowIDs, none = Filter) ->
+    describe_flow_logs(FlowIDs, Filter, default_config());
+describe_flow_logs(FlowIDs, [{_, _} | _] = Filter)
```
ptrakhtman: @SergeyOvchinnik I am trying to ensure that the filter is either `none` or a valid list of filter tuples.

src/erlcloud_ec2.erl
ptrakhtman: @SergeyOvchinnik if you provide a token, the filter parameter is invalid. The `NextToken` points to already generated snapshot that was a result of the previous call to describe_flow_logs

src/erlcloud_ec2.erl
```diff
@@ -2936,6 +2946,221 @@ net_if_params(#ec2_net_if{}=X) ->
...
+    when is_integer(MaxResults) andalso MaxResults >= ?FLOWS_MR_MIN andalso MaxResults =< ?FLOWS_MR_MAX,
+    is_record(Config, aws_config) ->
+    describe_flow_logs([], MaxResults, Config);
+describe_flow_logs(MaxResults, NextToken)
+    when is_integer(MaxResults) andalso MaxResults >= ?FLOWS_MR_MIN andalso MaxResults =< ?FLOWS_MR_MAX,
+    is_list(NextToken) ->
+    describe_flow_logs(MaxResults, NextToken, default_config()).
+
+-spec describe_flow_logs(ec2_flow_ids(), filter_list(), aws_config()) -> ok_error(proplist());
+                        (ec2_flow_ids(), filter_list(), ec2_max_result()) -> ok_error(proplist());
```
ptrakhtman: I added missed function instead of removing the spec.

src/erlcloud_ec2.erl
motobob: git does not see it. please push the change.

src/erlcloud_ec2.erl
ptrakhtman: Don't know, but the data isn't returned.

src/erlcloud_ec2.erl
ptrakhtman: I changed to use FlowIDs. The change has been pushed

src/erlcloud_ec2.erl
kkuzmin: this discussion was already held [here](https://github.com/alertlogic/erlcloud/pull/125#discussion-diff-81652895R187) and @motobob, @nalundgaard urged to consider using named types instead. Let's be consistent here as well.

test/erlcloud_ec2_tests.erl
```diff
@@ -82,6 +88,7 @@ common_params() ->
 -type expected_param() :: {string(), string()}.
 -spec validate_param(string(), [expected_param()]) -> [expected_param()].
 validate_param(Param, Expected) ->
+    % ?debugFmt("Param ~p~nExpected ~p", [Param, Expected]),
```
motobob: ?

test/erlcloud_ec2_tests.erl
```diff
@@ -160,6 +167,7 @@ output_test(Fun, {Line, {Description, Response, Result}}) ->
               meck:expect(erlcloud_httpc, request, output_expect(Response)),
               erlcloud_ec2:configure(string:copies("A", 20), string:copies("a", 40)),
               Actual = Fun(),
+              % ?debugFmt("Result ~p~nActual ~p", [Result, Actual]),
```
motobob: ?

motobob:
sorry for a delay with a merge

PR #367, opened 2017-01-09
Added volume.encrypted parsing 

aporubov:
[DescribeVolumes](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeVolumes.html) documentation page specifies `encrypted` property for EBS volume, however, it is missing in  `erlcloud_ec2:describe_volumes/2` output

@motobob please take a look

### DIFF
--- a/src/erlcloud_ec2.erl
+++ b/src/erlcloud_ec2.erl
@@ -2302,6 +2302,9 @@ extract_volume(Node) ->
      {status, get_text("status", Node, none)},
      {create_time, erlcloud_xml:get_time("createTime", Node)},
      {volumeType, get_text("volumeType", Node)},
+     {iops, get_integer("iops", Node)},
+     {encrypted, get_bool("encrypted", Node)},
+     {kms_key_id, get_text("kmsKeyId", Node, none)},
      {attachment_set,
       [[{volume_id, get_text("volumeId", Item)},
         {instance_id, get_text("instanceId",Item)},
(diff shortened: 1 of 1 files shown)


motobob:
while we are here would it be too much of an ask to also add?
iops,
kmsKeyId,
tagSet

aporubov:
@motobob requested properties parsing added

src/erlcloud_ec2.erl
```diff
@@ -2302,6 +2302,9 @@ extract_volume(Node) ->
      {status, get_text("status", Node, none)},
      {create_time, erlcloud_xml:get_time("createTime", Node)},
      {volumeType, get_text("volumeType", Node)},
+     {iops, get_integer("iops", Node)},
+     {encrypted, get_bool("encrypted", Node)},
+     {kms_key_id, get_text("kmsKeyId", Node, none)},
```
motobob: any particular reason for defaulting to `none`?

aporubov: the only reason is these new properties are optional and their default value (`none`) is the same as for the other properties in `volume` object (see `status`, for example). Please note that `iops` is optional as well, and it is defaulted to `0` by `erlcloud_xml: get_integer/2,3`, please see https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_xml.erl#L100-L105

motobob: ah...missed that other use none too

PR #368, opened 2017-01-09
enabled ?NEW_API_VERSION for erlcloud_ec2:describe_addresses/2

aporubov:
#### Problem
`erlcloud_ec2:describe_addresses/2` currently uses ?API_VERSION which is quite dated ("2009-11-30") and thus important fields (i.e. `allocation_id`) are missing in the output. 

#### Solution
- make `erlcloud_ec2:describe_addresses/2` use ?NEW_API_VERSION (currently "2015-10-01")
- update public IP address object parsing

@motobob please take a look


motobob:
👍

PR #371, opened 2017-01-10
Fix #370

### LINKED ISSUE #370: Cannot use filters in erlcloud_ec2:describe_instances/4
The function clause inhibits use of the filter function.

`erlcloud_ec2:describe_instances([{"tag:Cluster", ["clustername"]}], undefined, undefined, Config).`

results in a `function_clause` error.

When setting the `MaxResults` to 5 (the minimum) I get an AWS EC2 error:

`{error,{http_error,400,"Bad Request", <<"<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Response><Errors><Error><Code>InvalidParameterCombination</Code><Message>The parameter 'maxResults' cannot be used with tag filters in the parameter 'filterSet'. Remove either the tag filters from 'filterSet' or the 'maxResults' parameter and try again.</Message></Error></Errors><RequestID>6d578aa5-ebd2-4a09-8b25-e27c505e87fd</RequestID></Response>">>}}`

I can fix this by adding expanding the guard clause to this, to allow a filter with an undefined MaxResults:

```
describe_instances(Filter, MaxResults, NextToken, Config)
    when is_list(Filter) orelse Filter =:= none,
         is_integer(MaxResults) andalso MaxResults >= ?INSTANCES_MR_MIN andalso MaxResults =< ?INSTANCES_MR_MAX orelse Filter =/= none andalso MaxResults =:= undefined,
         is_list(NextToken) orelse NextToken =:= undefined,
         is_record(Config, aws_config) ->
```

I will try and make a pull request.

jvdvleuten:
See https://github.com/erlcloud/erlcloud/issues/370

### DIFF
--- a/src/erlcloud_ec2.erl
+++ b/src/erlcloud_ec2.erl
@@ -1351,7 +1351,8 @@ describe_instances(Filter, MaxResults, NextToken)
     -> ok_error(proplist(), ec2_token()).
 describe_instances(Filter, MaxResults, NextToken, Config)
     when is_list(Filter) orelse Filter =:= none,
-         is_integer(MaxResults) andalso MaxResults >= ?INSTANCES_MR_MIN andalso MaxResults =< ?INSTANCES_MR_MAX,
+         is_integer(MaxResults) andalso MaxResults >= ?INSTANCES_MR_MIN andalso MaxResults =< ?INSTANCES_MR_MAX
+	 orelse Filter =/= none andalso MaxResults =:= undefined,
          is_list(NextToken) orelse NextToken =:= undefined,
          is_record(Config, aws_config) ->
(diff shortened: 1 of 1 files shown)


SergeyOvchinnik:
As I mentioned in #370, allowing `MaxResult` to be `undefined` in `describe_instances/4` to make calls with just `Filter` and `Config` is the same as using a non-paginated clause `describe_instances/3` with `Filter`, `Config` and using `[]` for `InstanceIDs`.

As for filters, Filters work with pagination anyway, all except tag filters (for [AWS API](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstances.html) reasons).

This addition to interface just adds another way to make calls that were supported anyway.

PR #372, opened 2017-01-12
Allow setting SNS MessageAttributes parameter.

kzemek:
Adds support for setting message attributes as described in https://docs.aws.amazon.com/sns/latest/api/API_Publish.html


src/erlcloud_sns.erl
```diff
@@ -57,6 +57,8 @@
                                 | enabled
                                 | token.
 
+-type sns_message_attributes() :: #{string() => string() | binary() | number()}.
```
motobob: the repo must be 16.3 compatible. please use proplist.

motobob:
thank you for contribution.

PR #373, opened 2017-01-12
erlcloud_ec2:describe_addresses/3

aporubov:
#### Problem
[AWS DescribeAddresses API](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeAddresses.html) allows to specify a number of filters when requesting a list of public IP addresses, however, `erlcloud` doesn't expose any interface for that at the moment. 

#### Solution
added `erlcloud_ec2:describe_addresses/3` with filters support.

@motobob please take a look


motobob:
lgtm

PR #374, opened 2017-01-17
Support tagging for kinesis streams

arthurnovak:
Implemented `add_tags`, `list_tags`, `list_all_tags` and `remove_tags` functionality for `erlcloud_kinesis`.

Please review @motobob @nalundgaard @edhollandAL

### DIFF
--- a/src/erlcloud_kinesis.erl
+++ b/src/erlcloud_kinesis.erl
@@ -14,7 +14,12 @@
          put_record/3, put_record/4, put_record/5, put_record/6, put_record/7,
          put_records/2, put_records/3,
          merge_shards/3, merge_shards/4,
-         split_shards/3, split_shards/4
+         split_shards/3, split_shards/4,
+         add_tags_to_stream/2, add_tags_to_stream/3,
+         list_tags_for_stream/1, list_tags_for_stream/2,
+         list_tags_for_stream/3, list_tags_for_stream/4,
+         list_all_tags_for_stream/1, list_all_tags_for_stream/2,
+         remove_tags_from_stream/2, remove_tags_from_stream/3
         ]).
 
 -include("erlcloud.hrl").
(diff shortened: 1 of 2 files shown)


src/erlcloud_kinesis.erl
```diff
@@ -560,3 +563,100 @@ split_shards(StreamName, ShardToSplit, NewStartingHashKey) ->
...
+
+%%------------------------------------------------------------------------------
+%% @doc
+%% Kinesis API: ListTagsForStream
+%% Lists the tags for the specified Amazon Kinesis stream.
+%%
+%% erlcloud_kinesis:list_tags(<<"stream_name">>, <<"key1">>, 1).
+%% {ok, [
+%%       {<<"HasMoreTags">>, true},
+%%       {<<"Tags">>,        [[{<<"Key">>,   <<"key2">>},
```
edholland: is this a proplist or a list of proplists?

arthurnovak: list of proplists
```
{ok,[{<<"HasMoreTags">>,false},
     {<<"Tags">>,
      [[{<<"Key">>,<<"k1">>},{<<"Value">>,<<"v1">>}],
       [{<<"Key">>,<<"k3">>},{<<"Value">>,<<"v3">>}]]}]}
```
Hm, wondering if it makes sense to be more specific in spec

arthurnovak: Or just update sample

edholland: why do we need list of proplists here at all? It looks like the api response is just a list
```
{
   "HasMoreTags": boolean,
   "Tags": [ 
      { 
         "Key": "string",
         "Value": "string"
      }
   ]
}
```

arthurnovak: Right, this is how erlang decodes maps

edholland: My concern is that i can't just pipe result from `list_tags` back into `add_tags` call if i want to nondestructively add a tag. 
I think the api should support something like
```
{ok, Tags} = erlcloud_kinesis:get_tags(<<"stream">>),
erlcloud_kinesis:add_tags([{<<"new_tag">>, <<"new_Value">>} | Tags])
```
I guess this is fundamentally inconsistency AWS side since the expect tags as
```
{
  "StreamName": "exampleStreamName",
  "Tags": {
     "Project" : "myProject",
     "Environment" : "Production"
   }
}
```
but return them as
```
{
   "HasMoreTags": boolean,
   "Tags": [ 
      { 
         "Key": "string",
         "Value": "string"
      }
   ]
}
```
But i think we can smooth this in the library api

src/erlcloud_kinesis.erl
```diff
@@ -560,3 +563,100 @@ split_shards(StreamName, ShardToSplit, NewStartingHashKey) ->
...
+%%                             {<<"Value">>, <<"value">>}]]}
+%%      ]
+%% }.
+%%
+%% @end
+%%------------------------------------------------------------------------------
+
+-spec list_tags(binary()) -> {ok, proplist()} | {error, any()}.
```
edholland: i wonder if we should also add a `list_all_tags` function that hides pagination logic from caller?

arthurnovak: Agree, `list_all_tags` should be the right solution here.
Regarding previous comment, I don't think we have to force changing AWS response, and `HasMoreTags` has its practical value in some use cases. In case of `list_tags` we just return what we get from AWS.

edholland: I guess it would somewhat violate principle of least surprise to modify the shape of aws response. I wonder if we could solve my "confusing api" issue by supporting nested proplists as input for `add_tags`?

arthurnovak: `list_all_tags` should do it. It will return list of tags only in format they are put in `add_tags` function. So this should work
```
{ok, OldTags} = erlcloud_kinesis:get_all_tags(<<"stream">>),
erlcloud_kinesis:add_tags([{<<"new_tag">>, <<"new_Value">>} | OldTags])
```

src/erlcloud_kinesis.erl
```diff
@@ -560,3 +564,117 @@ split_shards(StreamName, ShardToSplit, NewStartingHashKey) ->
...
+%% erlcloud_kinesis:list_all_tags(<<"stream_name">>).
+%% {ok, [{<<"k1">>, <<"v1">>},
+%%       {<<"k2">>, <<"v2">>}]
+%% }.
+%%
+%% @end
+%%------------------------------------------------------------------------------
+-spec list_all_tags(binary()) -> {ok, proplist()} | {error, any()}.
+list_all_tags(StreamName) ->
+    case erlcloud_kinesis:list_tags(StreamName) of
```
nalundgaard: seems like an antipattern to do a module external call to the local module.

src/erlcloud_kinesis.erl
```diff
@@ -560,3 +564,117 @@ split_shards(StreamName, ShardToSplit, NewStartingHashKey) ->
...
+        {ok, [_HasMoreTags, {<<"Tags">>, Tags}]} ->
```
nalundgaard: don't you need a recursive call when `HasMoreTags` is `true`? This looks broken.

src/erlcloud_kinesis.erl
```diff
@@ -560,3 +564,117 @@ split_shards(StreamName, ShardToSplit, NewStartingHashKey) ->
 split_shards(StreamName, ShardToSplit, NewStartingHashKey, Config) when is_record(Config, aws_config) ->
   Json = [{<<"StreamName">>, StreamName}, {<<"ShardToSplit">>, ShardToSplit}, {<<"NewStartingHashKey">>, NewStartingHashKey}],
   erlcloud_kinesis_impl:request(Config, "Kinesis_20131202.SplitShard", Json).
+
+%%------------------------------------------------------------------------------
+%% @doc
+%% Kinesis API: AddTagsToStream
```
nalundgaard: Can you use a link, like [http://docs.aws.amazon.com/kinesis/latest/APIReference/API_AddTagsToStream.html](http://docs.aws.amazon.com/kinesis/latest/APIReference/API_AddTagsToStream.html), on all your references? I don't understand the divergence from the module style here.

src/erlcloud_kinesis.erl
```diff
@@ -560,3 +564,117 @@ split_shards(StreamName, ShardToSplit, NewStartingHashKey) ->
...
+%% Adds or updates tags for the specified Amazon Kinesis stream.
+%%
+%% erlcloud_kinesis:add_tags(<<"stream_name">>,
+%%                           [{<<"tag_key">>, <<"tag_value">>}]).
+%% ok.
+%%
+%% @end
+%%------------------------------------------------------------------------------
+-spec add_tags(binary(), [{binary(), binary()}]) -> ok | {error, any()}.
```
nalundgaard: I don't like the break with the convention of converting the TitleCase function names to underscore_case. I would just call this `add_tags_to_stream`.

src/erlcloud_kinesis.erl
arthurnovak: yes, AWS API site didn't work yesterday

src/erlcloud_kinesis.erl
arthurnovak: I don't think we really need that here. Max number of tags is limited by 10 only. If Limit parameters is not specified that the response will include all tags in it

src/erlcloud_kinesis.erl
arthurnovak: added it accidentally, will remove

src/erlcloud_kinesis.erl
arthurnovak: do we really need that complexity? `erlcloud_kinesis:add_tags` already gives a clear understanding that those tags related to kinesis stream imo. What do you think?

src/erlcloud_kinesis.erl
nalundgaard: ["Each stream can have up to 10 tags"](http://docs.aws.amazon.com/kinesis/latest/APIReference/API_AddTagsToStream.html)

Ah, I see. Had to piece that together a bit.

In that case, I kind of wonder if this function is of that much value?

src/erlcloud_kinesis.erl
arthurnovak: If we decide to change to 'proper' convention, should this rule be also applied to list_tags, remove_tags?

nalundgaard: Yes, that is what I would suggest. I think the value of being consistent with the AWS API name is much higher than the value of succinctness.

motobob: @arthurnovak , i'm on @nalundgaard side? ;)

src/erlcloud_kinesis.erl
motobob: so this uses default list fun and i bet it uses max as default. please check and if so this fun is not necessary.

arthurnovak: Yes, it uses max as default, but I suggest to have `list_all_tags` interface anyway.
The idea of introducing `list_all_tags` was to have response in format `{ok, TagsList}`, where `TagsList` is in the same format as we put it in `add_tags(Stream, TagsList)`, so simple key-value tuple.
`list_tags` is default request to AWS and the response here is `{ok, HasMoreTags, TagsProplist}`, which makes it dodgy to use in the following case:
```
{ok, OldTags} = erlcloud_kinesis:get_tags(<<"stream">>),
erlcloud_kinesis:add_tags([{<<"new_tag">>, <<"new_Value">>} | OldTags])
```

src/erlcloud_kinesis.erl
```diff
@@ -560,3 +565,130 @@ split_shards(StreamName, ShardToSplit, NewStartingHashKey) ->
...
+%% Wrapper around 'list_tags_for_stream' to return {ok, <list_of_all_tags>}.
+%%
+%% erlcloud_kinesis:list_all_tags_for_stream(<<"stream_name">>).
+%% {ok, [{<<"k1">>, <<"v1">>},
+%%       {<<"k2">>, <<"v2">>}]
+%% }.
+%%
+%% @end
+%%------------------------------------------------------------------------------
+-spec list_all_tags_for_stream(binary()) -> {ok, proplist()} | {error, any()}.
```
edholland: if we decide to keep this we need a version that accept an aws config also

arthurnovak: added

src/erlcloud_kinesis.erl
```diff
@@ -560,3 +565,135 @@ split_shards(StreamName, ShardToSplit, NewStartingHashKey) ->
...
+list_all_tags_for_stream(StreamName) ->
+    list_all_tags_for_stream(StreamName, default_config()).
+
+-spec list_all_tags_for_stream(binary(), aws_config()) ->
+    {ok, proplist()} | {error, any()}.
+list_all_tags_for_stream(StreamName, Config) ->
+    case list_tags_for_stream(StreamName, Config) of
+        {ok, [_HasMoreTags, {<<"Tags">>, Tags}]} ->
```
edholland: we probably shouldn't depend on specific structure of aws response here, they may add fields in future

src/erlcloud_kinesis.erl
motobob: i still think this fun is broken as it just ignores `HasMoreTags`.
let's remove it and i'll merge.

src/erlcloud_kinesis.erl
```diff
@@ -560,3 +565,136 @@ split_shards(StreamName, ShardToSplit, NewStartingHashKey) ->
...
+    ok | {error, any()}.
+remove_tags_from_stream(StreamName, TagKeys, Config)
+  when is_record(Config, aws_config) ->
+    Json      = [{<<"StreamName">>, StreamName},
+                 {<<"TagKeys">>,    TagKeys}],
+    Operation = "Kinesis_20131202.RemoveTagsFromStream",
+    Response  = erlcloud_kinesis_impl:request(Config, Operation, Json),
+    normalize_tags_response(Response).
+
+normalize_tags_response({ok, []}) -> ok;
```
motobob: and what is Rsp is `{ok, NonEmptyList}` ?

arthurnovak: It cannot be non empty:
```
Response Elements
If the action is successful, the service sends back an HTTP 200 response with an empty HTTP body.
```

nalundgaard:
@arthurnovak thanks!

PR #376, opened 2017-01-20
Make erlcloud_sns:publish dialyzer friendly

rslota:
In order to make dialyzer pass checks when using erlcloud_sns:publish, some types needed correction. `retry_fun` just seems to be outdated, while `aws_request_form_raw` cannot expect `Scheme` to be `string()` since one of possible arguments is `[Protocol, "://"]` (`string()` type is equal to `[integer()]`).


motobob:
thanks for contribution.

Issue #303, opened 2016-08-30
Support for Kinesis Firehose


derek121:
Is there any planned support for [Firehose](http://docs.aws.amazon.com/firehose/latest/APIReference/Welcome.html)?

motobob:
@derek121 I'm not aware of anybody working on adding this functionality at the moment but it's definitely worth adding given it becomes very popular. 
If you have some time a PR is more than welcome.

derek121:
Thanks. I wound up adding the couple functions from the Firehose API that I need on a local fork for now- if I'm able to make the time then I'll complete it and submit a PR.

onepremise:
Since nothing recently came of this, went ahead and added the API changes. See referenced pull request #377 for review.

pankajsoni19:
Is there any action on this? I am particularly looking to use the PutRecord in firehose data stream.

PR #377, opened 2017-01-26
Firehose updates

onepremise:
I started adding support for firehose access to the erlcloud api. The erlcloud_firehose_impl works. erlcloud_firehose module soon to follow.


src/erlcloud_firehose_impl.erl
```diff
@@ -0,0 +1,177 @@
+%% -*- mode: erlang;erlang-indent-level: 4;indent-tabs-mode: nil -*-
```
edholland: this seems to be duplicate from `erlcloud_kinesis_impl`, can we not reuse that implementation here?

onepremise: There just a few differences in configuration. I can consolidate there.

onepremise: Would you like the firehose calls in erlcloud_kinesis or separate module as well?

motobob: from _impl module perspective I see no value in having a separate one. 
To have a better picture is would be nice to see full firehose thing as it might be that separating out _impl module is unnecessary at all and be put directly into `erlcloud_firehose`

onepremise: Sorry for the delay, I added the erlcloud_firehose module. I'm going to go back and refactor the _impl into the module so we don't have the duplicate impl for kinesis and firehose.

onepremise:
Requested impl files consolidated. Pull-request is ready for review. Let me know if there are any further updates.

onepremise:
updating tests.

onepremise:
Updates complete. Also noticed a pre-existing ticket open against this, Support for Kinesis Firehose #303. This should cover initial implementation. Let me know if there are any further updates.

onepremise:
Thanks ykersule, added missing service_config entry. Not sure about the ExplicitHashKey/PartitionKey stuff, I didn't see anything applicable in the aws doc:

http://docs.aws.amazon.com/firehose/latest/APIReference/API_PutRecord.html

I may defer that change for now, unless it is found to be a certain issue.  Thanks for input!

onepremise:
Just wanted to check and see if there's anything further you need from me. Otherwise, this should be good to go.

rebar.config
```diff
@@ -1,8 +1,8 @@
 %% -*- mode: erlang;erlang-indent-level: 4;indent-tabs-mode: nil -*-
-{require_otp_vsn, "R16|17|18|19"}.
+{require_otp_vsn, "R16|17|18|19|20|21|22"}.
 {erl_opts,
  [debug_info,
-  {platform_define, "^19.", 'ERLANG_OTP_VERSION_19'},
+  {platform_define, "^(19|2[0-9])", 'ERLANG_OTP_VERSION_19'},
```
nalundgaard: Seems like this makes the macro name a little misleading. Maybe `ERLANG_19_PLUS` would be more succinct and accurate?

PR #378, opened 2017-01-27
Provide partial EMR support

mrallen1:
This is the beginning of Elastic Map Reduce support. Currently the four API calls which are supported are `AddJobStep`, `RunJobFlow`, `TerminateJobFlow`, and `DescribeSteps`.  This is the "minimum viable" amount of support needed to use EMR.

This PR supersedes #348 - since that PR, the code has been rebased against current master and a test added, as per comments.


nalundgaard:
Please make sure and add this to [`erlcloud_aws:service_config/3` and/or `erlcloud_aws:service_host/2`](https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_aws.erl#L495-L632) as appropriate.

include/erlcloud_aws.hrl
```diff
@@ -30,6 +30,9 @@
           sqs_host="queue.amazonaws.com"::string(),
           sqs_protocol=undefined::string()|undefined,
           sqs_port=undefined::non_neg_integer()|undefined,
+          emr_scheme="https://"::string()|undefined,
+          emr_host="elasticmapreduce.amazonaws.com"::string(),
```
nalundgaard: should this be `elasticmapreduce.us-east-1.amazonaws.com`?

mrallen1:
How's this look?

nalundgaard:
Am I reading correctly that the tests simply cover the common request mechanics and the run_job_flow function? Any reason to not have I/O tests for the other added functions?

mrallen1:
You're reading correctly. 

I don't understand the utility of adding more mocked tests. At best they are ensuring that jsx can serialize and deserialize JSON correctly. At worst they are a false sense of security that real-world Amazon services behave as the mocked HTTP calls.

Are there tests that do not involve mocks and make live service calls? I'd be happy to write some of those for EMR.

mrallen1:
OK, I mocked out the functions which have example input/output data in the API docs. I think that's all the tests which are feasible given the docs.

test/erlcloud_emr_tests.erl
```diff
@@ -0,0 +1,207 @@
...
+     [{<<"Jar">>,
+       <<"s3n:\\/\\/elasticmapreduce\\/samples\\/cloudburst\\/cloudburst.jar">>},
+      {<<"Args">>,
+       [<<"s3n:\\/\\/elasticmapreduce\\/samples\\/cloudburst\\/input\\/s_suis.br">>,
+        <<"s3n:\\/\\/elasticmapreduce\\/samples\\/cloudburst\\/input\\/100k.br">>,
+        <<"s3n:\\/\\/examples-bucket\\/cloudburst\\/output">>,<<"36">>,
+        <<"3">>,<<"0">>,<<"1">>,<<"240">>,<<"48">>,<<"24">>,<<"24">>,
+        <<"128">>,<<"16">>]}]}]]}]).
+
+%% AddJobFlowSteps doesn't have an example output format.
```
nalundgaard: that is wild, it looks like there is a response syntax that includes [StepIds](http://docs.aws.amazon.com/ElasticMapReduce/latest/API/API_AddJobFlowSteps.html#API_AddJobFlowSteps_ResponseSyntax). I guess their docs guy fell asleep at the wheel.

nalundgaard: Would it be reasonable to expect output like
```json
{"JobFlowIds": ["j-3TS0OIYO4NFN"]}
```
based on the doc?

mrallen1:
Really making me work for it. OK, how about now?

nalundgaard:
👍 LGTM. @motobob or @waisbrot any objections? If not, I'll merge this later today.

Issue #347, opened 2016-12-02
erlcloud_cloudformation:describe_stacks_all() doesn't return "Parameters"


jonathn-al:
erlcloud_cloudformation:describe_stacks_all() doesn't return the stack's "Parameters" similar to the return value from 'aws cloudformation describe-stacks --stack-name teststack'
(it does return the stack's "Outputs" as expected).

motobob:
Confirm that parsing in this module is greatly outdated and simplified. 
thanks for reporting. 
If you have a change ready - PR is welcome.
Otherwise we will add it to our backlog ti right after all sigV4 updates several other changes planned.

PR #381, opened 2017-01-29
Fix bug #347 missing parameters in the describe_stacks_all result

posilva:
- Returning paramters
- Outputs are returning also Description


posilva:
In my opinion it would be better if we could remove the "member" element from result of "Outputs" and "Parameters". It's better to access directly to the list of parameters avoiding the nested "member" element.

nalundgaard:
👍 from me. I agree that the "member" key is ugly, but I think it's best to leave it in. Unexpected asymmetry seems worse than elegant, unexpected exceptions. @motobob any thoughts?

motobob:
fully agree with keeping `member`

PR #386, opened 2017-02-09
describe_stacks fix

EuMcG:
Problem:
The parameter of describe_stacks was never converted into the proper aws format, so passing a list of:
[{stack_name, StackName}] would return an error and in the layout of [{"stack_name", StackName}] would just return everything. 

Solution:
Convert params before passing to aws.

### DIFF
--- a/src/erlcloud_cloudformation.erl
+++ b/src/erlcloud_cloudformation.erl
@@ -261,7 +261,6 @@ get_stack_policy(Params, StackName) ->
 -spec get_stack_policy(params(), string(), aws_config()) ->
     {ok, cloudformation_list()} | {error, error_reason()}.
 get_stack_policy(Params, StackName, Config = #aws_config{}) ->
-
     FullParams = [{"StackName", StackName}
         | lists:map(
             fun(T) ->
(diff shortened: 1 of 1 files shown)


EuMcG:
test it

src/erlcloud_cloudformation.erl
```diff
@@ -405,7 +404,8 @@ convert_param({Key, Value}) ->
         stack_name -> {"StackName", Value};
         stack_id -> {"StackName", Value};
         logical_resource_id -> {"LogicalResourceId", Value};
-        next_token -> {"NextToken", Value}
+        next_token -> {"NextToken", Value};
+        _ -> {Key, Value}
```
motobob: do we really need this?
as we may just ask for for all params to be actual `{"", ""}`

SergeyOvchinnik: That's what erlcloud should do ideally - support param names only in the same format they are written in the API. But it seems like this module has support for atoms as param names in a few functions, so it might make sense to implement it for the sake of module consistency. As long as `{"", ""}` is still fully supported, I think it's fine

PR #390, opened 2017-02-15
Rewritten some erlcloud_ec2:describe_flow_logs clauses

SergeyOvchinnik:
Problem:

- Current implementation of `erlcloud_ec2:describe_flow_logs` has some clauses that violate API rule about not using `FlowIDs` and `MaxResults` together and some clauses are missing

Solution:

- Removed all clauses where `FlowIDs` and `MaxResults` were used together
- Added missing clauses
- Minor fixes

Reviewers:

@motobob


motobob:
tnx for cleanup

PR #391, opened 2017-02-16
Update erlcloud_mon.erl

gustavotiago-moymer:
When passing my aws_config as param, I got the timeout error related to the credentials. I had a look on the code and thought the line I removed was making my param being ignored.


motobob:
that is correct. sorry for that unfortunate defect. there is lots of legacy code here to be cleaned up

gustavotiago-moymer:
Thank you for the reply and no problem, I just thought that could be something to fix.

PR #394, opened 2017-03-10
Allow profile name on auto_config/1

fp:
Allow a named profile to be used instead of the `default` when
calling auto_config/1. The order of credential resolution does not
change, but if a credentials profile is selected, then the named
profile will be used.


motobob:
👍

Issue #398, opened 2017-03-17
erlcloud_aws:default_config() is broken


KingBrewer:
It seems that `erlcloud_aws:default_config()` is broken. It crashes whenever `AWS_REGION` environment variable is set.
```erlang
erlcloud_aws:default_config().
** exception error: no function clause matching erlcloud_aws:service_config(<<"cloudfront">>,<<"us-east-1">>,
                                                                            {aws_config,"autoscaling.us-east-1.amazonaws.com",
                                                                                        "ec2.us-east-1.amazonaws.com","iam.amazonaws.com",
                                                                                        "sts.us-east-1.amazonaws.com","https://",
                                                                                        "s3-external-1.amazonaws.com",80,false,2,vhost,false,
                                                                                        "sdb.amazonaws.com",
                                                                                        "elasticloadbalancing.us-east-1.amazonaws.com",
                                                                                        "rds.us-east-1.amazonaws.com",
                                                                                        "email.us-east-1.amazonaws.com",
                                                                                        "sqs.us-east-1.amazonaws.com",undefined,undefined,
                                                                                        "https://","elasticmapreduce.us-east-1.amazonaws.com",
                                                                                        undefined,"http://","sns.us-east-1.amazonaws.com",
                                                                                        "mechanicalturk.amazonaws.com",
                                                                                        [...],...}) (src/erlcloud_aws.erl, line 515)
     in function  lists:foldl/3 (lists.erl, line 1248)
```

Probably the crash is caused by missing `<<"cloudfront">>` clause in this function:
https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_aws.erl#L515

PR #388, opened 2017-02-13
added cloudfront service (get_distribution, list_distributions, list_distributions_all)

andrey-smirnov:
CloudFront GetDistribution and ListDistributions API calls implemented
@aporubov @motobob

PR #399, opened 2017-03-17
Filter clause for cloudfront service_config

KingBrewer:
Apparently CloudFront PR https://github.com/erlcloud/erlcloud/pull/388 broke `default_config` function.
Fixes: https://github.com/erlcloud/erlcloud/issues/398

Filter is required, because cloudfront endpoint is agnostic to region (http://docs.aws.amazon.com/general/latest/gr/rande.html#cf_region).

### DIFF
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -530,6 +530,7 @@ service_config( <<"cloudformation">> = Service, Region, Config ) ->
     Config#aws_config{ cloudformation_host = Host };
 service_config( <<"cfn">>, Region, Config ) ->
     service_config( <<"cloudformation">>, Region, Config );
+service_config( <<"cloudfront">>, _Region, Config ) -> Config;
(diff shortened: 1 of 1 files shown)


nalundgaard:
@KingBrewer This looks like it would work, but it doesn't seem to conform with the method used to process other region-agnostic services, like IAM, etc.

I'd prefer to see a cloudfront clause in `erlcloud_aws:service_config/3` [like this](https://github.com/erlcloud/erlcloud/blob/24f46264d159089de65c28a943c6fcd1be0a1874/src/erlcloud_aws.erl#L571), personally. @motobob what do you think?

KingBrewer:
@nalundgaard I completely agree with you. To be fair I wasn't completely aware of this "method". I was just following the crash stacktrace and found this filter.

KingBrewer:
Voila

nalundgaard:
👍 Thanks for your quick fix @KingBrewer!

PR #403, opened 2017-03-30
Match 'US' and 'EU' S3 locations to us-east-1 and eu-west-1

ikemsley:
**Problem**

S3 bucket location attributes of "", "US" and "EU" do not map correctly to "us-east-1" and "eu-west-1".

**Solution**

Add "/LocationConstraint" attribute handling to erlcloud_s3.erl for "", "US" (maps to "us-east-1") and "EU" (maps to "eu-west-1").

### DIFF
--- a/src/erlcloud_s3.erl
+++ b/src/erlcloud_s3.erl
@@ -550,9 +550,9 @@ get_bucket_attribute(BucketName, AttributeName, Config)
             case erlcloud_xml:get_text("/LocationConstraint", Doc) of
                 %% logic according to http://s3tools.org/s3cmd
                 %% s3cmd-1.5.2/S3/S3.py : line 342 (function get_bucket_location)
-                [] -> "us-east-1";
-                ["US"] -> "us-east-1";
-                ["EU"] -> "eu-west-1";
+                "" -> "us-east-1";
+                "US" -> "us-east-1";
+                "EU" -> "eu-west-1";
                 Loc -> Loc
             end;
         logging ->


kkuzmin:
+1. Thanks @ikemsley

kkuzmin:
> Valid Values: [ us-west-1 | us-west-2 | ca-central-1 | EU or eu-west-1 | eu-west-2 | eu-central-1 | ap-south-1 | ap-southeast-1 | ap-southeast-2 | ap-northeast-1 | ap-northeast-2 | sa-east-1 | empty string (for the US East (N. Virginia) region) | us-east-2]

From here http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGETlocation.html

nalundgaard:
@kkuzmin @ikemsley according to the doc, it looks like `["US"]` and `["EU"]` are just plain bugs—the document does not describe returning an array attribute under any circumstance. Moreover, there is nothing about mapping `"US"` to `"us-east-1"`—seems like that is only for `""`. Should this be changed to this?

```erlang
case erlcloud_xml:get_text("/LocationConstraint", Doc) of
    "" -> "us-east-1";
    "EU" -> "eu-west-1";
    Loc -> Loc
end
```

nalundgaard:
Note also that this is apropos of #199 by @izinin. 

It looks like the [latest version of s3cmd](https://github.com/s3tools/s3cmd/blob/master/S3/S3.py#L446-L451) works like this:

```python
        location = getTextFromXml(response['data'], "LocationConstraint")
        if not location or location in [ "", "US" ]:
            location = "us-east-1"
        elif location == "EU":
            location = "eu-west-1"
        return location
```

So maybe we should keep the `"US"` thing, but I think the `["US"]` and `["EU"]` matches are just BS.

kkuzmin:
@nalundgaard this is the first time we face with `"EU"` region returned by AWS in production. I can't give you examples of other cases.
+1 to the implementation you've suggested + `"US"` case and removing `["EU"]`, `["US"]` ones.

ikemsley:
@nalundgaard @kkuzmin @motobob How about now?

kkuzmin:
+1

izinin:
Hi guys !
You have introduced bug in the function . I think next time i build my code it is got broken until i specify version that is not broken . 

here is my testing :
(ejabberd@eu1.mim.jongla.im)8> erlcloud_s3:get_bucket_attribute("eu-jongla", location).
"eu-west-1"
(ejabberd@eu1.mim.jongla.im)9> erlcloud_s3:get_bucket_attribute("coverimagedevsg", location).
"ap-southeast-1"
(ejabberd@eu1.mim.jongla.im)10> erlcloud_s3:get_bucket_attribute("na-jongla", location).      
"us-west-1"

kkuzmin:
Hi @izinin.

From the example above it is not clear to me what is the actual issue.
Could you please elaborate more about expected output?

Thanks in advance.

izinin:
Hi @kkuzmin !

this assumption from the message thread is wrong. and match patterns in the `case` operator should be array string , not just string. the latter will not match 

> @kkuzmin @ikemsley according to the doc, it looks like ["US"] and ["EU"] are just plain bugs—the document does not describe returning an array attribute under any circumstance. Moreover, there is nothing about mapping "US" to "us-east-1"—seems like that is only for "". Should this be changed to this?

> case erlcloud_xml:get_text("/LocationConstraint", Doc) of
    "" -> "us-east-1";
    "EU" -> "eu-west-1";
    Loc -> Loc
end

Please test you submission on different bucket location

Regards,
Igor

kkuzmin:
@izinin According to current implementation cases for `["EU"]` and `["US"]` should be returned as is.
Do you have a real life examples when erlcloud returns them? I was not able to get such.
The reason for this change was our experience when erlcloud returned `"EU"` as a bucket location.

izinin:
@kkuzmin , i posted output from live environment . Yes i use erlcloud_s3:get_bucket_attribute on production and now it is working as it should.

kkuzmin:
@izinin the output from the live environment you've posted looks good to me because it doesn't contain any `["EU"]`/`["US"]` returns.
I think if a consumer relies on `erlcloud_s3:get_bucket_attribute(location)` returns `"EU"`/`"US"` then it should be fixed on a consumer side. But in general it shouldn't be a problem because a consumer of this API should normally expect regular region names, like "us-east-1".

PR #410, opened 2017-06-08
route53 resource sets pagination fixed

andrey-smirnov:
Route53 record sets api missing part of pagination token. Fixed.

### DIFF
--- a/src/erlcloud_route53.erl
+++ b/src/erlcloud_route53.erl
@@ -182,8 +182,9 @@ describe_zones_all(Options, Config) when is_list(Options),
 %% --------------------------------------------------------------------
 -spec describe_resource_sets(ZoneId :: string()) ->
              {ok, list(aws_route53_resourceset())} |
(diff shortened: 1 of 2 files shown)


src/erlcloud_route53.erl
```diff
@@ -343,12 +346,16 @@ do_describe_resource_sets(ZoneId, Params, AwsConfig) ->
...
-                         {undefined, undefined} ->
-                             erlcloud_xml:get_text(
-                               ?DESCRIBE_RS_NEXT_ID, Doc);
-                         {NextName, NextType} ->
-                             {NextName, NextType}
+                                                 Doc, undefined),
+                           erlcloud_xml:get_text(?DESCRIBE_RS_NEXT_ID,
+                                                 Doc, undefined)
+						 } of
+                         {NextName, NextType, undefined} ->
```
motobob: three params have more than three permutations ;)

andrey-smirnov: there cannot be name or type missing if other one is present

PR #411, opened 2017-06-30
Cloudwatch logs

rvangraan:
@motobob I've created two functions that are needed to send log events into cloudwatch. I've had to change the way the `cw_request` function works somewhat, as it didn't support complex requests such as the nested log events. So now it supports returning a map or a list (as it used to do). 

The `put_log_events` function returns a map of the AWS response:
```
erlcloud_cloudwatch_logs:put_log_events(_LogGroup = <<"loggroup">>,
                                        _LogStream = <<"logstream">>,
                                        _SequenceToken = <<"49574602524945087512519741331171438190807382809047470706">>,
                                        _Events = [#{message=><<"aaaa">>, timestamp=>11}],Config).
{ok,#{nextSequenceToken =>
          <<"49574602524945087512519741869202665284478508564977029746">>,
      rejectedLogEventsInfo => #{tooOldLogEventEndIndex => 1}}}
```

### DIFF
--- a/src/erlcloud_cloudwatch_logs.erl
+++ b/src/erlcloud_cloudwatch_logs.erl
@@ -31,6 +31,15 @@
 
 -type log_group() :: jsx:json_term().
 
+-type log_group_name() :: binary().
+-type log_stream_name() :: binary().
+-type sequence_token() :: binary().
+-type log_event() :: #{message => binary(), timestamp => integer()}.
+
+-type log_events() :: [log_event()].
+-type log_put_result() :: #{nextSequenceToken => binary(),
+                            rejectedLogEventsInfo => map()}.
+
 
 %% Library initialization
 -export([
@@ -49,7 +58,9 @@
     describe_log_groups/2,
     describe_log_groups/3,
     describe_log_groups/4,
-    create_log_stream/3
+    create_log_stream/3,
+    put_log_events/4,
+    put_log_events/5
 ]).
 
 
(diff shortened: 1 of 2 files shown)


rvangraan:
How would you suggest I deal with errors?
```
{error,{http_error,400,"Bad Request",
                   <<"{\"__type\":\"InvalidSequenceTokenException\",\"expectedSequenceToken\":\"49574602524945087512519741869202665284478508564977029746\",\"message\":\"The given sequenceToken is invalid. The next expected sequenceToken is: 49574602524945087512519741869202665284478508564977029746\"}">>}}
```

rvangraan:
Also what do you want me to do about the tests failing on R16 which has no maps?

motobob:
@rvangraan as for 400 - just return it to the application.
as for maps - i'm not sure it's actually necessary to implement this PR. so i would suggest simplifing the PR.
R16 so far is needed as backward compatibility. I was planning to deprecate it but have not yet came to it.

motobob:
@nalundgaard what do you think?

nalundgaard:
@motobob sorry for the delay, I agree. I think this should continue to use `jsx` proplists rather than maps. Converting to maps (or introducing them to the library) is an effort that needs to be undertaken in the future once R16 is fully deprecated, which we really ought to do soon.

motobob:
seems this will now see a green light as R16 will be gone from this repo next week

motobob:
see #511 which supersedes this one. please rebase add more if necessary.

PR #412, opened 2017-06-30
Add support for S3 inventory

KirillDavydov:


src/erlcloud_s3.erl
```diff
@@ -1227,6 +1232,228 @@ set_bucket_attribute(BucketName, AttributeName, Value, Config)
...
+    make_list_inventory_result(
+        Truncated,
+        ContinuationToken,
+        NextContinuationToken,
+        UpdatedResponse
+    ).
+
+make_list_inventory_result(false, _, _, Result)->
+    {ok, Result};
+make_list_inventory_result(_, ContinuationToken, [], Result)->
```
motobob: nope. see http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketListInventoryConfigs.html
only two clauses are here

src/erlcloud_s3.erl
```diff
@@ -1227,6 +1232,228 @@ set_bucket_attribute(BucketName, AttributeName, Value, Config)
...
+
+    InventoryId = proplists:get_value(id, Inventory),
+    XmlInventory = encode_inventory(Inventory),
+    put_bucket_inventory(BucketName, InventoryId, list_to_binary(XmlInventory), Config).
+
+-spec put_bucket_inventory(string(), string(), binary(), aws_config())
+        -> ok | {error, Reason::term()}.
+put_bucket_inventory(BucketName, InventoryId, XmlInventory, #aws_config{} = Config)
+    when is_list(BucketName), is_list(InventoryId), is_binary(XmlInventory) ->
+    Md5 = base64:encode(crypto:hash(md5, XmlInventory)),
```
motobob: is it really needed i do not see it in docs. http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUTInventoryConfig.html

isn't SigV4 doing it all for you?

src/erlcloud_s3.erl
```diff
@@ -1227,6 +1232,228 @@ set_bucket_attribute(BucketName, AttributeName, Value, Config)
...
+            {inv_key_to_name(Key), inv_encode_subtype(Value)}
+        end,
+        ListOfValues
+    );
+inv_encode_subtype({Key, Value}) ->
+    {inv_key_to_name(Key), inv_encode_subtype(Value)};
+inv_encode_subtype(Value) ->
+    [erlcloud_util:to_string(Value)].
+
+inv_key_to_name(inventory_configuration) ->  'InventoryConfiguration';
```
motobob: did you mean `''` or `""`? if former its unclear why we need atom2atom?

KirillDavydov: `' '`
atoms like `inventory_configuration` are more usable in a code while `'InventoryConfiguration'` are required by xml parser to build xml

src/erlcloud_s3.erl
KirillDavydov: just copied that from bucket lifecycle

PR #413, opened 2017-07-05
Support AT_TIMESTAMP iterator type for kinesis

arthurnovak:
This fix implements support for new shard iterator type in kinesis
http://docs.aws.amazon.com/kinesis/latest/APIReference/API_GetShardIterator.html

Please review @motobob @nalundgaard


danilagamma:
:+1:

src/erlcloud_kinesis.erl
```diff
@@ -267,27 +267,45 @@ describe_stream(StreamName, Limit, ExcludeShard, Config)
...
-
--spec get_shard_iterator(string(), string(), string(), string(), aws_config()) -> proplist().
-
-get_shard_iterator(StreamName, ShardId, ShardIteratorType, StartingSequenceNumber, Config) when is_record(Config, aws_config) ->
-  Json = [{<<"StreamName">>, StreamName}, {<<"ShardId">>, ShardId}, {<<"ShardIteratorType">>, ShardIteratorType}, {<<"StartingSequenceNumber">>, StartingSequenceNumber}],
-  erlcloud_kinesis_impl:request(Config, "Kinesis_20131202.GetShardIterator", Json).
+    get_shard_iterator(StreamName, ShardId, ShardIteratorType, default_config()).
 
+-spec get_shard_iterator(binary(), binary(), binary(),
+                         string() | aws_config()) ->
```
nalundgaard: Is that right? `string()`? should this be `binary() | integer() | float() | aws_config()`?

nalundgaard:
What's with this module? why are all the text input params spec'ed as `string()`? That seems like it's just... wrong.

nalundgaard:
Other than my question about the spec, this looks good to me.

arthurnovak:
@nalundgaard have no idea why specs are screwed here (shrug). I can fix that in a separate pr after

PR #420, opened 2017-07-15
Get credentials from ECS Task Role if available

### LINKED ISSUE #406: Support for ECS task specific IAM credentials in erlcloud_aws.autoconfig()
In July 2016, AWS accounced IAM Roles for ECS tasks:

https://aws.amazon.com/about-aws/whats-new/2016/07/amazon-ec2-container-service-now-supports-iam-roles-for-tasks/

This prevents that the underlying ECS Cluster Instances need all the possible IAM permissions for the different ECS services it runs: each ECS task can be linked to it's own permissions and secrets.

As a result, fetching the credentials from the EC2 instance metadata doesn't suffice anymore. You also have a container specific URL now where you can fetch the task specific IAM credentials from:

http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html

The chain to follow in fetching credentials is:
* env variables
* AWS profile
* ECS credentials only if the ECS Relative URI is set
* EC2 Instance Metadata

As an example, here is the implementation of this in the AWS SDK for Javascript:
https://github.com/aws/aws-sdk-js/blob/515fcbcbcc4b01a05ace7b98c1284387f70cb541/lib/node_loader.js#L36
https://github.com/aws/aws-sdk-js/blob/master/lib/credentials/ecs_credentials.js

keichan34:
Fixes #406 

I've verified this using ECS tasks in our ECS setup, but I would really like to have tests for this. Unfortunately, I couldn't find the tests for the instance profile metadata, so I wasn't sure where to start.

### DIFF
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -476,7 +476,16 @@ update_config(#aws_config{access_key_id = KeyId} = Config)
     {ok, Config};
 update_config(#aws_config{} = Config) ->
     %% AccessKey is not set. Try to read from role metadata.
-    case get_metadata_credentials(instance_metadata, Config) of
+    case get_metadata_credentials(task_credentials, Config) of
+        {error, _Reason} ->
+            case get_metadata_credentials(instance_metadata, Config) of
+                {error, _Reason} = Error -> Error;
+                {ok, Credentials} ->
+                    {ok, Config#aws_config {
+                           access_key_id = Credentials#metadata_credentials.access_key_id,
+                           secret_access_key = Credentials#metadata_credentials.secret_access_key,
+                           security_token = Credentials#metadata_credentials.security_token}}
+            end;
         {ok, Credentials} ->
             {ok, Config#aws_config {
                    access_key_id = Credentials#metadata_credentials.access_key_id,
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -414,19 +414,13 @@ auto_config_profile( ProfileOptions ) ->
     end.
 
 auto_config_task_metadata() ->
-    case erlcloud_ecs_container_credentials:task_credentials_available() of
-        ok -> perform_config_task_metadata();
-        error -> auto_config_metadata()
-    end.
-
-perform_config_task_metadata() ->
-    case config_task_metadata() of
+    case config_metadata(fun get_task_metadata_credentials/1) of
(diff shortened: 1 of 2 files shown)


src/erlcloud_aws.erl
```diff
@@ -664,6 +690,21 @@ get_metadata_credentials(Config) ->
             get_credentials_from_metadata(Config)
     end.
 
+-spec get_task_metadata_credentials(aws_config()) -> {ok, #metadata_credentials{}} | {error, term()}.
+get_task_metadata_credentials(Config) ->
+    %% See if we have cached credentials
+    case application:get_env(erlcloud, task_metadata_credentials) of
```
motobob: not sure why we need this. it's either in ECS or in EC2 so reusing `metadata_credentials` should be enough

src/erlcloud_aws.erl
```diff
@@ -447,6 +459,20 @@ config_metadata() ->
         {error, _Reason} = Error -> Error
     end.
 
+config_task_metadata() ->
```
motobob: let's not copy-paste these two functions but pass the `get_*` fun instead

src/erlcloud_aws.erl
```diff
@@ -410,6 +410,18 @@ auto_config_profile( ProfileOptions ) ->
     Profile = proplists:get_value( profile, ProfileOptions, default ),
     case profile( Profile, ProfileOptions ) of
         {ok, _Config} = Result -> Result;
+        {error, _} -> auto_config_task_metadata()
+    end.
+
+auto_config_task_metadata() ->
+    case erlcloud_ecs_container_credentials:task_credentials_available() of
```
motobob: not sure why we need this wrapper and not just call `config_task_metadata()` right away.

src/erlcloud_aws.erl
```diff
@@ -664,6 +669,21 @@ get_metadata_credentials(Config) ->
             get_credentials_from_metadata(Config)
     end.
 
+-spec get_task_metadata_credentials(aws_config()) -> {ok, #metadata_credentials{}} | {error, term()}.
+get_task_metadata_credentials(Config) ->
```
motobob: one last c/p here to remove and good to go
`get_task_metadata_credentials/1` is not that different for `get_metadata_credentials/1`.
it only differs in calling fun either `get_credentials_from_task_metadata/1` or `get_credentials_from_metadata/1`

PR #422, opened 2017-08-09
increase max RoleSessionName length to 64 for sts AssumeRole api

knyastasia:
max RoleSessionName length is now 64 (was 32) 
see [assumeRole api](http://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html)


motobob:
thank you

PR #423, opened 2017-08-09
add ability to pass in the awsRegion when executing aws_request4

knyastasia:
[The assumption](https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_aws.erl#L139) that the second element in the host is an aws region is not always true.

### DIFF
--- a/include/erlcloud_aws.hrl
+++ b/include/erlcloud_aws.hrl
@@ -118,8 +118,9 @@
           %% Currently matches DynamoDB retry
           %% It's likely this is too many retries for other services
           retry_num=10::non_neg_integer(),
-          assume_role = #aws_assume_role{} :: aws_assume_role() %% If a role to be assumed is given
+          assume_role = #aws_assume_role{} :: aws_assume_role(), %% If a role to be assumed is given
           %% then we will try to assume the role during the update_config
+          region=undefined::string()|undefined
          }).
 -type(aws_config() :: #aws_config{}).
 
@@ -146,5 +147,3 @@
           should_retry :: boolean() | undefined
         }).
 -type(aws_request() :: #aws_request{}).
-
-
--- a/include/erlcloud_aws.hrl
+++ b/include/erlcloud_aws.hrl
@@ -120,7 +120,7 @@
           retry_num=10::non_neg_integer(),
           assume_role = #aws_assume_role{} :: aws_assume_role(), %% If a role to be assumed is given
           %% then we will try to assume the role during the update_config
-          region=undefined::string()|undefined
+          aws_region=undefined::string()|undefined
          }).
 -type(aws_config() :: #aws_config{}).
 
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -162,7 +162,7 @@ aws_request4(Method, Protocol, Host, Port, Path, Params, Service, Headers, Confi
     end.
 
 aws_request4_no_update(Method, Protocol, Host, Port, Path, Params, Service,
-                       Headers, #aws_config{region = AwsRegion} = Config) ->
+                       Headers, #aws_config{aws_region = AwsRegion} = Config) ->
     Query = erlcloud_http:make_query_string(Params),
     Region = case AwsRegion of
       undefined -> aws_region_from_host(Host);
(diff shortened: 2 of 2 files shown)


knyastasia:
@motobob you can set up [a custom domain name](http://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-custom-domains.html) for the API gateway.
if you agree, that the idea itself is valid, I'll move the implementation into `aws_region_from_host` method

knyastasia:
@motobob I'm also planning to make another PR which will allow to use `Content-Type: application/json` (the most popular one in this case). so maybe it make sense to create a separate function for making the requests to the API gateways?

motobob:
ok, so your use case is that you have an API in front of some of AWS APIs, probably for having sig4 everywhere and better audit and you want applications using erlcloud to talk to such endpoints natively, right? 
or you really have some custom API under the gateway? 

in both cases i think making `erlcloud_aws` work with both is the right direction

motobob:
I would suggest updating the `aws_region_from_host` to try the host url assumption first and then fallback to `application:get_env(aws_region)` as majority should NOT be cross-region api calls but still possible. for that leave /10 arrity fun.

src/erlcloud_aws.erl
```diff
@@ -153,18 +153,24 @@ aws_request4(Method, Protocol, Host, Port, Path, Params, Service, Config) ->
...
         {error, Reason} ->
             {error, Reason}
     end.
 
 aws_request4_no_update(Method, Protocol, Host, Port, Path, Params, Service,
-                       Headers, #aws_config{} = Config) ->
+                       Headers, #aws_config{} = Config, AwsRegion) ->
     Query = erlcloud_http:make_query_string(Params),
-    Region = aws_region_from_host(Host),
+    Region = case AwsRegion of
```
motobob: move this case into /9 fun instead of `undefined`

knyastasia:
first case, right. 
I'll try to make all suggested changes later today. thanks!

knyastasia:
I've been reading the aws documentation, and found next statement:
> Some services, such as IAM, do not support regions; therefore, their endpoints do not include a region. Some services, such as Amazon EC2, let you specify an endpoint that does not include a specific region, for example, https://ec2.amazonaws.com. In that case, AWS routes the endpoint to us-east-1.
I also took a look into some official aws SDKs and they're getting region either from asw config objects or from env vars.
I suggest to enrich `aws_config` with region and get it either from config record or from `application:get_env(aws_region)`. and remove host parsing completely (altogether with passing region to the function params). I understand it may break backward compatibility, so the `aws_region_from_host` have to stay, but you may deprecate it in future if you want.
@motobob what do you think?

motobob:
the library already works with region agnostic endpoints so 
>and remove host parsing completely

is not an option. even if aws_config region variable is added, proper url formatting is still required in 99%+ cases.

knyastasia:
can you pls point me to the region agnostic logic for the aws sigv4 requests?

knyastasia:
@motobob 
I've added region to the `aws_config`, default value is `undefined`
I've also removed the 10 args function that I've added before.
please take a look

include/erlcloud_aws.hrl
```diff
@@ -118,8 +118,9 @@
           %% Currently matches DynamoDB retry
           %% It's likely this is too many retries for other services
           retry_num=10::non_neg_integer(),
-          assume_role = #aws_assume_role{} :: aws_assume_role() %% If a role to be assumed is given
+          assume_role = #aws_assume_role{} :: aws_assume_role(), %% If a role to be assumed is given
           %% then we will try to assume the role during the update_config
+          region=undefined::string()|undefined
```
motobob: s/region/aws_region please to follow rest of the naming

src/erlcloud_aws.erl
```diff
@@ -146,7 +146,7 @@ aws_region_from_host(Host) ->
         [_, Value, _, _ | _Rest] ->
             Value;
         _ ->
-            "us-east-1"
+            application:get_env(erlcloud, aws_region, "us-east-1")
```
motobob: stylish: default it in record define so to have only a single place for deafults

knyastasia: do you mean `aws_config` record? I can't do that, it will break everyone, who's been relying on `aws_region_from_host`.
or do you mean I need to create a constant in the module?

include/erlcloud_aws.hrl
knyastasia: done

PR #424, opened 2017-08-09
uri encode path before sign_v4

knyastasia:
if path contains any special characters, it needs to be uri encoded before calculating the signature, otherwise the signature will do not match with the one calculated by aws.

NOTE: we can't pass the uri encoded path to the `aws_request4` method, because it will be uri encoded second time (while making the actual request). 
example: 
instead of `:` -> `%3`, we'll get `:` -> `%3` -> `%253`

### DIFF
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -165,12 +165,13 @@ aws_request4_no_update(Method, Protocol, Host, Port, Path, Params, Service,
                        Headers, #aws_config{} = Config) ->
     Query = erlcloud_http:make_query_string(Params),
     Region = aws_region_from_host(Host),
+    Uri = erlcloud_http:url_encode_loose(Path),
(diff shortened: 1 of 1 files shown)


src/erlcloud_aws.erl
```diff
@@ -165,12 +165,13 @@ aws_request4_no_update(Method, Protocol, Host, Port, Path, Params, Service,
                        Headers, #aws_config{} = Config) ->
     Query = erlcloud_http:make_query_string(Params),
     Region = aws_region_from_host(Host),
+    Uri = erlcloud_http:url_encode_loose(Path),
```
motobob: one more question - which service api caused issues for you? 
in S3 the request is formed/signed other way and it works there.

knyastasia: service `execute-api` which is api gateway

Issue #397, opened 2017-03-16
Dialyzer does not pass in Erlang 19


ssepulvedam:
Dialyzer is failing (at least) under Erlang 19 (_Erlang/OTP 19 [erts-8.1]_), which breaks it for the applications using  `erlcloud`. 

I'm not including the full output as it is quite big, just a few samples from `erlcloud_ddb2`.

```erlang
erlcloud_ddb2.erl:758: Function undynamize_folder/4 will never be called
erlcloud_ddb2.erl:769: Function undynamize_record/3 will never be called
erlcloud_ddb2.erl:773: The created fun has no local return
erlcloud_ddb2.erl:979: The pattern {'simple', Record} can never match the type {'error','metadata_not_available' | {'socket_error',_} | {'http_error',_,_,_}} | {'ok','false' | 'null' | 'true' | binary() | [any()] | number() | map()}
erlcloud_ddb2.erl:999: Function consumed_capacity_record/0 has no local return
erlcloud_ddb2.erl:1000: Record construction #ddb2_consumed_capacity{capacity_units::'undefined',global_secondary_indexes::'undefined',local_secondary_indexes::'undefined',table::'undefined',table_name::'undefined'} violates the declared type of field capacity_units::number() and global_secondary_indexes::[{binary(),number()}] and local_secondary_indexes::[{binary(),number()}] and table::number() and table_name::binary()
erlcloud_ddb2.erl:1017: Function undynamize_consumed_capacity/2 has no local return
erlcloud_ddb2.erl:1024: Function item_collection_metrics_record/0 has no local return
erlcloud_ddb2.erl:1025: Record construction #ddb2_item_collection_metrics{size_estimate_range_gb::'undefined'} violates the declared type of field size_estimate_range_gb::{number(),number()}
erlcloud_ddb2.erl:1034: Function undynamize_item_collection_metrics/2 has no local return
erlcloud_ddb2.erl:1057: Function global_secondary_index_description_record/0 has no local return
[...]

PR #426, opened 2017-08-15
[sqs-batch-entry-spec-error]: Fixup batch_entry() type.

jkrukoff:
It appears that the batch_entry() type has been doubly wrapped in a list
type, as all usages of the type are also wrapped in a list. From the
naming, I assume the intent was for the type to be singular.

Found while attempting to type check usage of
erlcloud_sqs:delete_message_batch, but problem appears to be general to
all usages of erlcloud_sqs:batch_entry().


jkrukoff:
By no means an attempt at fixing #397, but is one small part of the problem.

PR #427, opened 2017-08-18
add erlang 20 for test

getong:


nalundgaard:
It appears that, [`require_otp_vsn`](https://github.com/erlcloud/erlcloud/blob/54288931b62a21a1bedfae2eb277abfb2264d325/rebar.config#L2) would need to include 20 for this to work.

Issue #414, opened 2017-07-10
Support for SQS Message Attributes inside send_message


arcusfelis:
Message attributes can be received, but not send.
Docs
http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-attributes.html

motobob:
thanks for contribution and fixing this API gap.

PR #429, opened 2017-08-21
Add support for sending and receiving message attributes in SQS; Issue #414

ctbarbour:
Based on the [SendMessage](http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html) and [ReceiveMessage](http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html) documentation we have added support for including MessageAttributes. Currently, we have added support for Binary, String, and Number (e.g. integer or float). The specification includes support for custom types which we have chosen to exclude in this implementation for the time being. We have also included the ability to optionally receive message attributes. The API works similar to the way the existing attributes API works. A caller can specify, 'none', 'all' or a list of attributes to include in the response.

We have also included support for MessageAttributes in SendMessageBatch.
```erlang
Batch = [{"test-msg-01", "Hello, World!", [{"content-type", "text/plain"}]},
         {"test-msg-02", "More Cowbell", [{"correlationId", "1234"}]}],
erlcloud_sqs:send_message_batch("MyQueue", Batch).
```

### DIFF
--- a/src/erlcloud_sqs.erl
+++ b/src/erlcloud_sqs.erl
@@ -363,7 +363,12 @@ decode_message_attribute(Attributes) ->
       erlcloud_xml:get_text("Value/StringValue", Attr)} || Attr <- Attributes].
 
 decode_message_attribute_value("Number", Value) ->
-    list_to_integer(Value);
+    case lists:member($., Value) of
+        true ->
+            list_to_float(Value);
+        false ->
+            list_to_integer(Value)
+    end;
 decode_message_attribute_value("Binary", Value) ->
     list_to_binary(Value);
 decode_message_attribute_value("String", Value) ->
--- a/src/erlcloud_sqs.erl
+++ b/src/erlcloud_sqs.erl
@@ -45,7 +45,7 @@
                                     created_timestamp | last_modified_timestamp | policy |
                                     queue_arn).
 
--type(batch_entry() :: {string(), string()}).
+-type(batch_entry() :: {string(), string()} | {string(), string(), [message_attribute()]}).
 -type(message_attribute() :: {string(), string() | integer() | float() | binary()}).
 
 -spec new(string(), string()) -> aws_config().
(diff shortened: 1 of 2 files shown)


src/erlcloud_sqs.erl
```diff
@@ -550,3 +587,41 @@ queue_path([$/|QueueName]) -> [$/ |erlcloud_http:url_encode(QueueName)];
...
+    erlcloud_aws:param_list(lists:reverse(lists:foldl(Map, [], Attributes)), "MessageAttribute").
+
+encode_message_attribute({Key, Value}) ->
+    [
+      {message_attribute_title(datatype), encode_message_attribute_type(Value)},
+      {message_attribute_title(string_value), encode_message_attribute_value(Value)},
+      {message_attribute_title(name), Key}
+    ].
+
+message_attribute_title(datatype) ->
```
motobob: this fun seem unnecessary and i would just put the strings right above.

src/erlcloud_sqs.erl
```diff
@@ -332,11 +347,28 @@ decode_message(Message) ->
...
+decode_message_attributes(Attributes) ->
+    [{decode_msg_attribute_name(Name), decode_message_attribute_value(DataType, StringValue)} ||
+        {Name, DataType, StringValue} <- decode_message_attribute(Attributes)].
+
+decode_message_attribute(Attributes) ->
+    [{erlcloud_xml:get_text("Name", Attr),
+      erlcloud_xml:get_text("Value/DataType", Attr),
+      erlcloud_xml:get_text("Value/StringValue", Attr)} || Attr <- Attributes].
+
+decode_message_attribute_value("Number", Value) ->
```
ctbarbour: There is an issue here with determining if the Number type defined in the response from SQS is a float or an integer. Here, we naively assume Number will always be an integer. Not sure the best way to support coercion to float or integer.

ctbarbour: Since [this logic](https://github.com/erlcloud/erlcloud/blob/81243b191550256c6452e35f7455d89b15759bac/src/erlcloud_ddb_streams.erl#L196) is already present elsewhere in the project to determine whether a string value represents a float or an integer I'll go ahead and use the same logic here.

PR #434, opened 2017-09-04
Custom datatypes in message attributes

ctbarbour:
Here we have added support for custom message attribute data types for
SQS messages based on the documentation found
http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-attributes.html#message-attributes-data-types-validation.

We have also encoded Float and Integer types using custom data types
since the SQS API only supports a generic Number type. When we receive
a message with a Number attribute we don't know whether it is expected
to be represented as a float or an integer. Previously, we would parse
the data attribute string value for the presence of a `$.`.

### DIFF
--- a/src/erlcloud_sqs.erl
+++ b/src/erlcloud_sqs.erl
@@ -361,11 +361,11 @@ decode_message_attribute(Attributes) ->
     F = fun(Attr) ->
                 Name = erlcloud_xml:get_text("Name", Attr),
                 DataType = erlcloud_xml:get_text("Value/DataType", Attr),
-                Value = case string:find(DataType, "Binary") of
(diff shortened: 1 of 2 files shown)


ctbarbour:
I'm not sure about [using custom datatypes](https://github.com/erlcloud/erlcloud/pull/434/files#diff-48b0d470a823908f6a04e039d6e949c3R643) to explicitly distinguish a Number values as a Float or Integer. We can revert to the string parsing behaviour if we think that's better. With the current implementation there could be conflicts with other SQS clients.

motobob:
Thank you for contribution.

PR #436, opened 2017-09-05
Fix some faulty specs

rolkar:
In order to be able to run dialyzer on my code I had to fix some specs in erlcloud.
This is by far not all specs that needs fixing, but rather the ones I need.


motobob:
thank you .

PR #437, opened 2017-09-12
allow option to return all headers from put_object

lazedo:
we have the need to collect other metadata


lazedo:
@motobob you mean something like this ?
```
 [{version_id, proplists:get_value("x-amz-version-id", Headers, "null")} | Headers].
```
`version_id` is to keep compatibility

motobob:
yes, that works for me

PR #438, opened 2017-09-12
export s3_request

lazedo:
allow direct requests


motobob:
I would suggest to add missing APIs instead. 
secondly, exposing old SigV2 API is no-go. Use of sigV4 only is recommended.

PR #443, opened 2017-09-26
Implement Athena support

arthurnovak:
Please review @motobob @nalundgaard @danilagamma


src/erlcloud_athena.erl
```diff
@@ -0,0 +1,436 @@
...
+batch_get_query_execution(QueryExecutionIds, Config) ->
+    Request = #{<<"QueryExecutionIds">> => QueryExecutionIds},
+    request(Config, "BatchGetQueryExecution", Request).
+
+%% @doc
+%% Athena API:
+%% http://docs.aws.amazon.com/athena/latest/APIReference/API_CreateNamedQuery.html
+%%
+%% `
+%% erlcloud_kinesis:create_named_query(
```
nalundgaard: s/kinesis/athena?

src/erlcloud_athena.erl
```diff
@@ -0,0 +1,436 @@
...
+get_query_execution(QueryExecutionId, Config) ->
+    Request = #{<<"QueryExecutionId">> => QueryExecutionId},
+    request(Config, "GetQueryExecution", Request).
+
+%% @doc
+%% Athena API:
+%% http://docs.aws.amazon.com/athena/latest/APIReference/API_GetQueryResults.html
+%%
+%% `
+%% erlcloud_kinesis:get_query_results(<<"some-query-uuid">>,
```
nalundgaard: s/kinesis/athena?

nalundgaard:
This doesn't compile on R16. I don't think we have officially retired that in the mainline. Would be nice to make maps optional and only supported on 17/18+.

arthurnovak:
@nalundgaard Do we really want to support R16 yet? 20 has been already released, maybe it's time to get rid of old things to avoid support them.

nalundgaard:
I don't disagree, but it should be formally retired, not inadvertently retired with the introduction of a new module that silently breaks compatibility with no prior warning.

src/erlcloud_athena.erl
```diff
@@ -0,0 +1,436 @@
...
+               {"content-type", "application/x-amz-json-1.1"}],
+    Region = erlcloud_aws:aws_region_from_host(Host),
+    erlcloud_aws:sign_v4_headers(Config, Headers, Body, Region, "athena").
+
+get_url(#aws_config{athena_scheme = Scheme,
+                    athena_host   = Host,
+                    athena_port   = Port}) ->
+    Scheme ++ Host ++ ":" ++ integer_to_list(Port).
+
+get_retry_fun() ->
```
motobob: what is the point of this?

src/erlcloud_athena.erl
```diff
@@ -0,0 +1,436 @@
...
+-spec start_query_execution(binary(), binary(), binary(), binary(),
+                            binary() | undefined,
+                            binary() | undefined,
+                            aws_config()) ->
+    {ok, binary()} | {error, any()}.
+start_query_execution(ClientReqToken, Db, Query, OutputLocation,
+                      EncryptionOption, KmsKey, Config) ->
+    EncryptConfig = get_encrypt_config(EncryptionOption, KmsKey),
+    OutputConfig  = #{<<"OutputLocation">> => OutputLocation},
+    ResultConfig  = maps:merge(OutputConfig, EncryptConfig),
```
motobob: stylish
why not `EncryptConfig#{<<"OutputLocation">> => OutputLocation}`
and same for all other maps:merge()

arthurnovak: ah gosh, i thought i got rid of this 'merges' everywhere. Just missed those

PR #444, opened 2017-09-26
Announce R16 deprecation + roadmap update

arthurnovak:
Please take a look @motobob @nalundgaard

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,31 +46,26 @@ The libraries can be used two ways:
 
 Below is the library roadmap update along with regular features and fixes.
 
-- 0.13.10
- * pre Alert Logic Fork merge
+- 3.0.X __R16 deprecation is scheduled on October 9, 2017__
+  - Remove R16 support
+  - Support maps
 
-- 2.0.0
- Existing code
- merge of [Alert Logic](https://github.com/alertlogic/erlcloud/tree/v1.2.4) fork into upstream.
- This is a major version bump which contains lots of new features and functionality.
- Unfortunately, it also contains quite a number of low level APIs incompatibilities since the fork diverged for a long while.
- Making it backward compatible does not seem feasible and valuable at the moment.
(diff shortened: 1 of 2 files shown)


motobob:
👍 but just thinking about the level of comm required apart from just pushing an new 2.3.0..

nalundgaard:
Also, I would propose that we tie this to wider maps support as a 3.0 update, this is not at all a 2.x change by traditional definitions.

motobob:
i agree that canonically it's 3.0.0.
but still just merge  ~week, no?

nalundgaard:
I would propose to deprecate it in the README now, with a notification of when we will remove official support, and then take it out of travis separately at that communicated time. I'd also propose simply gutting `require_otp_vsn`, personally. Seems less fragile to simply document which Erlangs we officially support in the README or something, rather than keep a regex around whose main function is to just prevent the client from working on future Erlangs until there is a PR. When we introduce real maps support, without anything to make it backwards compatible, the library will simply no longer compile on R16 and lower, so why bother?

README.md
```diff
@@ -47,31 +47,41 @@ The libraries can be used two ways:
...
+  This is a major version bump which contains lots of new features and functionality.
+  Unfortunately, it also contains quite a number of low level APIs incompatibilities since the fork diverged for a long while.
+  Making it backward compatible does not seem feasible and valuable at the moment.
+  - No APIs have been removed and it's on branched of 0.13.10 at the moment. Any minor version delta added during notice time will be compensated before the merge.
+  - intentionally jumping to 2.0.0 as AL fork has v1.X.X
 
 - 2.1.X
- * fix dialyzer findings and make it mandatory for the library
- * ~~make full support of Mix/HEX~~ done
+  - fix dialyzer findings and make it mandatory for the library
```
nalundgaard: If this isn't done in 2.1.X, should we move it to another release?

PR #446, opened 2017-10-03
Initial support of Glue catalog table and Jobs APIs

arthurnovak:
Please review @motobob @nalundgaard


src/erlcloud_glue.erl
```diff
@@ -0,0 +1,551 @@
...
+               {"content-type", "application/x-amz-json-1.1"}],
+    Region = erlcloud_aws:aws_region_from_host(Host),
+    erlcloud_aws:sign_v4_headers(Config, Headers, Body, Region, "glue").
+
+get_url(#aws_config{glue_scheme = Scheme,
+                    glue_host   = Host,
+                    glue_port   = Port}) ->
+    Scheme ++ Host ++ ":" ++ integer_to_list(Port).
+
+get_retry_fun() ->
```
motobob: why not use the fun directly ?

src/erlcloud_glue.erl
```diff
@@ -0,0 +1,548 @@
...
+    Request1 = update_catalog_id(CatalogId, Request0),
+    case request(Config, "CreateTable", Request1) of
+        {ok, _} -> ok;
+        Error   -> Error
+    end.
+
+%% @doc
+%% Glue API:
+%% https://docs.aws.amazon.com/glue/latest/webapi/API_DeleteJob.html
+%%
```
nalundgaard: Any reason for not including a sample invocation here?

arthurnovak: i don't think it makes sense to document any obvious fun arguments and responses everywhere. We have spec for that, which looks descriptive in current case: jobname binary as input, ok as output:
```erlang
-spec delete_job(binary()) -> ok | {error, any()}.
delete_job(JobName) ->
    delete_job(JobName, default_config()).
```
The main problem with docs: they are forgotten to update very often
In other cases, where we have some not obvious arguments like `TableInput` as map, I've added docs

nalundgaard: 👍

PR #447, opened 2017-10-04
Add environment-level configurability for ec2_metadata_url

sargun:
In some environments, for example, when you're doing K8s on AWS,
you may want to run the likes of a metadata proxy. In this case,
the default EC2 metadata URI needs to be changed. To do this
as transparently as possible, we pass this as an environment variable.

This introduces a general configuration mechanism for passing
erlang application-specific environment variables as OS environment.


sargun:
This relies on R16 deprecation. New features don't work on <20. LMK if you need me to backport.

sargun:
I'd be fine with using the 18/19 feature set. It's just a gate around the new strings functions on 20. 

I have a strong preference to introduce this generic mechanism here as opposed to piecemeal in each library. 

The next step I would like to take is introdue a state machine that takes IAM credentials from IAM service and automatically loads them in when on EC2. Gen statem is only 19 and above. Is that a reasonable gate for a new feature?

PR #448, opened 2017-10-04
respect content type header provided by user in aws_request4

knyastasia:


src/erlcloud_aws.erl
```diff
@@ -293,6 +290,15 @@ format_timestamp({{Yr, Mo, Da}, {H, M, S}}) ->
       io_lib:format("~4.10.0b-~2.10.0b-~2.10.0bT~2.10.0b:~2.10.0b:~2.10.0bZ",
                     [Yr, Mo, Da, H, M, S])).
 
+encode_params(Params, Headers) ->
+  LowerCaseHeaders = lists:map(fun({K, V}) -> {string:to_lower(K), V} end, Headers),
+  case proplists:get_value(<<"content-type">>, LowerCaseHeaders) of
```
motobob: string vs binary?  how this can ever match?

src/erlcloud_aws.erl
```diff
@@ -293,6 +290,15 @@ format_timestamp({{Yr, Mo, Da}, {H, M, S}}) ->
...
+    undefined -> {erlcloud_http:make_query_string(Params),
+                  [{"content-type", ?DEFAULT_CONTENT_TYPE} | LowerCaseHeaders]};
+    % when `content-type` is specified, this is developer's responsibility to encode params properly
```
motobob: seems to be part of readme or at least top level `aws_request` fun

src/erlcloud_aws.erl
knyastasia: wow. not sure what I was thinking about. it should be just double quotes, right?

motobob: yes.  the line just above does `string:to_lower(K)` and that is all strings

knyastasia: sorry, it happened because in elixir string and binary are the same. completely forgot, that erlang strings are just lists of chars. Fixed

src/erlcloud_aws.erl
knyastasia: moved it on the `aws_request4/9` level, because this is the only place where user can provide headers (including content type) and params that need to be encoded.
I've also added the content type check to the `aws_request_form` (previously it was the place where headers were provisioned with default content type) because this function is also exposed

knyastasia:
@motobob please take a look at this PR again. I've made all requested fixes. thanks!

src/erlcloud_aws.erl
```diff
@@ -187,16 +189,17 @@ aws_request4_no_update(Method, Protocol, Host, Port, Path, Params, Service,
...
 
 
 -spec aws_request_form(Method :: atom(), Protocol :: undefined | string(), Host :: string(),
                         Port :: undefined | integer() | string(), Path :: string(), Form :: string(),
                         Headers :: list(), Config :: aws_config()) -> {ok, binary()} | {error, tuple()}.
 aws_request_form(Method, Protocol, Host, Port, Path, Form, Headers, Config) ->
-    RequestHeaders = [{"content-type",
-                      "application/x-www-form-urlencoded; charset=utf-8"} |
-                     Headers],
+    RequestHeaders = case proplists:is_defined("content-type", Headers) of
```
motobob: last question for this PR
seems this case is unnecessary given the cahgnes at line 176 above where the `"content-type"` is always added.
agree?

knyastasia: `aws_request_form` method is also used [here](https://github.com/erlcloud/erlcloud/blob/81243b191550256c6452e35f7455d89b15759bac/src/erlcloud_ses.erl#L989)

motobob:
thank you and sorry for delays...was out.
hope this fine grained tuning help your use case.

knyastasia:
it helps a lot, thanks!

PR #453, opened 2017-10-26
Fix spelling in SNS plugin

sebastiw:


motobob:
thank you for the fix.
in spite one may treat this as API braking change it's a nice fix for this

PR #456, opened 2017-10-31
adding the possibility to omit shadow trails in describe_trails

kstarikov:


kstarikov:
@motobob please review

src/erlcloud_cloudtrail.erl
```diff
@@ -115,6 +115,14 @@ describe_trails(Trails, Config) ->
     Json = [{<<"TrailNameList">>, [list_to_binary(T) || T <- Trails]}],
     ct_request("DescribeTrails", Json, Config).
 
+-spec describe_trails([string()], aws_config(), boolean()) -> ct_return().
```
motobob: i would appreciate if we keep to the general api scheme where `Config` is the last parameter.
```erlang
spec describe_trails([string()], boolean(), aws_config()) -> ct_return().
```

src/erlcloud_cloudtrail.erl
```diff
@@ -115,6 +115,14 @@ describe_trails(Trails, Config) ->
...
+describe_trails([], Config, IncludeShadowTrails)
+        when IncludeShadowTrails =:= false ->
```
motobob: IMO this guard/function clause is unnecessary and simple guard for boolean is enough.
```erlang
describe_trails([], IncludeShadowTrails, Config) when is_boolean(IncludeShadowTrails) ->
    Json = [{<<"includeShadowTrails">>, IncludeShadowTrails}],
    ct_request("DescribeTrails", Json, Config).

File: src/erlcloud_cloudtrail.erl [hunk 3]
```diff
@@ -115,6 +115,14 @@ describe_trails(Trails, Config) ->
     Json = [{<<"TrailNameList">>, [list_to_binary(T) || T <- Trails]}],
     ct_request("DescribeTrails", Json, Config).
 
+-spec describe_trails([string()], aws_config(), boolean()) -> ct_return().
+describe_trails([], Config, IncludeShadowTrails)
```
motobob: lastly , first parameter is ignored here while it should not and this is a bug 
http://docs.aws.amazon.com/awscloudtrail/latest/APIReference/API_DescribeTrails.html

src/erlcloud_cloudtrail.erl
kstarikov: @motobob and what if IncludeShadowTrails is false?

src/erlcloud_cloudtrail.erl
kstarikov: It's not a bug. It's just limited API which is no worse than before. But I'll add the trail names support anyway.

kstarikov: Speaking about bugs, erlcloud wasn't actually working with the `trailNameList` parameter correctly, and I fixed it.

kstarikov:
@motobob all comments addressed.

Issue #452, opened 2017-10-24
hackney as the default http client? 


motobob:
what do we think about moving away from `lhttpC` to `hackney` as the default client?
this is already supported in the lib if rebar.config updated/configs changed

the only problem I can see are consumers which use lhttpc_pools...

motivation: 
https://github.com/erlcloud/lhttpc/commit/4067c7ea9718a181d1aa3fc522e749ce6cc1fed2 supposed to fix some issues but actually introduced issues to other cases.

@nalundgaard @kkuzmin

define-null:
@motobob why not just introduce the configuration variable, so that the user can specify which http library he want's to use? Without need to fetch extra dependencies and having hackney/lhttpc being part of the application section in app.src file?

motobob:
it's already there as part of https://github.com/erlcloud/erlcloud/blob/master/include/erlcloud_aws.hrl#L111 
but smb just need to start the desired client before.

define-null:
Yes, I have seen that. 
I meant lhttpc or hackney should not be part of https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud.app.src#L19

define-null:
And by configuration variables I meant application env.

define-null:
@motobob any thoughts on that?

nanshakov:
Hello. Hackney [support](https://github.com/benoitc/hackney#use-the-default-pool) default and custom pool, and a custom pool handler. [Default setting is max_connections = 50](https://github.com/benoitc/hackney/blob/master/src/hackney.app.src#L23). How do you image moving from `lhttpC` to `hackney` ? In `erlcloud_aws` we have [hardcode](https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_aws.erl#L44) return's types from lhttpc. How can i make the most correct moving ?

PR #457, opened 2017-11-01
Don't force lhttpc to be started when using Erlcloud.

psytale:
Party of the work towards #452 removes the forced dependency on lhttpc at least.


psytale:
Its required to be able to use just hackney without lhttpc, its a stepping  stone. What approach would you prefer? Surely if you ever want to support something in addition to hackney you'd need this?

PR #460, opened 2017-11-22
fix guard for erlcloud:send_message/5

lehoff:
The logic in the guard related to `DelaySeconds` seemed wrong. 
The `orelse` as lower precedence than `andalso` so the old guard was effectively an or consisting of two ands.

Also replaced `,` with `andalso` in the guard to make it more uniform.
Plus moved the check `is_record(Config, aws_config)` into the function clause as a pattern match as that is the more common way of doing it.


motobob:
nice catch.

PR #461, opened 2017-11-29
avoid that aws:auto_config/0 returns {ok, undefined}

lehoff:
If you call erlcloud_aws:auto_config/0 without the correct environment variables it
returns `{ok, undefined}`, which is a violation of its spec.

A deeper fix would be to fix the underlying problem in either auto_config_profile,
default_config_region or default_config_get.

This solution allows usage of the function in a fail-fast manner, which is good enough
for starters.

### DIFF
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -415,7 +415,12 @@ default_config_region_services() ->
 %% return <code>undefined</code>.
 %%
 auto_config() ->
-    auto_config( [] ).
+    case auto_config( [] ) of
+        {ok, #aws_config{}} = Res ->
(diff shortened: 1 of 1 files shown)


lehoff:
Not sure that I agree with that solution. I can appreciate that adding more comments to state that this it a temporary protection against an underlying problem.

Right now we have a top level function that does no live up to its spec, which is not at all good.
Postponing to protect the users until a fix for the underlying problem has been fixed does not seem like the best way forward.

motobob:
why cant we fix the underlining problem?

lehoff:
Underlying problem fixed in other PR #472

PR #466, opened 2017-12-15
GuardDuty support

kstarikov:


kstarikov:
@motobob please review

rebar.config.script
```diff
@@ -7,7 +7,8 @@ case erlang:function_exported(rebar3, main, 1) of
         %% profiles
         [{deps, [{meck, ".*",{git, "https://github.com/eproxus/meck.git", {tag, "0.8.4"}}},
                  {jsx, ".*", {git, "git://github.com/talentdeficit/jsx.git", {tag, "2.8.0"}}},
-                 %% {hackney, ".*", {git, "git://github.com/benoitc/hackney.git", {tag, "1.2.0"}}},
+                 {unicode_util_compat, ".*", {git, "https://github.com/benoitc/unicode_util_compat.git", {tag, "0.3.1"}}},
+                 {hackney, ".*", {git, "git://github.com/benoitc/hackney.git", {tag, "1.10.1"}}},
```
motobob: are we sure we still desperately need hackney for GD endpoints?

src/erlcloud_guardduty.erl
```diff
@@ -0,0 +1,144 @@
...
+    ok.
+
+%%
+%% API
+%%
+
+%%------------------------------------------------------------------------------
+%% @doc
+%% GuardDuty API:
+%% [http://docs.aws.amazon.com/guardduty/....html]
```
motobob: can we have actual link here.

src/erlcloud_guardduty.erl
```diff
@@ -0,0 +1,144 @@
...
+    Path = base_path() ++ "detector",
+    QParams = filter_undef([{"Marker", Marker},
+                            {"MaxItems", MaxItems}]),
+    guardduty_request(Config, get, Path, undefined, QParams).
+
+
+filter_undef(List) ->
+    lists:filter(fun({_Name, Value}) -> Value =/= undefined end, List).
+
+base_path() ->
```
motobob: is this really necessary now?

src/erlcloud_guardduty.erl
```diff
@@ -0,0 +1,144 @@
...
+                        Config       :: aws_config()) -> gd_return().
+describe_detector(DetectorId, Config) ->
+    Path = base_path() ++ "detector/" ++ binary_to_list(DetectorId),
+    guardduty_request(Config, get, Path, undefined).
+
+
+%%------------------------------------------------------------------------------
+%% @doc
+%% GuardDuty API:
+%% [http://docs.aws.amazon.com/guardduty/....html]
```
motobob: actual link too.

src/erlcloud_guardduty.erl
```diff
@@ -0,0 +1,144 @@
...
+            {error, Reason}
+    end.
+
+guardduty_request_no_update(Config, Method, Path, Body, QParam) ->
+    Form = case encode_body(Body) of
+               <<>>   -> erlcloud_http:make_query_string(QParam);
+               Value  -> Value
+           end,
+    Headers = headers(Method, Path, Config, encode_body(Body), QParam),
+    io:format("HEADERS: ~p~n", [Headers]),
```
motobob: remove

src/erlcloud_guardduty.erl
```diff
@@ -0,0 +1,144 @@
...
+    case erlcloud_aws:aws_request_form_raw(
+           Method, Config#aws_config.guardduty_scheme, Config#aws_config.guardduty_host,
+           Config#aws_config.guardduty_port, Path, Form, Headers, Config) of
+        {ok, Data} ->
+            {ok, jsx:decode(Data)};
```
motobob: to my memory jsx does not like `<<>>` and that deserves a separate wrapper/clause.

rebar.config.script
kstarikov: `lhttpc` doesn't play well with TLS1.2 used by AWS.

src/erlcloud_guardduty.erl
kstarikov: But AWS requests don't return empty binaries, do they? In case an empty binary is passed to `jsx:decode` here, the call should fail, and rightfully so: it's an exceptional situation.

src/erlcloud_guardduty.erl
kstarikov: I changed it to be more meaningful.

kstarikov:
@motobob all addressed.

PR #469, opened 2018-01-05
Add subscription attributes support

MikeBenza:
Adds support for subscription attributes in SNS.  Getting returns the same keys that may be set.


MikeBenza:
Note: this is not affected by #470, though some of the attributes (the ones not specific to subscription attributes) are lowercased.  This was done in anticipation of #470 being fixed.

PR #472, opened 2018-01-10
Protece auto_config/1 against default_config_region/2 returning undef…

lehoff:
…ined

default_config_region/2 can return undefined, so we cannot blindly wrap that result in `{ok, ...}` as that would
violate the return type for auto_config/1, which is `{ok, #aws_config{}} | undefined`

### DIFF
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -457,7 +457,7 @@ auto_config( ProfileOptions ) ->
     % the result in a {ok, ...} tuple.
     case default_config_region(Cfg, default_config_get(?AWS_REGION, aws_region)) of
         undefined  ->
-            undefined
+            undefined;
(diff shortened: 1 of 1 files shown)


src/erlcloud_aws.erl
```diff
@@ -446,11 +446,21 @@ auto_config() ->
...
+            {ok, _Config} = Result ->
+                Result;
+            {error, _} ->
+                auto_config_profile( ProfileOptions )
+        end,
+    % Since default_config_region/2 can return undefined we cannot blindly wrap
+    % the result in a {ok, ...} tuple.
+    case default_config_region(Cfg, default_config_get(?AWS_REGION, aws_region)) of
+        undefined  ->
+            undefined
```
motobob: ;

Issue #204, opened 2015-09-07
Authorization being in process dictionary is unexpected / unintuitive


jacobgreenleaf:
`erlcloud_s3:configure(X, Y)` configures the `erlcloud_s3` module to use access key X, but only for the caller, since it uses the process dictionary. This is unexpected; other libraries such as `httpc` that have default configuration usually do such configuration for all callers. 

Using `os:putenv` is an _okay_ workaround, but it would be great if it had an API similar to `httpc`, preferably using built-in `application:get_env`.

ransomr:
I agree this is an issue and I think using application environment would be an improvement. I'd be open to a PR that does that as long as it wouldn't break current clients. As I work-around I just pass the config explicitly into all ercloud methods.

jacobgreenleaf:
Unfortunately there is an issue with uploading large files that causes a crash because the node runs out of memory. I think it is related to the way the worker gen_server handles the binaries it gets from reading each piece of the file. I'm keeping this PR open for overall review but don't advise merging.

motobob:
@jacobgreenleaf there are uses cases when even `application:set_env/get_env` is not *okay*. Imagine a common situation with a server running `erlcloud` pointed to several different AWS accounts and regions. 
Passing different `#aws_config{}` records to the calling API is not just a work-around but a preferable option, IMO. That's why these API funs exist in every module.

Let's me know if you question has been answered and if I can close the issue

ondevsergeyl:
I'm worry about this issue also. I think config should be passed via arguments of functions. It is best way.

It is not good, when we can have garbage in process dictionary (PD). 

What happens if i will have 1M processes and will call s3-functions within each process. It means i will have 1M records inside PD. As minimum i should be able to delete this config from PD and this opportunity should be provided by this library.

Thanks.

loguntsov:
I'm worry about this issue also. I think config should be passed via arguments of functions. It is best way.

It is not good, when we can have garbage in process dictionary (PD).

What happens if i will have 1M processes and will call s3-functions within each process. It means i will have 1M records inside PD. As minimum i should be able to delete this config from PD and this opportunity should be provided by this library.

Thanks.

motobob:
@ondevsergeyl almost all API functions in all modules allow you to pass `#Aws_config{}` records and there is no requirement to use configure API at. that exists solely for the use case when the library is deployed in an application sharing a single identity. 
this is widely used in our code where we operate with thousands of different configs for different AWS accounts. 
[...]

PR #476, opened 2018-02-09
AWS Cost and Usage Report API

alexturkin:
* `describe_report_definitions` for Cost and Usage Report (https://docs.aws.amazon.com/cli/latest/reference/cur/describe-report-definitions.html)
* output is a map


src/erlcloud_cur.erl
```diff
@@ -0,0 +1,133 @@
...
+%% {ok, R} = erlcloud_cur:describe_report_definitions(Config).
+%% `
+%%
+-spec describe_report_definitions(aws_config()) -> {ok, map()} | {error, any()}.
+describe_report_definitions(Config) ->
+    describe_report_definitions(#{}, Config).
+
+
+%% `
+%% Opts = #{<<"MaxResults">> => 5},
```
alexturkin: from https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/describe-report-definitions.html:

```
MaxResults
Optional integer. Specifies the maximum number of results to return in response.
Type: Integer
Valid value: 5
Required: No
```

* so it's possible only to have it equal 5, and
* also it looks like 5 is maximum number of reports which is possible (the button becomes inactive in UI when trying to create > 5)
* I haven't found documentation on whether there is `next_token`, so I left this part of retrieving all objects not implemented (although it's already possible to submit token as part of options)

src/erlcloud_cur.erl
```diff
@@ -0,0 +1,133 @@
...
+
+%% @doc
+%% Cost & Usage Report API:
+%% https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/describe-report-definitions.html
+%%
+%% `
+%% {ok, R} = erlcloud_cur:describe_report_definitions(Config).
+%% `
+%%
+-spec describe_report_definitions(aws_config()) -> {ok, map()} | {error, any()}.
```
alexturkin: current output:

```
{ok,#{<<"ReportDefinitions">> =>
          [#{<<"AdditionalArtifacts">> => [],
             <<"AdditionalSchemaElements">> => [],
             <<"Compression">> => <<"GZIP">>,
             <<"Format">> => <<"textORcsv">>,
             <<"ReportName">> => <<"test-report">>,
             <<"S3Bucket">> => <<"user-test">>,
             <<"S3Prefix">> => <<"prefix">>,
             <<"S3Region">> => <<"us-east-1">>,
             <<"TimeUnit">> => <<"DAILY">>},
           #{<<"AdditionalArtifacts">> => [],
             <<"AdditionalSchemaElements">> => [],
             <<"Compression">> => <<"GZIP">>,
             <<"Format">> => <<"textORcsv">>,
             <<"ReportName">> => <<"test-report3">>,
             <<"S3Bucket">> => <<"user-test">>,<<"S3Prefix">> => <<>>,
             <<"S3Region">> => <<"us-east-1">>,
             <<"TimeUnit">> => <<"DAILY">>}
         ]}}.
```

which is basically map version of the json which is return by AWS. 

Shall we leave it like that or possible do list of maps like:

```
{ok,[#{<<"AdditionalArtifacts">> => [],
             <<"AdditionalSchemaElements">> => [],
             <<"Compression">> => <<"GZIP">>,
             <<"Format">> => <<"textORcsv">>,
             <<"ReportName">> => <<"test-report">>,
             <<"S3Bucket">> => <<"user-test">>,
             <<"S3Prefix">> => <<"prefix">>,
             <<"S3Region">> => <<"us-east-1">>,
             <<"TimeUnit">> => <<"DAILY">>},
           #{<<"AdditionalArtifacts">> => [],
             <<"AdditionalSchemaElements">> => [],
             <<"Compression">> => <<"GZIP">>,
             <<"Format">> => <<"textORcsv">>,
             <<"ReportName">> => <<"test-report3">>,
             <<"S3Bucket">> => <<"user-test">>,<<"S3Prefix">> => <<>>,
             <<"S3Region">> => <<"us-east-1">>,
             <<"TimeUnit">> => <<"DAILY">>}
         ]}.
```

src/erlcloud_cur.erl
```diff
@@ -0,0 +1,133 @@
+-module(erlcloud_cur).
```
alexturkin: although the docs for Cost&Usage Reports are placed together with budget API: https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/api-reference.html

in aws cli those are separate entities:

```
aws cur help
aws budgets help
```

So I decided to keep it in separate module, to have names compatible with `aws cli`

src/erlcloud_cur.erl
alexturkin: Actually, `aws cur describe-report-definitions help` says:

```
NextToken -> (string)
          A generic string.
```

so I can implement describe_report_definitions_all which would retrieve all reports, and this would work once in future it changed (provided they really will use `NextToken` as I don't how to test it now)

alexturkin:
@kkuzmin please review

src/erlcloud_cur.erl
kkuzmin: I'm ok with current map output because AWS may add more properties at the same level with `ReportDefinitions`

src/erlcloud_cur.erl
kkuzmin: Let's keep it as is and have `describe_report_definitions_all` implemented once the doc is updated with `NextToken` information.

src/erlcloud_cur.erl
```diff
@@ -0,0 +1,133 @@
...
+%%------------------------------------------------------------------------------
+
+request(Config0, Operation, Input) ->
+    case erlcloud_aws:update_config(Config0) of
+        {ok, Config} ->
+            Body       = jsx:encode(Input),
+            Headers    = get_headers(Config, Operation, Body),
+            AwsRequest = #aws_request{service         = cur,
+                                      uri             = get_url(Config),
+                                      method          = post,
```
kkuzmin: Is it always POST? What about GETs?

alexturkin: it is `post` for this service, even for `Describe*` functionality. (Probably the reason behind is that we actually have the query inside of the request body). If it's not the same is for other operations for `CUR` service (which is `DeleteReportDefinition` and `PutReportDefinition`(, I suggest to change it later when we add it

alexturkin:
@kkuzmin shall we merge it?

kkuzmin:
@motobob @nalundgaard could you please help us to review/merge this change?

src/erlcloud_cur.erl
```diff
@@ -0,0 +1,133 @@
+-module(erlcloud_cur).
+
+
+%% AWS Cost and Usage Report API implementation
+%% https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/api-reference.html
+
+-include("erlcloud_aws.hrl").
+
+%% Initialization API
+-export([new/2, new/3, new/4, new/5]).
```
alexturkin: @motobob do you think it's ok to remove `configure` functions and ban using of process dictionary for this new module?

re https://github.com/erlcloud/erlcloud/issues/204

src/erlcloud_cur.erl
motobob: no they will not

motobob: > Returns one or more ReportDefinition objects.

from AWS docs. yes they wrap it in one more `ReportDefinitions`. this given API will note get paginated. so i'll remove unnecessary wrapping

src/erlcloud_cur.erl
motobob: it's a GET but via a POST with a body. so this is fine

src/erlcloud_cur.erl
motobob: token is to be passed as `Options`. no more *_all funs

PR #477, opened 2018-02-14
US73085 - Add AWS config support

ikemsley:
**Problem**

Erlcloud does not have support for AWS Config.

**Solution**

Add it.


ikemsley:
@kkuzmin @alexturkin @tomdos Please review.

src/erlcloud_config.erl
```diff
@@ -0,0 +1,632 @@
...
+
+-spec describe_compliance_by_config_rule(Opts :: map()) ->
+    result_paged(map()).
+describe_compliance_by_config_rule(Opts) ->
+    describe_compliance_by_config_rule(Opts, default_config()).
+
+-spec describe_compliance_by_config_rule(Opts :: map(),
+    Config :: aws_config()) ->
+    result_paged(map()).
+describe_compliance_by_config_rule(Opts, Config) when is_map(Opts) orelse Opts =:= #{} ->
```
tomdos: I would say it is not a good idea in general to have a map (`Opts` in this case) as an input parameter.  In order to use this method, I need look into the spec and then search below in maps. 
```
+    Request = #{?COMPLIANCE_TYPES => maps:get(?COMPLIANCE_TYPES, Opts, []),
+        ?CONFIG_RULE_NAMES        => maps:get(?CONFIG_RULE_NAMES, Opts, []),
+        ?NEXT_TOKEN_LABEL         => maps:get(?NEXT_TOKEN_LABEL, Opts, <<"">>)
+    }
```

There are only macros thus I have to go through macros to find how to call parameters in my input map. Moreover, I can find what is in a map however it doesn't contain data type of values. 

As an user of this interface, I see it a bit difficult. I think it would be better to have all method parameters instead of a single map.

test/erlcloud_config_tests.erl
```diff
@@ -0,0 +1,513 @@
...
+        erlcloud_config:describe_compliance_by_config_rule(Request)
+               end,
+    do_test(Request, Expected, TestFun).
+
+test_describe_compliance_by_resource() ->
+    Request = #{?COMPLIANCE_TYPES => ?TEST_COMPLIANCE_TYPES,
+        ?LIMIT                    => ?TEST_LIMIT,
+        ?RESOURCE_ID              => ?TEST_RESOURCE_ID,
+        ?RESOURCE_TYPE            => ?TEST_RESOURCE_TYPE,
+        ?NEXT_TOKEN_LABEL         => <<"">>},
```
alexturkin: Interface concern:
It does not seen obvious to me to add empty token for the first query. Will the call work if I omit the next token?

src/erlcloud_config.erl
ikemsley: I infer it as a best practice based on extending the logic here to its logical conclusion: https://github.com/ikemsley/erlcloud/blame/master/README.md#L184

Personally I much prefer it this way otherwise functions get an explosion of brittle and unwieldy arguments that are not safe since its easy to mix up their order, and you still have to look up their meaning anyway!  At least this way you can rtfm on the Amazon doc.

src/erlcloud_config.erl
```diff
@@ -0,0 +1,632 @@
...
+    end.
+
+%%------------------------------------------------------------------------------
+%% @doc
+%% DescribeConfigRules
+%%
+%% https://docs.aws.amazon.com/config/latest/APIReference/API_DescribeConfigRules.html
+%%------------------------------------------------------------------------------
+-spec describe_config_rules() ->
+    result_paged(map()).
```
alexturkin: for every function that returns `result_paged` I think we should add `describe_something_all` function, for those who just want to get it all :)

Examples where we have it already:
* https://github.com/erlcloud/erlcloud/blob/b50123a62da79/src/erlcloud_autoscaling.erl#L68
* https://github.com/erlcloud/erlcloud/blob/b50123a62da79/src/erlcloud_cloudformation.erl#L81
* https://github.com/erlcloud/erlcloud/blob/b50123a62da79/src/erlcloud_ec2.erl#L1872
* https://github.com/erlcloud/erlcloud/blob/b50123a62da79/src/erlcloud_elb.erl#L216
* https://github.com/erlcloud/erlcloud/blob/b50123a62da79/src/erlcloud_iam.erl#L132
* https://github.com/erlcloud/erlcloud/blob/b50123a62da79/src/erlcloud_rds.erl#L119
* https://github.com/erlcloud/erlcloud/blob/b50123a62da79/src/erlcloud_redshift.erl#L99
* https://github.com/erlcloud/erlcloud/blob/b50123a62da79/src/erlcloud_route53.erl#L150
* https://github.com/erlcloud/erlcloud/blob/b50123a62da79/src/erlcloud_sdb.erl#L279
* https://github.com/erlcloud/erlcloud/blob/b50123a62da79/src/erlcloud_sns.erl#L320

src/erlcloud_config.erl
tomdos: @ikemsley I don't know what is the best practice in erlcloud. I would say @motobob meant that `NextToken` and `NextMarker` should be in `Opts` not all input parameters in general. 
If it is recommended way then it is fine, I am just saying it is a bit difficult to follow.  

FYI @motobob

src/erlcloud_config.erl
```diff
@@ -0,0 +1,632 @@
...
+new(AccessKeyID, SecretAccessKey, Host, Port, Scheme) ->
+    #aws_config{access_key_id     = AccessKeyID,
+                secret_access_key = SecretAccessKey,
+                config_host       = Host,
+                config_port       = Port,
+                config_scheme     = Scheme,
+                retry             = fun erlcloud_retry:default_retry/1}.
+
+-spec configure(string(), string()) -> ok.
+configure(AccessKeyID, SecretAccessKey) ->
```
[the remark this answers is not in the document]
alexturkin: (PS let's keep it as for now, as there is no reached agreement on whether it's really bad or good to use process dictionary)

src/erlcloud_config.erl
alexturkin: It is not written in Best Practices doc (and I want to add it there;) that we should:

* use Requires parameters as function param
* use optional parameters (including pagination stuff) as Options (map or proplist)

This way it makes it clear what is optional and what is not, and you get checking of required parameters presence in the compile time! (You still have to check the order yourself though)

src/erlcloud_config.erl
```diff
@@ -0,0 +1,632 @@
...
+    ok.
+
+%%------------------------------------------------------------------------------
+%% AWS Config API Functions
+%%------------------------------------------------------------------------------
+
+%%------------------------------------------------------------------------------
+%% @doc
+%% DescribeComplianceByConfigRule
+%%
```
alexturkin: @ikemsley I think what makes the description really good, is an example of input (especially if it's map with optional arguments). Could you please add it there?

Example: https://github.com/erlcloud/erlcloud/blob/b50123a62da79/src/erlcloud_glue.erl#L184-L186

alexturkin: Or actually, it could be seen in the tests module. probably not the best idea to duplicate it everywhere then. 

So please disregard my previous comment :)

src/erlcloud_config.erl
```diff
@@ -0,0 +1,632 @@
...
+%%% Internal Functions
+%%%------------------------------------------------------------------------------
+
+default_config() -> erlcloud_aws:default_config().
+
+request(Config0, OperationName, Request) ->
+    case erlcloud_aws:update_config(Config0) of
+        {ok, Config} ->
+            Body       = jsx:encode(Request),
+            Operation  = "StarlingDoveService." ++ OperationName,
```
kkuzmin: where does `StarlingDoveService` come from? I failed to find anything related on AWS website.

src/erlcloud_config.erl
ikemsley: Okay, I'll do it like that.  Most of the parameters are optional anyway so I'm not sure anything will change much.  However, I'll go through carefully and check.

src/erlcloud_config.erl
alexturkin: @kkuzmin one of the ways is to see it turning debug mode for `aws cli`:

```
$ aws configservice describe-config-rules --profile mine --region us-east-1 --debug
2018-02-15 09:57:38,827 - MainThread - awscli.clidriver - DEBUG - CLI version: aws-cli/1.11.180 Python/3.6.3 Darwin/17.3.0 botocore/1.7.38
2018-02-15 09:57:38,827 - MainThread - awscli.clidriver - DEBUG - Arguments entered to CLI: ['configservice', 'describe-config-rules', '--profile', 'mine', '--region', 'us-east-1', '--debug']

...


2018-02-15 09:57:39,134 - MainThread - botocore.endpoint - DEBUG - Making request for OperationModel(name=DescribeConfigRules) (verify_ssl=True) with params: {'url_path': '/', 'query_string': '', 'method': 'POST', 'headers': {'X-Amz-Target': 'StarlingDoveService.DescribeConfigRules', 'Content-Type': 'application/x-amz-json-1.1', 'User-Agent': 'aws-cli/1.11.180 Python/3.6.3 Darwin/17.3.0 botocore/1.7.38'}, 'body': b'{}', 'url': 'https://config.us-east-1.amazonaws.com/', 'context': {'client_region': 'us-east-1', 'client_config': <botocore.config.Config object at 0x105c435c0>, 'has_streaming_input': False, 'auth_type': None}}

...

content-type:application/x-amz-json-1.1
host:config.us-east-1.amazonaws.com
x-amz-date:20180215T095739Z
x-amz-target:StarlingDoveService.DescribeConfigRules

content-type;host;x-amz-date;x-amz-target

...


{
    "ConfigRules": []
}
```

ikemsley: That came from hard graft.  I couldn't get this to work no matter what I tried.  Even the AWS console does not use the API directly, so it's hard to reverse engineer, and no doc tells you explicitly what to use.  Eventually I just got the AWS SDK and used Wireshark to see what it was using.  And presto!  It was "StarlingDoveService" and that worked. Weird, I know.  But what's a chap to do?

src/erlcloud_config.erl
```diff
@@ -0,0 +1,632 @@
...
+list_discovered_resources(Opts, Config) when is_map(Opts) orelse Opts =:= #{} ->
+    Request = #{?INCLUDE_DELETED_RESOURCES => maps:get(?INCLUDE_DELETED_RESOURCES, Opts, true),
+        ?LIMIT                             => maps:get(?LIMIT, Opts, ?LIMIT_MAX),
+        ?NEXT_TOKEN_LABEL                  => maps:get(?NEXT_TOKEN_LABEL, Opts, <<"">>),
+        ?RESOURCE_IDS                      => maps:get(?RESOURCE_IDS, Opts, []),
+        ?RESOURCE_NAME                     => maps:get(?RESOURCE_NAME, Opts, null),
+        ?RESOURCE_TYPE                     => maps:get(?RESOURCE_TYPE, Opts, null)},
+    case request(Config, "ListDiscoveredResources", Request) of
+        {ok, Res} ->
+            {ok, Res, get_next_token(Res)};
```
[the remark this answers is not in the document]
ikemsley: I just followed the README. I have no preference either way.

src/erlcloud_config.erl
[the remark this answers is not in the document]
alexturkin: updates: my https://github.com/erlcloud/erlcloud/pull/476 has been merged without them

src/erlcloud_config.erl
alexturkin: btw spekaing of opposite point of view, @motobob posted here 
 https://github.com/erlcloud/erlcloud/pull/476#discussion_r168361548 "no more _all funs", so sound like something to agree on & probably add to best practices section

ikemsley: Well okay.  I have to admit that I'm rather stunned by this.  

In my opinion implementing "get all" functions without limits or pagination is a software anti-pattern of gargantuan proportions.  So I confidently went off to Google it, confident that I would find about a million software 101 examples of why you should never, never, do this.  And guess what, I was amazed to find that I couldn't really find a single mention!  

This has left me all rather confused, because decades ago implementing a `describe_something_all` function, without limits or pagination, used to be such an obvious anti-pattern that it would instantly eliminate candidates from consideration in a job interview.  Now I'm uncertain as to whether our industry just got amnesia along the way or somewhere along the line I fell asleep.  Now possibly I'm a software dinosaur and I'm waking up to a new reality were all limitations on memory are considered fictional these days and crashing the system because of OOM conditions based on the arguments passed, without defensive programming, is now devolved to be the responsibility of calling client modules as a best practice.  I just hope I'm not the guy on call when the first customer has enough `config` resources to crash the CE process passes the critical threshold! 

Still, I realize that an even worse sin is not following the patterns and spirit of implementation of a software module.  Even if it's written in Swahili, maintainers should stick to Swahili.  So if you really, really, want to, I will hold my nose and implement  `describe_something_all` everywhere.  But you are going to have to justify it to me just one more time.

src/erlcloud_config.erl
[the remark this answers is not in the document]
alexturkin: I think README has  a bit of ambiguity on that, as it's just says about preferring one way to another, and the third way for records...

I'm up for `{ok, Res}` approach rather then `{ok, Results, Marker}` because it will be up to customer whether he wants to actually retrieve the token or not, so we don't do excessive action.

src/erlcloud_config.erl
ikemsley: I get it.  Except the issue is not whether it's useful or not.  Sure it's useful.  The problem is that it's dangerous.  Mostly it violates elementary principals of https://en.wikipedia.org/wiki/Defensive_programming.   Traditionally, you should never, never, implement a function that is guaranteed to fail in certain conditions based on the arguments you pass (or lack of them).  I can give you a guaranteed instance where this will crash: just give me a customer with more config results than will fit in working storage.  

It's an anti-pattern from so many perspectives they are hard to even list.  Being able to crash a system on demand is bad from a security perspective too, for a long list of reasons.

This pattern is also vaguely reminiscent of the https://docs.microsoft.com/en-us/azure/architecture/antipatterns/extraneous-fetching/ anti-pattern.

As I said, if everyone insists on this (and given the precedent in the module) I'll bow to the local zeitgeist and implement it.  But it churns my gut.  I'll just add it to a growing list of software war stories I can tell at the bar one day to old has-beens like myself.

ikemsley: @kkuzmin @motobob What's your take on this?

ikemsley:
@alexturkin @kkuzmin @tomdos Please would you review once more before I merge with Alex's changes.

src/erlcloud_config.erl
kkuzmin: @ikemsley @alexturkin I totally agree that having _all is an anti-pattern. These functions were introduced as convenient utils that one can use at his own risk accepting possible OOMs.
I like the idea of adding `erlcloud_util:process_all()` which accepts a user defined callback to process a result page. But this approach requires all `describe_*`/`list_*` function to have  common pagination interface.

ikemsley:
@alexturkin @kkuzmin @motobob Please review and merge.

src/erlcloud_config.erl
```diff
@@ -0,0 +1,370 @@
...
+%%
+%% https://docs.aws.amazon.com/config/latest/APIReference/API_DescribeComplianceByConfigRule.html
+%%------------------------------------------------------------------------------
+-spec describe_compliance_by_config_rule(Opts :: map(),
+    Config :: aws_config()) ->
+    result_paged(map()).
+describe_compliance_by_config_rule(Opts, Config) when is_map(Opts) orelse Opts =:= #{} ->
+    Request = #{?COMPLIANCE_TYPES => maps:get(?COMPLIANCE_TYPES, Opts, []),
+        ?CONFIG_RULE_NAMES        => maps:get(?CONFIG_RULE_NAMES, Opts, []),
+        ?NEXT_TOKEN_LABEL         => maps:get(?NEXT_TOKEN_LABEL, Opts, null)
```
kkuzmin:
```
1> jsx:encode(#{<<"NextToken">> => null}).
<<"{\"NextToken\":null}">>
```
@ikemsley have you checked that AWS API handles `null`s? Same question about other optional parameters below.

ikemsley: Si, senor.  It handles nulls.

PR #478, opened 2018-02-14
kms - fix data types in spec

tomdos:
Almost all kms specs requires `string()` as an input though most of them should be `binary()`. I think this has been wrong since the beginning as even tests are with binary input. 

@nalundgaard @motobob

### DIFF
--- a/src/erlcloud_kms.erl
+++ b/src/erlcloud_kms.erl
@@ -114,16 +114,16 @@ default_config() ->
 %%
 %%------------------------------------------------------------------------------
 -spec create_alias
-          (AliasName :: string(),
-           TargetKeyId :: string()) ->
+          (AliasName :: binary(),
+           TargetKeyId :: binary()) ->
           return_val().
 create_alias(AliasName, TargetKeyId) ->
     create_alias(AliasName, TargetKeyId, default_config()).
 
 
 -spec create_alias
-          (AliasName :: string(),
-           TargetKeyId :: string(),
+          (AliasName :: binary(),
+           TargetKeyId :: binary(),
            Config :: aws_config()) ->
           return_val().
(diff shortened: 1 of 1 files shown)


tomdos:
@kkuzmin @ikemsley @alexturkin

kkuzmin:
@tomdos maybe have both `string()` and `binary()`?

tomdos:
@kkuzmin I would rather keep only `binary()`. We don't have normally both options in other modules.

kkuzmin:
I'd also prefer to have some `is_list()`/`is_binary()` guard statements, but worrying about braking existing consumers.
@motobob @nalundgaard you call.

tomdos:
@kkuzmin I was thinking about it and I don't like "changing interface" though I think this is not the case. `string()` has never ever worked thus we can not break any existing customers.

src/erlcloud_kms.erl
```diff
@@ -114,16 +114,16 @@ default_config() ->
 %%
 %%------------------------------------------------------------------------------
 -spec create_alias
-          (AliasName :: string(),
-           TargetKeyId :: string()) ->
+          (AliasName :: binary(),
```
alexturkin: @kkuzmin @tomdos 
speaking of whether we should have `string()|binary()` here, I disagree with the idea of having two options, why we would need that at all?

`erlcloud_kms` gives you binaries on output (as it uses `jsx:decode`) and ask for binaries, which sounds logical and consistent. Having separate `string()` input sounds like something out of the place to me.

motobob:
it's all just back to `jsx:encode()` as @alexturkin mentioned. 
specs changes do not change the API ;) they change description of the API and this is a clear fix of that. similar to #478

PR #480, opened 2018-02-15
fix erlcloud_aws:default_config_region()

key-master:
## Problem:
https://github.com/erlcloud/erlcloud/pull/476 broke `erlcloud_aws:default_config_region()`

## Solution: 
add `cur` service to `erlcloud_aws:service_config()`

@motobob please take a look
@alexturkin FYI

### DIFF
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -698,7 +698,10 @@ service_config( <<"athena">> = Service, Region, Config ) ->
 service_config(<<"cloudwatch_logs">>, Region, Config)->
     Host = service_host(<<"logs">>, Region),
     Config#aws_config{cloudwatch_logs_host = Host};
-service_config( <<"waf">>, _Region, Config ) -> Config.
+service_config( <<"waf">>, _Region, Config ) -> Config;
+service_config( <<"cur">>, Region, Config ) ->
+    Host = service_host(<<"cur">>, Region),
+    Config#aws_config{cur_host = Host}.
 
 %%%---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


test/erlcloud_aws_tests.erl
```diff
@@ -360,6 +360,10 @@ profiles_assume_cleanup(P) ->
     profiles_test_cleanup(P),
     meck:unload( erlcloud_sts ).
 
+default_config_region_sunny_test() ->
+    Region = <<"ca-central-1">>,
```
motobob: should not this be us-east-1? why ca-central-1 ?

src/erlcloud_aws.erl
```diff
@@ -698,7 +698,10 @@ service_config( <<"athena">> = Service, Region, Config ) ->
 service_config(<<"cloudwatch_logs">>, Region, Config)->
     Host = service_host(<<"logs">>, Region),
     Config#aws_config{cloudwatch_logs_host = Host};
-service_config( <<"waf">>, _Region, Config ) -> Config.
+service_config( <<"waf">>, _Region, Config ) -> Config;
```
motobob: time to refactor this fun...as why does not WAF has the service_host call?

test/erlcloud_aws_tests.erl
key-master: here may be just `<<"some region">>`. that does not matter for this test.

src/erlcloud_aws.erl
key-master: As I understand, `waf` endpoint does not include region. e.g. `waf.amazonaws.com`

PR #481, opened 2018-02-21
erlcloud_ddb2: Add support for DynamoDB Global Table management

nalundgaard:
Adds functions to erlcloud_ddb2 to implement the following:

* CreateGlobalTable
* DescribeGlobalTable
* ListGlobalTables
* UpdateGlobalTable

### DIFF
--- a/src/erlcloud_ddb2.erl
+++ b/src/erlcloud_ddb2.erl
@@ -89,19 +89,23 @@
 %%% DynamoDB API
 -export([batch_get_item/1, batch_get_item/2, batch_get_item/3,
          batch_write_item/1, batch_write_item/2, batch_write_item/3,
+         create_global_table/2, create_global_table/3, create_global_table/4,
          create_table/5, create_table/6, create_table/7,
          delete_item/2, delete_item/3, delete_item/4,
          delete_table/1, delete_table/2, delete_table/3,
+         describe_global_table/1, describe_global_table/2, describe_global_table/3,
          describe_limits/0, describe_limits/1, describe_limits/2,
          describe_table/1, describe_table/2, describe_table/3,
          describe_time_to_live/1, describe_time_to_live/2, describe_time_to_live/3,
          get_item/2, get_item/3, get_item/4,
+         list_global_tables/0, list_global_tables/1, list_global_tables/2,
          list_tables/0, list_tables/1, list_tables/2,
          put_item/2, put_item/3, put_item/4,
          %% Note that query is a Erlang reserved word, so we use q instead
          q/2, q/3, q/4,
          scan/1, scan/2, scan/3,
          update_item/3, update_item/4, update_item/5,
+         update_global_table/2, update_global_table/3, update_global_table/4,
          update_table/2, update_table/3, update_table/4, update_table/5,
          update_time_to_live/2, update_time_to_live/3, update_time_to_live/4
         ]).
@@ -182,6 +186,9 @@
     q_return/0,
     range_key_name/0,
     read_units/0,
+    replica/0,
+    replica_description/0,
+    replica_update/0,
     return_consumed_capacity/0,
     return_consumed_capacity_opt/0,
     return_item_collection_metrics/0,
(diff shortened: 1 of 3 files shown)


src/erlcloud_ddb2.erl
```diff
@@ -1871,6 +2027,69 @@ get_item(Table, Key, Opts, Config) ->
...
+     ]}.
+
+-type list_global_tables_return() :: ddb_return(#ddb2_list_global_tables{}, [#ddb2_global_table{}]).
+
+-spec list_global_tables() -> list_global_tables_return().
+list_global_tables() ->
+    list_global_tables([], default_config()).
+
+-spec list_global_tables(list_global_tables_opts()) -> list_global_tables_return().
+list_global_tables(Opts) ->
```
motobob: add a pedantic guard for `when is_list(Opts)`
as other apis allow `/1 (#aws_config{})`

src/erlcloud_ddb2.erl
```diff
@@ -1644,6 +1753,53 @@ delete_table(Table, Opts, Config) ->
...
+     ]}. 
+
+-type describe_global_table_return() :: ddb_return(#ddb2_describe_global_table{}, #ddb2_global_table_description{}).
+
+-spec describe_global_table(table_name()) -> describe_table_return().
+describe_global_table(GlobalTableName) ->
+    describe_global_table(GlobalTableName, [], default_config()).
+
+-spec describe_global_table(table_name(), ddb_opts()) -> describe_table_return().
+describe_global_table(GlobalTableName, Opts) ->
```
motobob: guard for second argument here too.

src/erlcloud_ddb2.erl
```diff
@@ -766,6 +776,29 @@ undynamize_expression(Expression, _) ->
...
+undynamize_replica_description(ReplicaDescription, _) ->
+    #ddb2_replica_description{region_name = proplists:get_value(<<"RegionName">>, ReplicaDescription)}.
+
+-spec undynamize_replica_descriptions(jsx:json_term(), undynamize_opts()) -> [replica_description()].
+undynamize_replica_descriptions(ReplicaDescriptions, Opts) ->
+    [undynamize_replica_description(ReplicaDescription, Opts)
+     || ReplicaDescription <- ReplicaDescriptions].
+
+-spec undynamize_replica(jsx:json_term(), undynamize_opts()) -> replica().
+undynamize_replica(Replica, _) ->
```
motobob: so sad these two c/p funs differ only in the actual record used...but this is outside of scope of the PR

src/erlcloud_ddb2.erl
```diff
@@ -2356,6 +2575,66 @@ update_item(Table, Key, UpdatesOrExpression, Opts, Config) ->
...
+-type update_global_table_return() :: ddb_return(#ddb2_update_global_table{}, #ddb2_global_table_description{}).
+
+-spec update_global_table(table_name(), maybe_list(replica_update()))
+                         -> update_global_table_return().
+update_global_table(GlobalTableName, ReplicaUpdates) ->
+    update_global_table(GlobalTableName, ReplicaUpdates, [], default_config()).
+
+-spec update_global_table(table_name(), maybe_list(replica_update()), ddb_opts())
+                         -> update_global_table_return().
+update_global_table(GlobalTableName, ReplicaUpdates, Opts) ->
```
motobob: guard here?

src/erlcloud_ddb2.erl
nalundgaard: Yeah, I decided to shadow their difference for [Replica](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Replica.html) and [ReplicaDescription](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_ReplicaDescription.html). They are symmetric at this time, but it appears to me that they may change in the future, in particular, I think ReplicaDescription will gain additional fields once they, er... finish it?

src/erlcloud_ddb2.erl
```diff
@@ -766,6 +777,29 @@ undynamize_expression(Expression, _) ->
...
+-spec undynamize_replica_descriptions(jsx:json_term(), undynamize_opts()) -> [replica_description()].
+undynamize_replica_descriptions(ReplicaDescriptions, Opts) ->
+    [undynamize_replica_description(ReplicaDescription, Opts)
+     || ReplicaDescription <- ReplicaDescriptions].
+
+-spec undynamize_replica(jsx:json_term(), undynamize_opts()) -> replica().
+undynamize_replica(Replica, _) ->
+    #ddb2_replica{region_name = proplists:get_value(<<"RegionName">>, Replica)}.
+
+-spec undynamize_replicas(jsx:json_term(), undynamize_opts()) -> [replica_description()].
```
paulo-ferraz-oliveira: Assuming `undynamize_replica/2` is -spec().ed OK, this function actually receives `([jsx:json_term()], _)` and outputs `[replica()]`.

Issue #483, opened 2018-02-23
erlcloud_ddb2:global_table_status/0 doesn't exist?


paulo-ferraz-oliveira:
dialyzer just complained about this in 3.1.1: erlcloud_ddb2:global_table_status/0

nalundgaard:
Looks like my mistake, I namespaced a type in the erlcloud_ddb2.hrl file. I will PR a fix.

paulo-ferraz-oliveira:
Thanks.

We've also detected that e.g. erlcloud_ec2.hrl uses type datetime() but doesn't include erlcloud.hrl (where the type is defined).

Some other hrl files do (include erlcloud.hrl), like erlcloud_as.hrl, but there's probably a better approach to avoid cyclic dependencies.

What if we (this breaks interface!) declare and export types only from .erl files. We can then have .hrl files as stand-alone.

So, we could have datetime in erlcloud_aws.erl or erlcloud.erl and then export it from there and import it (fully qualified) as erlcloud:datetime() where required.

paulo-ferraz-oliveira:
Another way some other projects have found to go around this issue is to declare a _types.erl module, e.g. erlcloud_types.erl where all types would be defined.

paulo-ferraz-oliveira:
datetime() is used in:
* erlcloud_as.hrl
* erlcloud_ec2.hrl
* erlcloud_mon.hrl
* erlcloud_mturk.hrl

and it's actually the only used content from erlcloud.hrl, since it also defines -type proplist() :: proplists:proplist(). but this is unused (at least internally).

motobob:
please feel free to send a PR to fix datetime().
including that hlr should be of cyclic harm. 

in general, @paulo-ferraz-oliveira do i understand correct that your top level project requires dializer passing?

nalundgaard:
@paulo-ferraz-oliveira in general I think I would agree with exporting shared types from modules correspondent to the headers they are in and referencing them in other headers accordingly.

I wouldn't move type definitions out of the headers, as like you say, that could break others' type references.

If there is any concern with header inclusion being cyclic, I think we should wrap the text of the headers in `ifndef` checks, e.g.:

```erlang
-ifndef(erlcloud_example_hrl).
-define(erlcloud_example_hrl).

# all header file definitions here

-endif.
```

That way, there's no issue with cross-references in headers that depend on other headers.

paulo-ferraz-oliveira:
[...]

PR #485, opened 2018-02-24
Fix cyclic includes

paulo-ferraz-oliveira:
As per the discussion in #483...

The main goals of this PR are to:
1. allow inclusion of .hrl files without having to worry about internal -include(...)s, e.g.
`-include("erlcloud_ec2.hrl").` without having to also `-include("erlcloud.hrl").`,
2. allow inclusion of .hrl files without worrying about doing it more than once (e.g.
`-include("erlcloud_mturk.hrl").` and `-include("erlcloud_as.hrl").` in the same module with no issues; there would be issues otherwise, because both header files include "erlcloud.hrl",
3. add tests for the fixed issues.

I hope it's OK to have deleted a comment in erlcloud_as.hrl and an empty line from erlcloud_ec2.hrl.

### DIFF
--- a/include/erlcloud.hrl
+++ b/include/erlcloud.hrl
@@ -1,3 +1,6 @@
+-ifndef(erlcloud_hrl).
+-define(erlcloud_hrl, 0).
+
 -type proplist() :: proplists:proplist().
 -type datetime() :: {{pos_integer(), 1..12, 1..31}, {0..23, 0..59, 0..60}}.
 -type paging_token() :: binary() | undefined.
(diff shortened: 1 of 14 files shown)


paulo-ferraz-oliveira:
Travis CI build failed for use of rebar 2 it seems. I'm looking into it.

paulo-ferraz-oliveira:
I forced push on top of the last commit to have `rebar eunit` work alongside `rebar3 eunit`,

nalundgaard:
@motobob if you also approve, will you merge?

paulo-ferraz-oliveira:
The tests were written with the following in mind: if, in the future, someone doesn't take this approach or if somehow the implemented "fix" is reverted, we're easily warned. In any case, they put in evidence a behavior (if the code is not accepted) that is not correct.

PR #486, opened 2018-02-24
Fix dialyzer warning introduced by #481

paulo-ferraz-oliveira:


paulo-ferraz-oliveira:
Thanks! It would be great if we reached the point of "no dialyzer issues" and this check (rebar dialyzer) could be added to the PR tests 😃

PR #487, opened 2018-02-28
Retry after getting 429

AntonZaets:
**Problem statement**:

AWS uses HTTP code 429 for throttling errors in [many](https://www.google.ru/search?q=site:https://docs.aws.amazon.com+"HTTP+Status+Code:+429"&newwindow=1&ei=9XGWWruRJ4uOsgHl9abAAQ&start=20&sa=N&biw=1886&bih=949) different services. For example, [AWS Lambda](https://docs.aws.amazon.com/lambda/latest/dg/API_Invoke.html):
```
TooManyRequestsException
HTTP Status Code: 429
```
In AWS Lambda case `erlcloud` returns:
```erlang
{error,
 {http_error, 429, "Too Many Requests",
 <<"{\"Reason\":\"ReservedFunctionConcurrentInvocationLimitExceeded\",\"Type\":\"User\",\"message\":\"Rate Exceeded.\"}">>}}
```
but it would be nice to do a retry for that.

**Solution**:

That PR:
- introduces retry after 429;
- adds test;
- fixes old tests for retry, because those didn't work as expected - causes `instantiation_failed` instead of usual test failure;

@motobob please take a look

### DIFF
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -241,7 +241,7 @@ aws_request_form_raw(Method, Scheme, Host, Port, Path, Form, Headers, Config) ->
                          response_status = Status} = Request) when
                 %% Retry for 400, Bad Request is needed due to Amazon
                 %% returns it in case of throttling
-                    Status == 400 ->
+                    Status == 400; Status == 429 ->
                 ShouldRetry = is_throttling_error_response(Request),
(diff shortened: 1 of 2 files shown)


motobob:
tnx that fixes many retries, including Kinesis I was planning to fix...

PR #488, opened 2018-03-06
Allow to override default aws_config values

ddosia:
**Problem**
Already built service should be tweakable in order to specify AWS endpoints/ports/protocols.
Example: locally, instead of using real AWS S3 I use a mock, which provides same HTTP interface. I want to adjust app.config of my service and switch it to custom S3 URL.

It might be done differently:
  1. having an /etc/hosts alias
  2. providing custom `#aws_config{}` to each and every `erlcloud` function

Number 1 doesn't really work well, since because of strange combination of `erlcloud` default values and `lhttpc` peculiarities it always using ssl, no matter what.
Number 2 solves the problem, but it is tedious.

**Solution**
Allow to override any given field of `#aws_config{}` record produced by `erlcloud_aws:default_config()` function.

It is achievable through application variables and has such syntax:

```erlang
[
  {erlcloud, [
      {aws_config, [
          {s3_scheme, "http://"},
          {s3_host, "s3.example.com"}
      ]}
  ]}
].
``` 
After this, ` "true = http://" == (erlcloud_aws:default_config())#aws_config.s3_scheme`.

### DIFF
--- a/test/erlcloud_aws_tests.erl
+++ b/test/erlcloud_aws_tests.erl
@@ -156,6 +156,7 @@ get_service_status_test(_) ->
      ?_assertEqual(OKStatus, ok)
      ].
 
+
 auto_config_with_env(_) ->
     % Note: meck do not support os module
     X_AWS_ACCESS  = os:getenv("AWS_ACCESS_KEY_ID"),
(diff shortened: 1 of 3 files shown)


motobob:
i'm confused why
> Number 2 solves the problem, but it is tedious.

that is the way to do it in an application which does multiple different calls to AWS, if role assumption is needed + supported everywhere (might be exceptions but these are gaps).

the approach in this PR can be considered but I would like to understand why passing` #aws_config{}` of your flavour is not an option. we use it everywhere...

ddosia:
@motobob 
So imagine this situation
```erlang
erlcloud_s3:erlcloud_s3(Bucket, Cfg#aws_config{s3_host = "s3.example.com"}).
```
In order to make it flexible in my app I have to write a layer of configuration so I can take this address from app env.

```erlang
erlcloud_s3:erlcloud_s3(
    Bucket,
    Cfg#aws_config{s3_host = application:get_env(my_app, s3_host, "s3.example.com")}
).
```

Then goes another record param, and another. Eventually my application has half of the `erlcloud` settings in the config.
And I've seen this pattern across several different applications, everyone want to configure some values. Why not to provide same interface for everyone, so it would be solved once and for good?

motobob:
I'm not sure why having a *single* wrapper config fun is not enough, especially if you talk changing stuff in runtime...

```erlang
my_special_config() ->
    {ok, Cfg} = erlcloud_aws:auto_config(),
    MyApp = application:get_env(my_app, s3_host, "s3.example.com"),
    ...
    Cfg#aws_config{s3_host = MyApp, timeout = 5000, retry = 3, ...}
```
but fine this way can also work if you really like.

ddosia:
@motobob say, I have already built image of a service which uses S3. Service developer haven't thought in advance that someone wants to configure S3 address and left no way to configure it through app.config. When I try to use this image and run locally across fake S3 I have very few options to tune it.
But if there was a way not to depend on a service developer but just tune `erlcloud` that would make my life easier 😉.

motobob:
the usecase described is so unreal...just due to this statement
>Service developer haven't thought in advance that someone wants to configure S3 address and left no way to configure it 

but anyway.

test/erlcloud_aws_tests.erl
```diff
@@ -156,6 +157,29 @@ get_service_status_test(_) ->
      ?_assertEqual(OKStatus, ok)
      ].
 
+
+%% Tests that if application environment variable "erlcloud.aws_config" contains
+%% overrides for default `#aws_config{}', those overrides are indeed applied.
+default_config_override_test(_) ->
```
kkuzmin: Please add one more test to cover `application:get_env(erlcloud, aws_config) = undefined` case

ddosia:
@kkuzmin 
> My only concern is the performance of this approach

I can investigate how to *generate* setters for the record, so this transformation wouldn't be necessary.

nalundgaard:
I don't see the point of this either, but I'm not fundamentally opposed to it.

I'm not sure what to make of the performance concern—to be honest, I hope it shouldn't impact anyone, as this sounds like the kind of thing you would not want to use in a production environment.

motobob:
@kkuzmin @ddosia if we care about any performance in this "local" mode one can save the result into PD here: https://github.com/erlcloud/erlcloud/pull/488/files#diff-b6a88197b95cfa7d84db56a756107eadR347

PR #495, opened 2018-04-02
Add s3 bucket default encryption apis

arthurnovak:
Add support for `PUT`/`GET`/`DELETE` default s3 bucket encryption
@motobob @nalundgaard @danilagamma

### DIFF
--- a/src/erlcloud_s3.erl
+++ b/src/erlcloud_s3.erl
@@ -40,7 +40,10 @@
          list_bucket_inventory/1, list_bucket_inventory/2, list_bucket_inventory/3,
          get_bucket_inventory/2, get_bucket_inventory/3,
          put_bucket_inventory/2, put_bucket_inventory/3,
-         delete_bucket_inventory/2, delete_bucket_inventory/3
+         delete_bucket_inventory/2, delete_bucket_inventory/3,
+         put_bucket_encryption/2, put_bucket_encryption/3, put_bucket_encryption/4,
+         get_bucket_encryption/1, get_bucket_encryption/2,
+         delete_bucket_encryption/1, delete_bucket_encryption/2
     ]).
 
 -ifdef(TEST).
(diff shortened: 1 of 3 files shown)


nalundgaard:
IMO this should include a reference to the [API Docs](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUTencryption.html) for each action.

arthurnovak:
@nalundgaard updated

src/erlcloud_s3.erl
```diff
@@ -1616,6 +1712,28 @@ s3_xml_request2(Config, Method, Host, Path, Subresource, Params, POSTData, Heade
...
+s3_xml_request3(Config, Method, Host, Path, SubResource, Params, POSTData,
+                Headers) ->
+    case s3_request2(Config, Method, Host, Path, SubResource, Params,
+                     POSTData, Headers) of
+        {ok, {_, <<>>}} ->
+            ok;
+        {ok, {_, Body}} ->
+            {XML, _} = xmerl_scan:string(binary_to_list(Body)),
+            {ok, XML};
+        {error, {http_error, _, _, Reason}} = Error ->
```
motobob: I may guess you no like exceptions :)
however why would we introduce a third way of rerun fir these APIs?

motobob: converting `http_error` into `s3_error` is this correct?

arthurnovak: There are several reasons for the implementation:
1. Yes, i'm sure return `{error, Reason}` is better than throw an exception here
2. There was no request function, which handles empty body results and do not throw exception at the same time:
`s3_xml_request2/8` doesn't handle `{ok, {_Headers, <<>>}}`
`s3_simple_request/8` throws an exception
3. I wanted to isolate error codes returned by s3, especially `ServerSideEncryptionConfigurationNotFoundError`, instead of returning whole XML
4. Looks like all errors we receive from s3 are considered as http errors in current implementation of `erlcloud_s3`. So here is the list of s3 errors: https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#ErrorCodeList
And in case of `NoSuchBucket` error, which is s3 error I suppose, we'd receive `{error,{http_error,404,"Not Found",
                   <<"<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>NoSuchBucket</Code><Message>The specifie"...>>}}`

@motobob what do you think?

motobob: all these reasons make sense and good...
but I'm not very comfortable of letting 3d way of doing things  for just a subset of a single module. consistency for the win.
how much of a burden for you will it be to update S3 module to get option (via params or via erlcloud_aws config) to either return errors this way for all or do current exceptions. e.g. separate PR?

arthurnovak: @motobob It shouldn't be a big deal I guess. I'll plan this work.
Ok, I've updated current PR to get rid of `request3`. Take a look please

motobob: 👍

PR #500, opened 2018-04-16
Fix #155

### LINKED ISSUE #155: Misleading error message when using access keys of the wrong type
When using character lists instead of strings for the AWS credentials, a timeout error is generated.

```
11> erlcloud_s3:configure('a', 'b').
ok
12> erlcloud_s3:list_buckets().
** exception error: {aws_error,{socket_error,timeout}}
     in function  erlcloud_s3:s3_request/8 (src/erlcloud_s3.erl, line 971)
     in call from erlcloud_s3:s3_xml_request/8 (src/erlcloud_s3.erl, line 955)
     in call from erlcloud_s3:list_buckets/1 (src/erlcloud_s3.erl, line 296)
13> 
```

nanshakov:
Fix #155 issue, add guards of all modules (configure fun's)

### DIFF
--- a/src/erlcloud_autoscaling.erl
+++ b/src/erlcloud_autoscaling.erl
@@ -34,16 +34,12 @@ new(AccessKeyID, SecretAccessKey, Host) ->
                 autoscaling_host=Host}.
 
 -spec configure(string(), string()) -> ok.
-configure(AccessKeyID, SecretAccessKey)
-    when is_list(AccessKeyID), is_list(SecretAccessKey) ->
-    put(aws_config, new(AccessKeyID, SecretAccessKey)),
-    ok.
+configure(AccessKeyID, SecretAccessKey) ->
+    erlcloud_config:configure(AccessKeyID, SecretAccessKey, fun new/2).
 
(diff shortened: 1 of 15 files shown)


nanshakov:
It's may be:
`erlcloud_config`
```erlang
-spec configure(string(), string(), function()) -> ok.
configure(AccessKeyID, SecretAccessKey, New)
    when is_list(AccessKeyID), is_list(SecretAccessKey), is_function(New) ->
    put(aws_config, New(AccessKeyID, SecretAccessKey)),
    ok.
```
and using (in something module) 
```erlang
-spec new(string(), string()) -> aws_config().
new(AccessKeyID, SecretAccessKey) ->
    #aws_config{access_key_id=AccessKeyID,
                secret_access_key=SecretAccessKey}.
...
-spec configure(string(), string()) -> ok.
configure(AccessKeyID, SecretAccessKey) ->
    erlcloud_config:configure(AccessKeyID, SecretAccessKey, fun new/2).
```
or it's bad solution ?

motobob:
this is ok for new/2... what about new/3? if that work too - ship it

PR #501, opened 2018-04-18
Add the replication status in object metadata

glejeune:
Update `erlcloud_s3:get_object_metadata/4` to return the replication status given by the `x-amz-replication-status` header.

This modification add a new tuple in the response : `{replication_status, Status}` with `Status = undefined | pending | completed | failed | replica`.


glejeune:
:+1:

Issue #505, opened 2018-05-04
lambda invoke does not return error headers


motobob:
as per https://docs.aws.amazon.com/lambda/latest/dg/API_Invoke.html#API_Invoke_ResponseSyntax

execution errors are returned via `X-Amz-Function-Error`...

PR #506, opened 2018-05-07
make lambda invokes headers visible

tumdash:
`lambda:invoke` returns `{ok, {Headers, Data}}` for success Lambda calls instead of `{ok, Data}` as currently

### DIFF
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -226,7 +226,7 @@ aws_request_form(Method, Protocol, Host, Port, Path, Form, Headers, Config) ->
 aws_request_form_raw(Method, Scheme, Host, Port, Path, Form, Headers, Config) ->
     do_aws_request_form_raw(Method, Scheme, Host, Port, Path, Form, Headers, Config, false).
 
-do_aws_request_form_raw(Method, Scheme, Host, Port, Path, Form, Headers, Config, IsHdrs4Return) ->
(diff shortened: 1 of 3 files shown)


motobob:
fixes #505

src/erlcloud_aws.erl
```diff
@@ -268,15 +272,16 @@ aws_request_form_raw(Method, Scheme, Host, Port, Path, Form, Headers, Config) ->
...
 
-    case request_to_return(Response) of
-        {ok, {_, Body}} ->
-            {ok, Body};
-        {error, {Error, StatusCode, StatusLine, Body, _Headers}} ->
-            {error, {Error, StatusCode, StatusLine, Body}};
-        {error, Reason} ->
-            {error, Reason}
-    end.
+request_to_return_headers({ok, {_, Body}}, false) ->
```
motobob: only lambda is like that . no need to change anything here

src/erlcloud_lambda.erl
```diff
@@ -715,31 +715,38 @@ lambda_request(Config, Method, Path, Body) ->
...
             {error, Reason}
     end.
 
-lambda_request_no_update(Config, Method, Path, Hdrs, Body, QParam) ->
+lambda_request_no_update(Config, Method, Path, Options, Body, QParam) ->
     Form = case encode_body(Body) of
                <<>>   -> erlcloud_http:make_query_string(QParam);
                Value  -> Value
            end,
+    ShowRespHeaders = proplists:get_value(show_headers, Options, false),
```
motobob: add documentation for invokeAPI above and let's ship this

PR #507, opened 2018-05-08
Hackney default http client

nanshakov:
Set `Hackney` as default http client. Pool configuration set up in `config/sys.config` and the same as lhttpc pool configuration.


motobob:
can we make it at least of some compatibility and like in #457
e.g. may be we even use `is_loaded` type of a thing? (mindblown)

@nalundgaard preferences?

I do want to rip lhttpC but do in it a nicer way... just having erlcloud 4.0.0 is not really cool option.

motobob:
what about use case of people having several pools per library configured? 
e.g. 100 connection to EC2, 100 to IAM...etc...

nanshakov:
About several pools.
We can use several pools:
```erlang
PoolName = mypool,
Options = [{timeout, 150000}, {max_connections, 100}],
ok = hackney_pool:start_pool(PoolName, Options).
...
PoolName2= mypool2,
Options2 = [{timeout, 150000}, {max_connections, 200}],
ok = hackney_pool:start_pool(PoolName2, Options2).
```
`PoolName` and `Options` can be defined in `erlcloud_aws_hrl:aws_config` as 
```erlang
erlcloud_ddb2_pool_name = default|atom();
erlcloud_ddb2_pool_options = [{timeout, 150000}, {max_connections, 100}];
...
erlcloud_sqs_pool_name = default|atom();
erlcloud_sqs_pool_options = [{timeout, 150000}, {max_connections, 100}];
```
in this case we use default pool (config/sys.config) for all services.
We must update `new` funcs to allow set `..._pool_name` & `..._pool_options` and when we create req. - gets pool configuration. Also we must change https://github.com/erlcloud/erlcloud/blob/d61b37d8f6e62acfb01ca77216f900b25d8fb121/src/erlcloud_httpc.erl#L76-L80. 
@motobob , @nalundgaard  preferences?

motobob:
wait...supporting setting of a pool at config level is enough. as long as it's used in the erlcloud_httpC..
no need to have pool per service in the library.

so the use case is covered.

nanshakov:
@motobob Some questions about compatibility.
If i understand correctly i have to
1. Remove `hakney` of `erlcloud.app.src`
2. Update ReadMe as a [#457](https://github.com/erlcloud/erlcloud/pull/457)

And users must run http server manually and must edite `aws config`. 
What do you mean 'use is_loaded' ? We cann't use `code:is_loaded` during the erlcloud loading, because erlcloud haven't application behavior. We may use `code:is_loaded` to get current http server in each request, but it's performance hard call may be.

PR #508, opened 2018-05-10
do not retry lhttpc errors and make default ddb timeout 10s

zdaniel86:
### Problem
The `al_ddb_lib:put_item(NewObj)` returns `{error, {<<"ConditionalCheckFailedException">>, _}}` while `{ok, NewObj} = al_ddb_lib:get_item()`.

The last reading returns NewObj because initial request timeouted but eventually succeeded in DDB. erlcloud retries timeouted request and gets 400 because condition on the NewObj no longer valid.

Application gets `{error, {<<"ConditionalCheckFailedException">>, _}}` and thinks that `al_ddb_lib:put_item` request failed. This error response is misleading and makes cause analysis difficult.

Initial timeout for the first attempt is 1 second which includes http response processing time which is not enough for heavily loaded application.

### Solution
* Do not retry http client errors.
* Make single timeout value


zdaniel86:
On the second thought. We can workaround this by changing default config. https://algithub.pd.alertlogic.net/defender/sources/pull/282

PR #519, opened 2018-06-07
add per shard metrics

motobob:


src/erlcloud_kinesis.erl
```diff
@@ -243,6 +246,114 @@ describe_stream(StreamName, Limit, ExcludeShard, Config)
...
+%% '
+%%
+%% @end
+%%------------------------------------------------------------------------------
+
+-spec enable_enhanced_monitoring(binary()) -> erlcloud_kinesis_impl:json_return().
+enable_enhanced_monitoring(StreamName) ->
+    enable_enhanced_monitoring(StreamName, default_config()).
+
+-spec enable_enhanced_monitoring(binary(), list() | aws_config()) -> erlcloud_kinesis_impl:json_return().
```
kkuzmin: could you please amend the spec to relet the list of binaries?

PR #524, opened 2018-07-16
Allow create volume snapshots with tags

sanmiguel:
This allows a single atomic request to create-and-tag a snapshot/volume, rather than via consecutive calls to e.g. `CreateSnapshot` then `CreateTags`.

Note that in its current form this will always return the `{tag_set, TagList}` in the proplist result.

### DIFF
--- a/src/erlcloud_ec2.erl
+++ b/src/erlcloud_ec2.erl
@@ -814,10 +814,22 @@ create_snapshot(VolumeID, Config)
 create_snapshot(VolumeID, Description) ->
     create_snapshot(VolumeID, Description, default_config()).
 
--spec create_snapshot(string(), string(), aws_config()) -> ok_error(proplist()).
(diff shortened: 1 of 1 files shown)


nalundgaard:
Looks like a good addition, but I wonder if there isn't a "standard" way in this module to use an `Options` proplist to pass this in. Positional parameter ordering for optional parameters is (granted) apparently the way this module is implemented, but I wonder if it's worth revisiting.

My other question is, should we have unit tests for the modules updated? It looks like there are none now.

sanmiguel:
Every other function (except a few of the `describe_*` functions that take a `Selector` argument) uses this same technique. I appreciate there are probably better ways to handle optional parameters but I'd strongly argue that change falls outside the scope of this PR...

sanmiguel:
Thanks folks! Much appreciated :)

PR #526, opened 2018-07-20
Update eini to 1.2.5

nalundgaard:
erlcloud/eini#11 removed the unneeded `covertool` dep/plugin from the dependencies.

PR #527, opened 2018-07-25
Fix eini reference in rebar.lock file

nalundgaard:
this was missed in #526.


motobob:
hm. i thought i checked that. sorry my bad

PR #529, opened 2018-08-09
step functions API

tumdash:
SF API
Test


src/erlcloud_sf.erl
```diff
@@ -0,0 +1,630 @@
...
+         update_state_machine/3, update_state_machine/1]).
+
+%%------------------------------------------------------------------------------
+%% Library initialization.
+%%------------------------------------------------------------------------------
+-spec new(string(), string()) -> aws_config().
+new(AccessKeyID, SecretAccessKey) ->
+    #aws_config{access_key_id     = AccessKeyID,
+                secret_access_key = SecretAccessKey,
+                retry             = fun erlcloud_retry:default_retry/1}.
```
motobob: nope. please keep one default which is no retry

include/erlcloud_aws.hrl
```diff
@@ -55,6 +55,9 @@
           lambda_scheme="https://"::string(),
           lambda_host="lambda.us-east-1.amazonaws.com"::string(),
           lambda_port=443::non_neg_integer(),
+          sf_scheme="https://"::string(),
+          sf_host="states.us-east-1.amazonaws.com"::string(),
```
motobob: can we call all this stepfun_host or states_host...
no value to safe on chars.

src/erlcloud_sf.erl
```diff
@@ -0,0 +1,630 @@
+-module(erlcloud_sf).
```
motobob: can we please call module erlcloud_stepfun or erlcloud_states

src/erlcloud_sf.erl
```diff
@@ -0,0 +1,630 @@
...
+
+-spec create_state_machine(Definition       :: map(),
+                           StateMachineName :: binary(),
+                           RoleArn          :: binary(),
+                           Config           :: aws_config()) ->
+    {ok, map()} | {error, any()}.
+create_state_machine(Definition, StateMachineName, RoleArn, Config)
+        when is_binary(StateMachineName), is_binary(RoleArn), is_map(Definition) ->
+    Req = #{
+        <<"definition">>    => jsx:encode(Definition),
```
motobob: should encode nested maps instead of letting jsx do it later?

src/erlcloud_sf.erl
```diff
@@ -0,0 +1,630 @@
...
+%%
+%%------------------------------------------------------------------------------
+-spec list_executions(StateMachineArn :: binary()) -> {ok, map()} | {error, any()}.
+list_executions(StateMachineArn) ->
+    list_executions(StateMachineArn, #{}, default_config()).
+
+-spec list_executions(StateMachineArn   :: binary(),
+                      Options           :: map(),
+                      Config            :: aws_config()) -> {ok, map()} | {error, any()}.
+list_executions(StateMachineArn, Options, Config)
```
motobob: should there be arrity /2 functions?

test/erlcloud_sf_tests.erl
```diff
@@ -0,0 +1,302 @@
+-module(erlcloud_sf_tests).
```
motobob: rename in a similar fashion

src/erlcloud_sf.erl
tumdash: in fact that should be JSON-in-JSON, so we need to do it twice

PR #530, opened 2018-08-15
don't duplicate the Accum on deep structures

cjkjellander:
explore_dirstructure/4 har a bug in that it duplicates the Accumulator for all deep structures, if you have more than one entry in Branches.

`25> erlcloud_s3:explore_dirstructure(Bucket, ["inbox/", "sftp/"], []).
["sftp/foo/20180814.csv",
 "sftp/foo/20180813.csv",
 "inbox/f52c9p37bcq64fnf58k5trs1ugakpt5956ljm181","inbox/",
 "inbox/f52c9p37bcq64fnf58k5trs1ugakpt5956ljm181","inbox/"]
`

Note how the last two entries are duplicates. This won't show up if there is only one branch since then the Accum is always []. It will probably double everything again for each extra branch.

Here is a tiny patch to fix this bug.


motobob:
let me test it on our folder structure and i'll get back to you soon

PR #532, opened 2018-09-05
add CW logs tags api

motobob:
add CW logs tags API

### DIFF
--- a/src/erlcloud_cloudwatch_logs.erl
+++ b/src/erlcloud_cloudwatch_logs.erl
@@ -429,7 +429,7 @@ maybe_cw_request({error, _} = Error, _Action, _Params) ->
     Error.
 
 maybe_json({ok, <<>>}) ->
-    ok;
+    {ok, []};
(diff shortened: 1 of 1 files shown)


src/erlcloud_cloudwatch_logs.erl
```diff
@@ -355,6 +428,8 @@ maybe_cw_request({ok, Config}, Action, Params) ->
 maybe_cw_request({error, _} = Error, _Action, _Params) ->
     Error.
 
+maybe_json({ok, <<>>}) ->
+    ok;
```
kkuzmin: Is it actually needed?
This may result in `case_clause`s in various places, like [here](https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_cloudwatch_logs.erl#L233)

motobob: yes since jsx fails on empty responses. 
but corrected to not fail rest

motobob:
tested in real deployment and 👍

PR #534, opened 2018-09-17
A few extra functions that we needed and may prove useful for others

elbrujohalcon:


src/erlcloud_aws.erl
```diff
@@ -816,6 +816,19 @@ timestamp_to_gregorian_seconds(Timestamp) ->
     {ok, [Yr, Mo, Da, H, M, S], []} = io_lib:fread("~d-~d-~dT~d:~d:~dZ", binary_to_list(Timestamp)),
     calendar:datetime_to_gregorian_seconds({{Yr, Mo, Da}, {H, M, S}}).
 
+-spec region(aws_config()) -> {ok, string()} | {error, term()}.
```
nalundgaard: This looks handy, but it seems like a more consistent/specific name would be `get_region_from_metadata`, no?

src/erlcloud_mon.erl
```diff
@@ -447,7 +447,29 @@ configure(AccessKeyID, SecretAccessKey, Host) ->
     put(aws_config, new(AccessKeyID, SecretAccessKey, Host)),
     ok.
 
-default_config() -> erlcloud_aws:default_config().
+default_config() ->
+    case get(aws_config) of
+        undefined ->
+            DefaultConfig = erlcloud_aws:default_config(),
```
nalundgaard: It seems like you can just use erlcloud_aws:service_config/3 here and delete most of this code?

src/erlcloud_sns.erl
```diff
@@ -52,6 +53,8 @@
          get_notification_attribute/2]).
 -export([new/2, new/3, configure/2, configure/3]).
 
+-export([set_sns_host/1]).
```
nalundgaard: This seems to be exported 2 times.

src/erlcloud_sns.erl
```diff
@@ -730,6 +737,10 @@ new_config(AccessKeyID, SecretAccessKey) ->
        secret_access_key=SecretAccessKey
       }.
 
+set_sns_host(Host) ->
```
nalundgaard: This looks like it's here 2 times? Does this compile?

src/erlcloud_as.erl
```diff
@@ -8,11 +8,13 @@
 %% AWS Autoscaling functions
 -export([describe_groups/0, describe_groups/1, describe_groups/2, describe_groups/4,
          set_desired_capacity/2, set_desired_capacity/3, set_desired_capacity/4,
-
-         describe_launch_configs/0, describe_launch_configs/1, describe_launch_configs/2, 
+         set_as_host/1,
```
motobob: see erlcloud_autoscalling module too - seems half of it is already there.
i know it's strange we have that ddb vs ddb2 and _as vs _autoscalling but we just never got to deprecating things...

src/erlcloud_as.erl
```diff
@@ -78,6 +82,20 @@
...
+
+-define(RECORD_LIFECYCLE_ACTION_HEARTBEAT_RESPONSE,
+    "/RecordLifecycleActionHeartbeatResponse/ResponseMetadata/RequestId").
+
+
+%% --------------------------------------------------------------------
+%% @doc Sets as hosts in #aws_config{}
+%% @end
+%% --------------------------------------------------------------------
+set_as_host(Host) ->
```
motobob: should not this come by magic from erlcloud_aws with service config?
unless you set this for "MyLocalASHost"...

src/erlcloud_as.erl
```diff
@@ -78,6 +82,20 @@
...
+    Config = erlcloud_aws:default_config(),
+    put(aws_config, Config#aws_config{as_host = Host}).
```
motobob: better to pass it once during config and always pass the aws_config{} for all calls.
proc dict method only work for a single basic app inside an application running on ec2

src/erlcloud_as.erl
```diff
@@ -695,15 +713,75 @@ complete_lifecycle_action(GroupName, LifecycleActionResult, LifecycleHookName, I
...
+complete_lifecycle_action(GroupName, HookName, ActionToken, LifecycleActionResult) ->
+    Config = erlcloud_aws:default_config(),
+    complete_lifecycle_action(GroupName, HookName, ActionToken, LifecycleActionResult, Config).
+
+-spec complete_lifecycle_action(string(), string(), string(), atom(), aws_config()) ->
+    {ok, string()} | {error, term()}.
+complete_lifecycle_action(GroupName, HookName, ActionToken, LifecycleActionResult, Config) ->
+    LifecycleActionResultIn = case LifecycleActionResult of
+                                  continue -> "CONTINUE";
+                                  abandom -> "ABANDON"
```
motobob: typo

src/erlcloud_aws.erl
```diff
@@ -73,23 +74,23 @@
 
 aws_request_xml(Method, Host, Path, Params, #aws_config{} = Config) ->
     Body = aws_request(Method, Host, Path, Params, Config),
-    element(1, xmerl_scan:string(binary_to_list(Body))).
+    element(1, xmerl_scan:string(Body)).
```
motobob: why? this seems a breaking change...

src/erlcloud_as.erl
motobob: secondly, i actually do encourage to switch to autoscalling to switch to SignV4 methods and dump SingV2. only SimpleDb will remain on SignV2, rest will die.

src/erlcloud_aws.erl
motobob: yes, please follow the module naming.

src/erlcloud_aws.erl
```diff
@@ -937,7 +950,7 @@ http_body(Return) ->
 http_headers_body({ok, {{OKStatus, _StatusLine}, Headers, Body}})
   when OKStatus >= 200, OKStatus =< 299 ->
     {ok, {Headers, Body}};
-http_headers_body({ok, {{Status, StatusLine}, _Headers, Body}}) ->
+http_headers_body({ok, {{_Version, Status, StatusLine}, _Headers, Body}}) ->
```
motobob: what is the purpose of this change?

src/erlcloud_ddb_impl.erl
```diff
@@ -185,7 +185,7 @@ request_and_retry(Config, Headers, Body, {attempt, Attempt}) ->
     RetryFun = retry_fun(Config),
     case erlcloud_httpc:request(
            url(Config), post,
-           [{<<"content-type">>, <<"application/x-amz-json-1.0">>} | Headers],
+           Headers, <<"application/x-amz-json-1.0">>,
```
motobob: breaking change?

src/erlcloud_elb.erl
```diff
@@ -144,6 +146,14 @@ configure_health_check(LB, Target, Config) when is_list(LB) ->
                        [{"LoadBalancerName", [LB]},
                         {"HealthCheck.Target", Target}]).
 
+describe_instance_health(LB) ->
```
motobob: API incomplete https://docs.aws.amazon.com/elasticloadbalancing/2012-06-01/APIReference/API_DescribeInstanceHealth.html
need instance Ids list. 
please add

src/erlcloud_mon.erl
```diff
@@ -447,7 +447,29 @@ configure(AccessKeyID, SecretAccessKey, Host) ->
     put(aws_config, new(AccessKeyID, SecretAccessKey, Host)),
     ok.
 
-default_config() -> erlcloud_aws:default_config().
+default_config() ->
```
motobob: completely unnecessary. see erlcloud_aws for regional configurations.

src/erlcloud_mon.erl
```diff
@@ -447,7 +447,29 @@ configure(AccessKeyID, SecretAccessKey, Host) ->
...
+            Config;
+        Config ->
+            Config
+    end.
+
+%%%===================================================================
+%%% Internal Functions
+%%%===================================================================
+
+endpoint("us-east-1")      -> "monitoring.us-east-1.amazonaws.com";
```
motobob: there are way more regions now...

elbrujohalcon:
Now that I cleaned it up enough I can see, thanks to your comments @motobob 🙏 , how we could've implemented what we need in a different way.
I'll close this PR and, if we get to the point where we can implement a proper version of `describe_instance_health`, I'll create a new PR.
Thanks for all the help, folks! You rock 👨‍🎤

posilva:
@elbrujohalcon do you think that is possible to add also set_sns_port, it's something I was intended to PR soonish because we are trying to use LocalStack that requires the service port to be handled.

I notice that you made the original PR for SNS support, but  without allow to change the default port of the endpoint.

Cheers

Pedro

elbrujohalcon:
No idea, @posilva … that implementation was done ages ago. 👴

posilva:
@elbrujohalcon you can check in the 'aws_config' record we do not have sns_port that would allow to change in certain cases, like using LocalStack :) 

https://github.com/erlcloud/erlcloud/blob/master/include/erlcloud_aws.hrl#L40
According to my git blame it was you that added this sns_host, sns_scheme
Cheers

elbrujohalcon:
@posilva I understand it was me and I understand the field is not there. I'm just saying that I have no recollection at this time if I didn't include it there for a reason or just because I didn't need it at that time. No idea, at all.

PR #535, opened 2018-09-18
Die noisily when unable to parse credentials

mrallen1:
When you have a credentials file and it is unparseable for some reason, the function should die noisily because those files are used for local development work and it's a mysterious failure when your profiles are not parsed and the system tries to get information from the AWS metadata server which of course fails and you get back `undefined` as the result of this function call.


mrallen1:
Thanks @motobob

PR #547, opened 2018-10-15
Introduce ecs placementStrategy option

jasonletbetter:
The net effect here is to add support for the placementStrategy option with run_task.  It is backward compatible in that not specifying the option has no effect.

### DIFF
--- a/include/erlcloud_ecs.hrl
+++ b/include/erlcloud_ecs.hrl
@@ -153,6 +153,11 @@
     source_volume :: undefined | binary()
 }).
 
+-record(ecs_placement_strategy, {
+    field :: undefined | binary(),
+    type :: undefined | binary()
+}).
+
 -record(ecs_port_mapping, {
     container_port :: undefined | pos_integer(),
     host_port :: undefined | pos_integer(),
@@ -188,6 +193,7 @@
     memory :: undefined | pos_integer(),
     memory_reservation :: undefined | pos_integer(),
     mount_points :: undefined | [#ecs_mount_point{}],
+    placement_strategy :: undefined | [#ecs_placement_strategy{}],
     name :: undefined | binary(),
     port_mappings :: undefined | [#ecs_port_mapping{}],
     privileged :: undefined | boolean(),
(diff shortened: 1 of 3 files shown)


src/erlcloud_ecs.erl
```diff
@@ -163,7 +163,10 @@
     volume_name_opt/0,
     volumes/0,
     volumes_opt/0,
-    volumes_opts/0
+    volumes_opts/0,
+    placement_strategy/0,
```
nalundgaard: It looks like these are exported alphabetically. Any reason not to put them in accordingly?

include/erlcloud_ecs.hrl
```diff
@@ -188,6 +193,7 @@
     memory :: undefined | pos_integer(),
     memory_reservation :: undefined | pos_integer(),
     mount_points :: undefined | [#ecs_mount_point{}],
+    placement_strategy :: undefined | [#ecs_placement_strategy{}],
```
nalundgaard: should this go after `name`?

src/erlcloud_ecs.erl
```diff
@@ -492,7 +516,7 @@ container_volumes_from_opt() ->
                                     container_log_configuration_opt() |
                                     {memory, pos_integer()} |
                                     {memory_reservation, pos_integer()} |
-                                    container_mount_points_opt() |
+                                    container_mount_points_opt() |
```
nalundgaard: 👾

include/erlcloud_ecs.hrl
nalundgaard: where is the code to actually set this to something other than default? I don't see it in this PR.

jasonletbetter: It's a [run_task option](https://github.com/erlcloud/erlcloud/pull/547/files#diff-67e666f62a38974173c59f4bcd66bb1dR2175)

src/erlcloud_ecs.erl
jasonletbetter: sure

include/erlcloud_ecs.hrl
jasonletbetter: actually, this is left over from a road not taken; deleting instead

PR #553, opened 2018-11-06
Allow binary payload in lambda invoke request

egobrain:


src/erlcloud_lambda.erl
```diff
@@ -761,6 +761,8 @@ decode_body(<<>>) ->
 decode_body(BinData) ->
     jsx:decode(BinData).
```
motobob: seem to me having decode removed under an option is needed to, no?

egobrain: Fixed.  `raw_response_body` option is added

PR #554, opened 2018-11-07
Fix raw_response_body option

egobrain:
Sorry for bug in previous pr.
Here is fix for:

```
 {error,{socket_error,{function_clause,[{lhttpc_lib,canonical_header,[raw_response_body],[{file,"
```

and unit test for it.


motobob:
test it

PR #561, opened 2018-11-30
update jsx to 2.9.0

getong:


motobob:
great... i think i'll remove <19.3 actually in a separate PR

PR #564, opened 2018-12-01
Fix dialyzer

getong:
Still some dialyzer not fix, see erlang/otp#2038.
You can cherry-pick the first commit and drop the second commit.


motobob:
that's fine - we will use both.

nalundgaard:
It seems like there is no main PLT file on travis. Should we add a step to build that? It sounds like it would add a lot of time to the runs.

PR #565, opened 2018-12-03
add list_tables_all and wait_for_table_active functions into erlcloud_ddb_util

pzLib:
1.  erlcloud_ddb2:list_all_tables is a helper function of list_tables.
2. erlcloud_ddb2:wait_for_table will wait until the table_status=active.

Thank you.

### DIFF
--- a/src/erlcloud_ddb2.erl
+++ b/src/erlcloud_ddb2.erl
@@ -2583,6 +2583,7 @@ list_tables(Opts, Config) ->
 list_all_tables() ->
     list_all_tables(default_config()).
 
+-spec list_all_tables(aws_config()) -> {ok, [table_name()]} | {error, any()}.
 list_all_tables(AWSCfg) ->
     do_list_all_tables(undefined, AWSCfg, []).
 
@@ -3575,6 +3576,7 @@ update_time_to_live(Table, AttributeName, Enabled) ->
 %% '''
 %% @end
 %%------------------------------------------------------------------------------
+-spec wait_for_table(table_name(), non_neg_integer(), aws_config()) -> ok | {error, any()}.
 wait_for_table(Table, Interval, AWSCfg) when is_binary(Table) ->
     case erlcloud_ddb2:describe_table(Table, [{out, record}], AWSCfg) of
         {ok, #ddb2_describe_table{table = #ddb2_table_description{table_status = active}}} ->
@@ -3583,16 +3585,16 @@ wait_for_table(Table, Interval, AWSCfg) when is_binary(Table) ->
             timer:sleep(Interval),
             wait_for_table(Table, Interval, AWSCfg);
         {error, Reason} ->
-            error(Reason)
+            {error, Reason}
     end;
 wait_for_table(Table, Interval, AWSCfg) when is_list(Table) ->
     wait_for_table(list_to_binary(Table), Interval, AWSCfg).
 
 wait_for_table(Table, AWSCfg) ->
-    wait_for_table(list_to_binary(Table), 3000, AWSCfg).
+    wait_for_table(Table, 3000, AWSCfg).
 
 wait_for_table(Table) ->
-    wait_for_table(list_to_binary(Table), default_config()).
+    wait_for_table(Table, default_config()).
 
 
 to_binary(X) when is_binary(X) ->
--- a/src/erlcloud_ddb2.erl
+++ b/src/erlcloud_ddb2.erl
@@ -105,7 +105,6 @@
          list_backups/0, list_backups/1, list_backups/2,
          list_global_tables/0, list_global_tables/1, list_global_tables/2,
          list_tables/0, list_tables/1, list_tables/2,
-         list_all_tables/0, list_all_tables/1,
          list_tags_of_resource/1, list_tags_of_resource/2, list_tags_of_resource/3,
          put_item/2, put_item/3, put_item/4,
          %% Note that query is a Erlang reserved word, so we use q instead
@@ -119,8 +118,7 @@
          update_item/3, update_item/4, update_item/5,
          update_global_table/2, update_global_table/3, update_global_table/4,
          update_table/2, update_table/3, update_table/4, update_table/5,
-         update_time_to_live/2, update_time_to_live/3, update_time_to_live/4,
-         wait_for_table/1, wait_for_table/2, wait_for_table/3
+         update_time_to_live/2, update_time_to_live/3, update_time_to_live/4
         ]).
 
 -export_type(
@@ -2579,28 +2577,6 @@ list_tables(Opts, Config) ->
     out(Return, fun(Json, UOpts) -> undynamize_record(list_tables_record(), Json, UOpts) end, 
         DdbOpts, #ddb2_list_tables.table_names, {ok, []}).
 
-
-list_all_tables() ->
-    list_all_tables(default_config()).
-
--spec list_all_tables(aws_config()) -> {ok, [table_name()]} | {error, any()}.
-list_all_tables(AWSCfg) ->
-    do_list_all_tables(undefined, AWSCfg, []).
-
-do_list_all_tables(LastTable, AWSCfg, Result) ->
-    Options = [{exclusive_start_table_name, LastTable}, {out, record}],
-    case erlcloud_ddb2:list_tables(Options, AWSCfg) of
-        {ok, #ddb2_list_tables{table_names = TableNames, last_evaluated_table_name = LastTableName}} ->
-            NewResult = TableNames ++ Result,
-            case LastTableName of
-                undefined -> {ok, NewResult};
-                _ -> do_list_all_tables(LastTableName, AWSCfg, NewResult)
-            end;
-        {error, _} = Error ->
-            Error
-    end.
-
-
 %%%------------------------------------------------------------------------------
 %%% ListTagsOfResource
 %%%------------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


src/erlcloud_ddb2.erl
```diff
@@ -2577,6 +2579,28 @@ list_tables(Opts, Config) ->
...
+
+-spec list_all_tables(aws_config()) -> {ok, [table_name()]} | {error, any()}.
+list_all_tables(AWSCfg) ->
+    do_list_all_tables(undefined, AWSCfg, []).
+
+do_list_all_tables(LastTable, AWSCfg, Result) ->
+    Options = [{exclusive_start_table_name, LastTable}, {out, record}],
+    case erlcloud_ddb2:list_tables(Options, AWSCfg) of
+        {ok, #ddb2_list_tables{table_names = TableNames, last_evaluated_table_name = LastTableName}} ->
+            NewResult = TableNames ++ Result,
```
nalundgaard: This is probably a good use case for [flatreverse](https://github.com/erlcloud/erlcloud/blob/895fb3b3a0891ae21b56fd8c8029c8dd7a7dc611/src/erlcloud_ddb_util.erl#L505-L507) in [erlcloud_ddb_util](https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_ddb_util.erl).

pzLib:
Thanks for your reply.

src/erlcloud_ddb2.erl
```diff
@@ -3540,6 +3564,39 @@ update_time_to_live(Table, Opts, Config) when is_list(Opts) ->
...
+%% @end
+%%------------------------------------------------------------------------------
+-spec wait_for_table(table_name(), non_neg_integer(), aws_config()) -> ok | {error, any()}.
+wait_for_table(Table, Interval, AWSCfg) when is_binary(Table) ->
+    case erlcloud_ddb2:describe_table(Table, [{out, record}], AWSCfg) of
+        {ok, #ddb2_describe_table{table = #ddb2_table_description{table_status = active}}} ->
+            ok;
+        {ok, _} ->
+            timer:sleep(Interval),
+            wait_for_table(Table, Interval, AWSCfg);
```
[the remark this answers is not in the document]
pzLib: Yes, a max tries threshold is a good idea.

src/erlcloud_ddb2.erl
pzLib: Sorry, I can't understand your mean. Result and TableNames are binary list, I think Result ++ TableNames could keep the order but why flatreverse is better than ++?

Thank you.

src/erlcloud_ddb2.erl
```diff
@@ -2577,6 +2579,28 @@ list_tables(Opts, Config) ->
     out(Return, fun(Json, UOpts) -> undynamize_record(list_tables_record(), Json, UOpts) end, 
         DdbOpts, #ddb2_list_tables.table_names, {ok, []}).
 
+
+list_all_tables() ->
```
nalundgaard: I should also note that I think this should be called `list_tables_all` to be in keeping with the overall style.

src/erlcloud_ddb2.erl
nalundgaard: The main reason is that flatreverse (see how it's used with [q_all](https://github.com/erlcloud/erlcloud/blob/895fb3b3a0891ae21b56fd8c8029c8dd7a7dc611/src/erlcloud_ddb_util.erl#L338) and [scan_all](https://github.com/erlcloud/erlcloud/blob/895fb3b3a0891ae21b56fd8c8029c8dd7a7dc611/src/erlcloud_ddb_util.erl#L401)) actually returns the tables in the order that they are returned by DynamoDB API, where as the above append strategy, while pretty efficient, appends the lists backwards (the last set of table names is at the beginning of the list). Flatreverse is as efficient as the above function but the order is symmetric to an in-order traversal of the table list from DynamoDB.

pzLib: I known the order problem, so I will use Result ++ TableNames to keep the order, but I don't agree with "efficient", in fact, flatreverse can't be used in here:

```
Tables = [<<"Tab1">>, <<"Tab2">>] ,  
Result = [<<"Tab3">>, <<"Tab4">>] ,  
erlcloud_ddb_util:flatreverse([Tables | Result]).  
```

output:

```
** exception error: bad argument  
     in operator  ++/2  
        called as <<"Tab3">> ++ [<<"Tab1">>,<<"Tab2">>]  
...  
```

English is not my first language, please ignore my English mistakes.  : )
Thank you.

nalundgaard: You can indeed switch the order using `Result ++ TableNames`, but this is comparatively inefficient because every element in `Result` ends up getting copied at every step of the function. This probably isn't terrible in most cases, but it's generally something to avoid, because it's the accumulated list and it can grow very large. It is best to append the small list into the larger one that is being accumulated.

I think you have a minor issue with your use above. Results should be accumulated as a list of lists, and then "reverse-flattened" at the end:

```erlang
T1 = [<<"table1">>, <<"table2">>],
T2 = [<<"table3">>, <<"table4">>],
T3 = [<<"table5">>, <<"table6">>],
erlcloud_ddb_util:flatreverse([T3, T2, T1]).
% [<<"table1">>,<<"table2">>,<<"table3">>,<<"table4">>,
%  <<"table5">>,<<"table6">>]
```

So, in your `list_tables_all` function, you would need to build the results as `[Tables | Result]` as you have it, but note that in the initial case, Result is `[]`, so you get `[[<<"Tab3">>, <<"Tab4">>]|[]]`, which is `[[<<"Tab3">>, <<"Tab4">>]]`. As you build from there, you get the backwards list of lists of table names that you then `flatreverse` at the end; this uses much less redundant list copying. The problem in your example above is just that you haven't expressed the `Result` variable correctly. If you implement the function as it is for `scan_all`,  it will just work in practice.

src/erlcloud_ddb2.erl
```diff
@@ -3540,6 +3564,39 @@ update_time_to_live(Table, Opts, Config) when is_list(Opts) ->
...
+%%
+%% ```
+%% erlcloud_ddb2:wait_for_table(<<"TableName">>, 3000, Config)
+%% '''
+%% @end
+%%------------------------------------------------------------------------------
+-spec wait_for_table(table_name(), non_neg_integer(), aws_config()) -> ok | {error, any()}.
+wait_for_table(Table, Interval, AWSCfg) when is_binary(Table) ->
+    case erlcloud_ddb2:describe_table(Table, [{out, record}], AWSCfg) of
+        {ok, #ddb2_describe_table{table = #ddb2_table_description{table_status = active}}} ->
```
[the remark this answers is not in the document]
nalundgaard: On further reflection, I think that actually parameterizing `active` here is a bad idea, but I think best to call this `wait_for_table_active`. Sorry for any confusion.

src/erlcloud_ddb2.erl
pzLib: Thanks for you tell me the correct way to use flatreverse.

I agree :
`It is best to append the small list into the larger one that is being accumulated.`, 
many Erlang books talk about this.

Normally, most user's tables less than one hundred or less than one thousand in a region, this is a simple test:

```erlang
flatreverse(Result, 0) ->
    Result;
flatreverse(Result, Max) ->
    List = erlcloud_ddb_util:flatreverse([[1,2,3,4,5,6,7,8,9,0], Result]),
    flatreverse(List, Max-1).


plus(Result, 0) ->
    Result;
plus(Result, Max) ->
    plus(Result ++ [1,2,3,4,5,6,7,8,9,0], Max-1).

tc(Max) ->
    {Time1, L1} = timer:tc(?MODULE, flatreverse, [[], Max]),
    {Time2, L2} = timer:tc(?MODULE, plus, [[], Max]),
    io:format("~p, ~p, ~p, ~p~n", [length(L1), length(L2), Time1, Time2]).
```  

output:

```
1> test:tc(1).
10, 10, 2, 0
ok
2> test:tc(50).
500, 500, 3035, 56
ok
3> test:tc(100).
1000, 1000, 372, 259
ok
4> test:tc(500).
5000, 5000, 9181, 9481
ok
5> test:tc(1000).
10000, 10000, 31010, 21884
ok
6> test:tc(5000).
50000, 50000, 502264, 448474
```

So I think `This probably isn't terrible in most cases` is enough, ++ and flatreverse both fine but ++ is more simple in this case.

Finally, I would like accept your suggestion and use flatreverse because the code lib should keep the same style and more safe than ++.

Thanks again. You are a nice man and good reviewer.

src/erlcloud_ddb2.erl
pzLib: OK! : )

src/erlcloud_ddb2.erl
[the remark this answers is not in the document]
pzLib: You mean I should change this function name to `wait_for_table_active`?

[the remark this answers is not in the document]
nalundgaard: Yes, I think that is more clear for caller what the function does.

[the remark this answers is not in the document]
pzLib: OK!

src/erlcloud_ddb_util.erl
```diff
@@ -477,6 +501,52 @@ batch_write_retry(RequestItems, Config) ->
...
+            ok;
+        {ok, #ddb2_describe_table{table = #ddb2_table_description{table_status = deleting}}} ->
+            {error, deleting};
+        {ok, _} ->
+            case RetryTimes of
+                0 ->
+                    timer:sleep(Interval),
+                    wait_for_table_active(Table, Interval, RetryTimes, Config);
+                1 ->
+                    {error, timeout};
```
nalundgaard: IMO this should just be a function head of wait_for_table_active, e.g.:

```erlang
wait_for_table_active(_Table, _Interval, RetryTimes, _Config) when RetryTimes =< 0 ->
    {error, retry_threshold_exceeded};
wait_for_table_active(Table, Interval, RetryTimes, Config) ->
    ...
        wait_for_table_active(Table, Interval, next_retry(RetryTimes), Config),
    ...

next_retry(infinity) -> infinity;
next_retry(RetryTimes) when is_integer(RetryTimes) -> RetryTimes -1.
```

src/erlcloud_ddb_util.erl
```diff
@@ -477,6 +501,52 @@ batch_write_retry(RequestItems, Config) ->
...
+            end;
+        {error, Reason} ->
+            {error, Reason}
+    end.
+
+wait_for_table_active(Table, Interval, RetryTimes) ->
+    wait_for_table_active(Table, Interval, RetryTimes, default_config()).
+
+wait_for_table_active(Table, AWSCfg) ->
+    wait_for_table_active(Table, 3000, 0, AWSCfg).
```
nalundgaard: IMO infinity is not a good default here.

src/erlcloud_ddb_util.erl
pzLib: I think this is a very simple logic, too many functions and codes make it difficult to read.

src/erlcloud_ddb_util.erl
pzLib: I don't think so but I accept your suggestion.

src/erlcloud_ddb_util.erl
```diff
@@ -477,6 +501,52 @@ batch_write_retry(RequestItems, Config) ->
             batch_write_retry(Unprocessed, Config)
     end.
 
+%%------------------------------------------------------------------------------
+%% @doc
+%%  wait until table_status==active.
+%%  RetryTimes = 0 means infinity.
```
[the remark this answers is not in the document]
pzLib: I don't like different data type in one parameter but I accept it.

nalundgaard:
I meant to merge this a long time ago, so sorry I forgot about it!

PR #576, opened 2019-01-14
Enable passing a function for put_records encoding

mrallen1:
## Problem ##
`put_records` assumes that you always want to use base64:encode/1 to encode records.  This is not a good assumption.

## Solution ##
Enable callers to pass their own encoding function. Add `put_records/4` to enable callers to pass their own configuration as well as an encoding function.

### DIFF
--- a/src/erlcloud_kinesis.erl
+++ b/src/erlcloud_kinesis.erl
@@ -609,21 +609,17 @@ put_records(StreamName, Items) ->
     put_records(StreamName, Items, default_config()).
 
 -spec put_records(binary(), put_records_items(), function() | aws_config()) -> erlcloud_kinesis_impl:json_return().
-%% @doc This function takes a fun which is responsible for encoding each of the
-%% data items in the Items list.
-put_records(StreamName, Items, EncodingFun) when is_function(EncodingFun) ->
-    Operation = put_records_operation(),
-    Json = prepare_put_records_data(StreamName, Items, EncodingFun),
-    erlcloud_kinesis_impl:request(default_config(), Operation, Json);
+
+put_records(StreamName, Items, EncodingFun) when is_function(EncodingFun, 1) ->
+    put_records(StreamName, Items, EncodingFun, default_config());
 
 put_records(StreamName, Items, Config) ->
-    Operation = put_records_operation(),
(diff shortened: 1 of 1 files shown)


src/erlcloud_kinesis.erl
```diff
@@ -608,28 +608,42 @@ put_record(StreamName, PartitionKey, Data, ExplicitHashKey, Ordering, Options, C
 put_records(StreamName, Items) ->
     put_records(StreamName, Items, default_config()).
 
--spec put_records(binary(), put_records_items(), Config) -> erlcloud_kinesis_impl:json_return() when
-      Config :: aws_config().
+-spec put_records(binary(), put_records_items(), function() | aws_config()) -> erlcloud_kinesis_impl:json_return().
+%% @doc This function takes a fun which is responsible for encoding each of the
+%% data items in the Items list.
+put_records(StreamName, Items, EncodingFun) when is_function(EncodingFun) ->
```
motobob: passing custom function from outside is fine but stylish comments:
need to export the function spec and probably worth adding arity guard too.

src/erlcloud_kinesis.erl
```diff
@@ -608,28 +608,42 @@ put_record(StreamName, PartitionKey, Data, ExplicitHashKey, Ordering, Options, C
...
+    Operation = put_records_operation(),
+    Json = prepare_put_records_data(StreamName, Items, EncodingFun),
+    erlcloud_kinesis_impl:request(default_config(), Operation, Json);
```
motobob: I would suggest to follow library style passing from /3 to /4 fun so that this is only one call to _impl module

src/erlcloud_kinesis.erl
nalundgaard: +1

src/erlcloud_kinesis.erl
nalundgaard: +1

nalundgaard:
100% agree with @motobob here. The interface looks good once the comments are addressed.

mrallen1:
I can haz merge plz?

motobob:
sorry for delay. missed the notification

PR #579, opened 2019-02-25
Fix function clause in mk_send_batch_entry when message body is binary

KirillDavydov:
According to the spec `mk_send_batch_entry` accepts `batch_entry()` as 2-nd param.
`batch_entry()` is a tuple where 1-st element is message id and 2-nd element is message body.
Both have type `string()` which can be list or binary.

Function is fixed to accept both lists and binaries.

### DIFF
--- a/src/erlcloud_sqs.erl
+++ b/src/erlcloud_sqs.erl
@@ -588,6 +588,12 @@ mk_send_batch_entry(N, {MessageId, MessageBody}, DelaySeconds) ->
 mk_send_batch_entry(N, {MessageId, MessageBody, MessageAttributes}, DelaySeconds) ->
     mk_send_batch_entry(N, {MessageId, MessageBody, MessageAttributes, []}, DelaySeconds);
 mk_send_batch_entry(N, {MessageId, MessageBody, MessageAttributes, Opts}, DelaySeconds)
+  when is_binary(MessageId) ->
+    mk_send_batch_entry(N, {binary_to_list(MessageId), MessageBody, MessageAttributes, Opts}, DelaySeconds);
+mk_send_batch_entry(N, {MessageId, MessageBody, MessageAttributes, Opts}, DelaySeconds)
+    when is_binary(MessageBody) ->
+    mk_send_batch_entry(N, {MessageId, binary_to_list(MessageBody), MessageAttributes, Opts}, DelaySeconds);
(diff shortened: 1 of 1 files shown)


posilva:
Do we want this behaviour? We are enforcing in several places the use of `string()` as a list:
for example:
`
%% if you try this with string() as lists:
C = erlcloud_s3:new("A_VALID_ACCESS_ID","A_VALID_SECRET_KEY","s3.us-west-2.amazonaws.com").
erlcloud_s3:list_buckets(C).
[{owner,[{id,"353fb...."},
         {display_name,"some_user"}]},
 {buckets,[[{name,"my_bucket"},
            {creation_date,{{2013,3,13},{9,59,9}}}],
]}]

% but if you try with string() as binary:
C = erlcloud_s3:new(<<"A_VALID_ACCESS_ID">>,<<"A_VALID_SECRET_KEY">>,<<"s3.us-west-2.amazonaws.com">>).
erlcloud_s3:list_buckets(C).
%% you will get: 
** exception error: {aws_error,{socket_error,{ehostunreach,[{lhttpc_client,send_request,

`
for the sake of consistency(if we can call it like that) it's up to the consumer app to call with string() as list() and if you pass binary() it will return error (I know that some times is really difficult to associate the error with a 'binary' use instead of 'list')

src/erlcloud_sqs.erl
```diff
@@ -587,6 +587,12 @@ mk_send_batch_entry(N, {MessageId, MessageBody}, DelaySeconds) ->
     mk_send_batch_entry(N, {MessageId, MessageBody, [], []}, DelaySeconds);
 mk_send_batch_entry(N, {MessageId, MessageBody, MessageAttributes}, DelaySeconds) ->
     mk_send_batch_entry(N, {MessageId, MessageBody, MessageAttributes, []}, DelaySeconds);
+mk_send_batch_entry(N, {MessageId, MessageBody, MessageAttributes, Opts}, DelaySeconds)
+  when is_binary(MessageId) ->
+    mk_send_batch_entry(N, {binary_to_list(MessageId), MessageBody, MessageAttributes, Opts}, DelaySeconds);
+mk_send_batch_entry(N, {MessageId, MessageBody, MessageAttributes, Opts}, DelaySeconds)
+    when is_binary(MessageBody) ->
```
nalundgaard: why the inconsistent indentation here?

KirillDavydov:
Yes. This will complicate code more than solve issue.

PR #580, opened 2019-02-28
initial guardduty support

KingBrewer:
GuardDuty Support


src/erlcloud_guardduty.erl
```diff
@@ -0,0 +1,144 @@
...
+
+
+%%%------------------------------------------------------------------------------
+%%% Internal Functions
+%%%------------------------------------------------------------------------------
+
+filter_undef(List) ->
+    lists:filter(fun({_Name, Value}) -> Value =/= undefined end, List).
+
+base_path() ->
```
motobob: ?do we really need this?

src/erlcloud_guardduty.erl
```diff
@@ -0,0 +1,144 @@
...
+                            {"MaxItems", MaxItems}]),
+    guardduty_request(Config, get, Path, undefined, QParams).
+
+
+%%%------------------------------------------------------------------------------
+%%% Internal Functions
+%%%------------------------------------------------------------------------------
+
+filter_undef(List) ->
+    lists:filter(fun({_Name, Value}) -> Value =/= undefined end, List).
```
motobob: i thought there was a common one for this

KingBrewer: Nope. The one you are referring to has apparently been declared within `erlcloud_lambda` module. I'll move it to the `erlcloud_util` one and import it from both modules respectively.

src/erlcloud_guardduty.erl
KingBrewer: Of course we don't. I'm fixing it.

KingBrewer:
@motobob I addressed your comments

PR #585, opened 2019-03-04
erlcloud_ddb_streams: add approximate_creation_date_time to #ddb_streams_stream_record{}

nalundgaard:
Adds [ApproximateCreationDateTime](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_streams_StreamRecord.html#DDB-Type-streams_StreamRecord-ApproximateCreationDateTime) field to `erlcloud_ddb_streams:get_records` function output.


motobob:
what could possibly go wrong

nalundgaard:
@motobob I spent a pretty inordinate amount of time setting up a live test to verify that I had it pulling the data out as expected. I love that this timestamp is in a different format than their usual one.

PR #588, opened 2019-03-29
Export list_multipart_uploads/4, fixes #587

### LINKED ISSUE #587: Impossible to call erlcloud_s3:list_multipart_uploads with custom config/headers
Internally there is `list_multipart_uploads` which takes 1, 2 or 4 arguments, however the 4 argument version is not exported. This makes it impossible to for example change any of the `s3_*` settings from default values, which is a bit of a problem if you connect to s3 using a port/scheme/host that is not default.

helanhalvan:
Exports `list_multipart_uploads/4`, which allows passing a `#aws_config` record to `list_multipart_uploads`. Fixes #587


nalundgaard:
Thanks for your contribution!

PR #589, opened 2019-04-08
Retry lambda call in case of x-amz-function-error exists

egobrain:
AWS lambda may return 200 and error headers instead of 5xx (for example in case of OOM or internal crash). Some times it's necessary to retry such requests.

see https://docs.aws.amazon.com/lambda/latest/dg/API_Invoke.html

### DIFF
--- a/include/erlcloud_aws.hrl
+++ b/include/erlcloud_aws.hrl
@@ -137,6 +137,11 @@
           %% If you provide a custom function be aware of this anticipated change.
           %% See erlcloud_retry for full documentation.
           retry=fun erlcloud_retry:no_retry/1::erlcloud_retry:retry_fun(),
+
+          %% By default treat all non 2xx http error codes as errors.
+          %% But in some cases, like lambda call it useful to override such
+          %% behaviour by custom one.
+          response_type=fun erlcloud_retry:only_http_errors/1::erlcloud_retry:response_type_fun(),
           %% Currently matches DynamoDB retry
           %% It's likely this is too many retries for other services
           retry_num=10::non_neg_integer(),
--- a/include/erlcloud_aws.hrl
+++ b/include/erlcloud_aws.hrl
@@ -141,7 +141,7 @@
           %% By default treat all non 2xx http error codes as errors.
           %% But in some cases, like lambda call it useful to override such
           %% behaviour by custom one.
-          response_type=fun erlcloud_retry:only_http_errors/1::erlcloud_retry:response_type_fun(),
+          retry_response_type=fun erlcloud_retry:only_http_errors/1::erlcloud_retry:response_type_fun(),
           %% Currently matches DynamoDB retry
           %% It's likely this is too many retries for other services
           retry_num=10::non_neg_integer(),
--- a/src/erlcloud_retry.erl
+++ b/src/erlcloud_retry.erl
@@ -72,7 +72,7 @@ request_and_retry(Config, ResultFun, {retry, Request}, MaxAttempts) ->
         {ok, {{Status, StatusLine}, ResponseHeaders, ResponseBody}} ->
             ResponseType = case Status >= 200 andalso Status < 300 of
                 true ->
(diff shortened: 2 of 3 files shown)


src/erlcloud_retry.erl
```diff
@@ -70,8 +70,18 @@ request_and_retry(Config, ResultFun, {retry, Request}, MaxAttempts) ->
         erlcloud_aws:get_timeout(Config), Config),
     case Rsp of
         {ok, {{Status, StatusLine}, ResponseHeaders, ResponseBody}} ->
+            ResponseType = case Status >= 200 andalso Status < 300 of
```
motobob: as much as it is the only place to put it here but we should not pollute general retry module with Lambda specifics.
but rather create a dedicate retry function for lambda and use it.

motobob: one the core reasons - this  will retry it silently but with will lose the ability to even log what went wrong.
was it a function error of USER or AWS lambda function error

egobrain: The retry function wouldn't be  called in case ResponseType is 'ok'.
I can generalize determine of ResponseType by storing new kind of lambda in #aws_config. 
Is it ok ?

egobrain: > This will retry it silently but with will lose the ability to even log what went wrong

You may write your own retry function and log reason there :)

motobob: IMO. this should top level application code smth like:
```erlang
my_lambda_retry(AwsReq) when <http 200 but lambda fail> ->
    %% decide if need retry my lambda, do logs and whatever else
    {retry, Req};
  (_) ->  erlcloud_retry:default_retry(AwsReq).
      
 AwsCfg#aws_config{
        retry = fun my_lambda_retry/1,
        retry_num = N
}
```

motobob: ignore ^

egobrain: @motobob please have a look on new solution

src/erlcloud_retry.erl
```diff
@@ -52,6 +56,27 @@ request(Config, #aws_request{attempt = 0} = Request, ResultFun) ->
...
+-spec only_http_errors(#aws_request{}) -> ok | error.
+only_http_errors(#aws_request{response_status=Status})
+  when Status >= 200, Status < 300
+       ->
+    ok;
+only_http_errors(_) ->
+    error.
+
+-spec aws_api_errors(#aws_request{}) -> ok | error.
+aws_api_errors(#aws_request{response_status=Status, response_headers=ResponseHeaders})
```
motobob: s/aws_api_errors/lambda_fun_errors

include/erlcloud_aws.hrl
```diff
@@ -137,6 +137,11 @@
           %% If you provide a custom function be aware of this anticipated change.
           %% See erlcloud_retry for full documentation.
           retry=fun erlcloud_retry:no_retry/1::erlcloud_retry:retry_fun(),
+
+          %% By default treat all non 2xx http error codes as errors.
+          %% But in some cases, like lambda call it useful to override such
+          %% behaviour by custom one.
+          response_type=fun erlcloud_retry:only_http_errors/1::erlcloud_retry:response_type_fun(),
```
kkuzmin: I wouldn't complicate retry logic with yet another flag, but instead I suggest we add a new helper retry function `erlcloud_retry:lambda_retry()` which pays attention to `x-amz-function-error` header and retries it. Essentially in this change rename `aws_api_errors` to `lambda_retry()` and remove everything else.

egobrain: @kkuzmin in current implementation we can't handle `x-amz-function-error` in retry, because response is marked as 'ok' and retry function wouldn't be called at all.
So the only solution is to allow to mark such responses as errors by introduction new 'response_type' callback.
It gives as flexibility without any breaking changes.

kkuzmin: I followed the code path and that makes sense now. Thank you.
In this case I think `retry_response_type` looks clearer than `response_type` in `#aws_config{}`. Could you please rename it.

PR #591, opened 2019-04-23
User defined s3 region

pedro-gutierrez:
Hi, 

This is a small change that seems to add compatibility with minio. It allows the user to define a s3 region, rather than inferring it from the hostname.

### DIFF
--- a/include/erlcloud_aws.hrl
+++ b/include/erlcloud_aws.hrl
@@ -17,6 +17,7 @@
           sts_host="sts.amazonaws.com"::string(),
           s3_scheme="https://"::string(),
           s3_host="s3.amazonaws.com"::string(),
+          s3_region="us-east-1"::string(),
           s3_port=80::non_neg_integer(),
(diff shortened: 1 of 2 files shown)


motobob:
i would rather change aws_region_from_host itself to understand the S3 on-prem version.

motobob:
i would rather change aws_region_from_host itself to understand the S3 on-prem version.

pedro-gutierrez:
Hey thank you for your feedback. I am happy to do that. Would you have any pointers on how you would identity that class of provider?

motobob:
i have not worked with MinIO before. they must have a notion of a region there...since it's used in SigV4 calculation. 
also adding a function clause to that function which detects <some_minio_host_pattern> should help

pedro-gutierrez:
Oh I see what you mean. Actually you may deploy a minio cluster under any arbitrary hostname. For instance, Minio themselves have a shared, public instance available at https://play.minio.io:9000. There is no region information encoded in the hostname.

PR #595, opened 2019-06-29
Add erlcloud_kinesis:list_shards/1|2|3

hmartel:
Problem to solve:

- new list_shards kinesis endpoint solves rate limitation of describe_stream, but is not exposed

Proposed solution:

- expose it
    - list_shards' api is a bit more complex than previous kinesis ones, and does not adapt easily to the approach used by erlcloud_kinesis:describe_stream (for example, one can list a large number of shards similarly to describe_stream, or use continuation token). Thus anything beside stream name and aws config is specified via an option proplists.

### DIFF
--- a/src/erlcloud_kinesis.erl
+++ b/src/erlcloud_kinesis.erl
@@ -224,35 +224,29 @@ list_shards(StreamName, Options) ->
 
 -spec list_shards(binary(), list_shards_opts(), aws_config()) -> erlcloud_kinesis_impl:json_return().
 list_shards(StreamName, Options, Config) when is_record(Config, aws_config) ->
-    case lists:keymember(next_token, 1, Options) of
+    DynamizedOptionsOrError = case lists:keymember(next_token, 1, Options) of
         false ->
-            ok;
+            dynamize_options(Options);
         true ->
             case lists:keymember(stream_creation_timestamp, 1, Options) of
                 false ->
-                    ok;
+                    dynamize_options(Options);
                 true ->
-                    error({erlcloud_kinesis, {incompatible_options, [
-                        next_token,
-                        stream_creation_timestamp
-                    ]}})
+                    {error, {incompatible_options, [next_token, stream_creation_timestamp]}}
             end,
             case lists:keymember(exclusive_start_shard_id, 1, Options) of
                 false ->
-                    ok;
+                    dynamize_options(Options);
                 true ->
-                    error({erlcloud_kinesis, {incompatible_options, [
-                        next_token,
-                        exclusive_start_shard_id
-                    ]}})
+                    {error, {incompatible_options, [next_token, exclusive_start_shard_id]}}
             end
     end,
-    case dynamize_options(Options) of
-        DynamizedOptions when is_list(DynamizedOptions) ->
-            Json = [{<<"StreamName">>, StreamName} | dynamize_options(Options)],
+    case is_list(DynamizedOptionsOrError) of
+        true ->
+            Json = [{<<"StreamName">>, StreamName} | DynamizedOptionsOrError],
             erlcloud_kinesis_impl:request(Config, "Kinesis_20131202.ListShards", Json);
-        Error ->
-            Error
+        false ->
+            DynamizedOptionsOrError
     end.
 
 %%------------------------------------------------------------------------------
--- a/src/erlcloud_kinesis.erl
+++ b/src/erlcloud_kinesis.erl
@@ -92,7 +92,12 @@ dynamize_option(Option) ->
 dynamize_options(Options) ->
     lists:foldr(fun
         (Option, Acc) when is_list(Acc) ->
-            [dynamize_option(Option) | Acc];
+            case dynamize_option(Option) of
+                {error, _} = Error ->
+                    Error;
+                DynamizedOption ->
+                    [DynamizedOption | Acc]
+            end;
         (_Option, Error) ->
             Error
     end, [], Options).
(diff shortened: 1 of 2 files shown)


hmartel:
@nalundgaard @motobob @jlbarrett

src/erlcloud_kinesis.erl
```diff
@@ -130,6 +150,101 @@ delete_stream(StreamName, Config) when is_record(Config, aws_config) ->
...
+list_shards(StreamName, Options, Config) when is_record(Config, aws_config) ->
+    case lists:keymember(next_token, 1, Options) of
+        false ->
+            ok;
+        true ->
+            case lists:keymember(stream_creation_timestamp, 1, Options) of
+                false ->
+                    ok;
+                true ->
+                    error({erlcloud_kinesis, {incompatible_options, [
```
motobob: please use error tuple return and not erlang:error

src/erlcloud_kinesis.erl
```diff
@@ -130,6 +155,106 @@ delete_stream(StreamName, Config) when is_record(Config, aws_config) ->
...
+                        next_token,
+                        exclusive_start_shard_id
+                    ]}})
+            end
+    end,
+    case dynamize_options(Options) of
+        DynamizedOptions when is_list(DynamizedOptions) ->
+            Json = [{<<"StreamName">>, StreamName} | dynamize_options(Options)],
```
jeremey: Need to elide `StreamName` if `NextToken` is provided.

```
{error,{<<"InvalidArgumentException">>,
        <<"NextToken and StreamName cannot be provided together.">>}}
```

src/erlcloud_kinesis.erl
```diff
@@ -130,6 +155,100 @@ delete_stream(StreamName, Config) when is_record(Config, aws_config) ->
...
+            case lists:keymember(exclusive_start_shard_id, 1, Options) of
+                false ->
+                    dynamize_options(Options);
+                true ->
+                    {error, {incompatible_options, [next_token, exclusive_start_shard_id]}}
+            end
+    end,
+    case is_list(DynamizedOptionsOrError) of
+        true ->
+            Json = [{<<"StreamName">>, StreamName} | DynamizedOptionsOrError],
```
jeremey: Need to make this list further up (line 229), i.e. whatever `DynamizedOptionsOrError` is here if it's a list is good to go, no?

src/erlcloud_kinesis.erl
```diff
@@ -72,6 +78,25 @@ configure(AccessKeyID, SecretAccessKey, Host, Port) ->
...
+dynamize_option({stream_creation_timestamp, Value}) ->
+    {<<"StreamCreationTimestamp">>, Value};
+dynamize_option(Option) ->
+    {error, {invalid_option, Option}}.
+
+dynamize_options(Options) ->
+    lists:foldr(fun
+        (Option, Acc) when is_list(Acc) ->
+            [dynamize_option(Option) | Acc];
+        (_Option, Error) ->
```
motobob: it will always be a list with ^ function

src/erlcloud_kinesis.erl
motobob: 1) does not address stream name + next token issue reported by @jlbarrett 
2) DynamizedOptionsOrError = dynamize_options(Options) can have `{error, {invalid...}}` and that is not caught here. 

I would suggest handling only two cases: 
- (next_token + max_result) and ignore other options. 
- any other combination and let user just get AWS error back.

jeremey: IMO you should discard all this error checking nonsense, AWS can check its own errors (as is true for all the other APIs here). i.e. just this:

```erlang
-spec list_shards(binary(), list_shards_opts(), aws_config()) -> erlcloud_kinesis_impl:json_return().
list_shards(StreamName, Options, Config) when is_record(Config, aws_config) ->
    StreamOpts = case lists:keymember(next_token, 1, Options) of
                     true -> [];
                     _ -> [{<<"StreamName">>, StreamName}]
                 end,
    Json = StreamOpts ++ dynamize_options(Options),
    erlcloud_kinesis_impl:request(Config, "Kinesis_20131202.ListShards", Json).
```

hmartel:
@motobob

src/erlcloud_kinesis.erl
motobob: addressed

src/erlcloud_kinesis.erl
```diff
@@ -130,6 +160,87 @@ delete_stream(StreamName, Config) when is_record(Config, aws_config) ->
...
+    Json = dynamize_options(Options),
+    erlcloud_kinesis_impl:request(Config, "Kinesis_20131202.ListShards", Json);
+list_shards(StreamName, Options) ->
+    list_shards(StreamName, Options, default_config()).
+
+-spec list_shards(binary(), list_shards_opts(), aws_config()) -> erlcloud_kinesis_impl:json_return().
+list_shards(StreamName, Options, Config) when is_record(Config, aws_config) ->
+    case dynamize_options(Options) of
+        DynamizedOptions when is_list(DynamizedOptions) ->
+            Json = [{<<"StreamName">>, StreamName} | DynamizedOptions],
```
motobob: one last thing @hmartel . I guess diff played a error game here with diff.
```erlang
{error,{<<"InvalidArgumentException">>,
        <<"NextToken and StreamName cannot be provided together.">>}}
```
i guess AWS also overdid some validation

motobob: >Don't specify StreamName or StreamCreationTimestamp if you specify NextToken because the latter unambiguously identifies the stream.

You can optionally specify a value for the MaxResults parameter when you specify NextToken. If you specify a MaxResults value that is less than the number of shards that the operation returns if you don't specify MaxResults, the response will contain a new NextToken value. You can use the new NextToken value in a subsequent call to the ListShards operation.

jeremey: See https://github.com/erlcloud/erlcloud/pull/595#discussion_r299117155 ;)

motobob: oh https://github.com/erlcloud/erlcloud/pull/595/files/50da38e8a4742985e8c3b3cf3dcb95aba1019c6f#diff-c1a9bde5cb98f49553dd4d52fac55a35R229

jeremey: I'm ok with this I suppose since it's just the same rope AWS gives you. Just can't have a real implementation using `/3` only b/c the second call must be to `/2`

src/erlcloud_kinesis.erl
```diff
@@ -72,6 +78,30 @@ configure(AccessKeyID, SecretAccessKey, Host, Port) ->
 
 default_config() -> erlcloud_aws:default_config().
 
+dynamize_option({exclusive_start_shard_id, Value}) ->
```
jeremey: Need support for `{stream_name, StreamName}` here so one can build a complete implementation without having to call /3 *and* /2.

src/erlcloud_kinesis.erl
```diff
@@ -72,6 +78,30 @@ configure(AccessKeyID, SecretAccessKey, Host, Port) ->
...
+    {<<"ExclusiveStartShardId">>, Value};
+dynamize_option({max_results, Value}) when Value >= 1, Value =< 10000 ->
+    {<<"MaxResults">>, Value};
+dynamize_option({next_token, Value}) ->
+    {<<"NextToken">>, Value};
+dynamize_option({stream_creation_timestamp, Value}) ->
+    {<<"StreamCreationTimestamp">>, Value};
+dynamize_option(Option) ->
```
motobob: need StreamName here i guess

PR #597, opened 2019-07-08
Implement Application Auto Scaling module

Santiniis:
Add API calls
Export functions
Add failed attempts retrying with exponential backoff
Offload parsing of responses separate functions
Add optional boolean type to json decoder


src/erlcloud_application_autoscaler.erl
```diff
@@ -0,0 +1,689 @@
...
+) -> response().
+register_scalable_target(Configurations, BodyConfigurations) ->
+    request_with_action(Configurations, BodyConfigurations, "AnyScaleFrontendService.RegisterScalableTarget").
+
+
+%%%------------------------------------------------------------------------------
+%%% Internal functions.
+%%%------------------------------------------------------------------------------
+
+request_with_action(Configuration, BodyConfigurations, Action) ->
```
motobob: why not using already erlcloud_retry module?

src/erlcloud_application_autoscaler.erl
```diff
@@ -0,0 +1,689 @@
...
+
+-export([configure/2, configure/3, configure/4, configure/5,
+         new/2, new/3, new/4, new/5]).
+-export([default_config/0]).
+
+%% ------------------------------------------------------------------
+%% Macro Definitions
+%% ------------------------------------------------------------------
+
+-define(NUM_ATTEMPTS, 10).
```
motobob: there is a param for this in `#aws_config.retry_num` please use that

src/erlcloud_application_autoscaler.erl
```diff
@@ -0,0 +1,689 @@
...
+configure(AccessKeyID, SecretAccessKey, Host, Port, Scheme) ->
+    erlcloud_config:configure(AccessKeyID, SecretAccessKey, Host, Port, Scheme, fun new/5).
+
+default_config() -> erlcloud_aws:default_config().
+
+%%%------------------------------------------------------------------------------
+%%% AWS Application Autoscaling functions.
+%%%------------------------------------------------------------------------------
+
+%% DeleteScalingPolicy
```
motobob: stylish: please add links to AWS API docs

src/erlcloud_application_autoscaler.erl
```diff
@@ -0,0 +1,689 @@
...
+        undefined -> [];
+        Min -> [{<<"MinCapacity">>, Min}]
+    end,
+    register_scalable_target(Configuration, BodyProps ++ MaybeBodyWithMax ++ MaybeBodyWithMin).
+
+-spec register_scalable_target(
+    Configurations :: erlcloud_aws:aws_config(),
+    BodyConfigurations :: aws_aas_request_body()
+) -> response().
+register_scalable_target(Configurations, BodyConfigurations) ->
```
motobob: singular? s/Configurations/Configuration

src/erlcloud_application_autoscaler.erl
```diff
@@ -0,0 +1,689 @@
...
+            {error, Reason}
+    end.
+
+request_and_retry(_, _, _, {error, Reason}) ->
+    {error, Reason};
+request_and_retry(_, _, _, {attempt, Attempt}) when Attempt =:= 11 ->
+    {error, no_answer};
+request_and_retry(Config, Headers, Body, {attempt, Attempt}) ->
+    case erlcloud_httpc:request(url(Config), post, Headers, Body, timeout(Attempt, Config), Config) of
+
```
motobob: optional stylish: no new lines inside the fun.

src/erlcloud_application_autoscaler.erl
```diff
@@ -0,0 +1,689 @@
...
+    "";
+port_spec(#aws_config{autoscaling_port=Port}) ->
+    [":", erlang:integer_to_list(Port)].
+
+
+headers(Config, Operation, Body) ->
+    Headers = [{"host", Config#aws_config.autoscaling_host},
+               {"x-amz-target", Operation}],
+
+    erlcloud_aws:sign_v4_headers(Config, Headers, Body, erlcloud_aws:aws_region_from_host(Config#aws_config.autoscaling_host), "application-autoscaling").
```
motobob: not sure if reuse of autoscaling_host inside the record is a good idea. suggest to have a dedicated one.

Santiniis: I am a bit confused about this comment, could you elaborate a bit more, or give an example of how it could go wrong? Is it because it can be confused with the `as_host`, or is there something else I'm missing ?

src/erlcloud_application_autoscaler.erl
Santiniis: I was taking as a basis the `erlcloud_ddb2` module and the implementation of its requests (`erlcloud_ddb_impl`), so I didn't acknowledge a possible implementation using `erlcloud_retry`.

src/erlcloud_application_autoscaler.erl
Santiniis: Or can it be that I may have been mistaken and that this endpoint is actually the EC2/basic autoscaling, as is the `as_host`, and it would be more correct to have something like `aws_config.application_autoscaling_*` fields?

motobob: sorry confusing comment. yes, the latter is what i meant. please use a dedicate host config params. Smth like `app_autoscalling`

motobob:
to fix test please change erclloud_aws:service_config/3

Santiniis:
@motobob could you review to check if it is up to standards now?

motobob:
PR is missing unit tests which test in/out but it's fine for initial go.

PR #607, opened 2019-07-18
Make 'TransactionCanceledException' error handling extract 'CancellationReasons' values and return it in the error

va17:
In order to make transaction error handling more pattern-match friendly, this change makes the `client_error` handler extract `CancellationReasons` values from `TransactionCanceledException` and return the error as `{error, {Type, {Message, CancellationReasons}}}` instead of just `{error, {Type, Message}}`.

**NOTE** that this output change applies to `TransactionCanceledException` error type only.

Although the documentation states it returns "NONE" code for no error in cancelation reasons output, it is actually `None`. `"message": null` is also not present, see real DDB transaction canceled example below:

```json
{
"__type": "com.amazonaws.dynamodb.v20120810#TransactionCanceledException",
"CancellationReasons": [
    {"Code": "None"},
    {"Code": "ConditionalCheckFailed",
     "Message": "The conditional request failed"}
],
"Message": "Transaction cancelled, please refer cancellation reasons for specific reasons [None, ConditionalCheckFailed]"
}
```

@nalundgaard

### DIFF
--- a/src/erlcloud_ddb2.erl
+++ b/src/erlcloud_ddb2.erl
@@ -61,12 +61,18 @@
 %% contain typed attribute values so that they may be correctly passed
 %% to subsequent calls.
 %%
-%% DynamoDB errors are return in the form `{error, {ErrorCode,
-%% Message}}' where `ErrorCode' and 'Message' are both binary
+%% DynamoDB errors for most cases are returned in the form
+%% `{error, {ErrorCode, Message}}' where `ErrorCode' and `Message' are both binary
 %% strings. List of error codes:
 %% [http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ErrorHandling.html]. So
 %% to handle conditional check failures, match `{error,
 %% {<<"ConditionalCheckFailedException">>, _}}'.
+%% Note that in the case of a `TransactionCanceledException' DynamoDB error,
+%% the error response has the form `{error, {<<"TransactionCanceledException">>,
+%% {Message, CancellationReasons}}}' where `Message' is a binary string and
+%% `CancellationReasons' is an ordered list in the form `[{Code, Message}]',
+%% where `Code' is the status code of the result and `Message' is the cancellation
+%% reason message description.
 %%
 %% `erlcloud_ddb_util' provides a higher level API that implements common
 %% operations that may require multiple DynamoDB API calls.
@@ -3213,6 +3219,12 @@ transact_get_items(RequestItems, Opts) ->
 %% DynamoDB API:
 %% [https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TransactGetItems.html]
 %%
+%% Note that in the case of a `TransactionCanceledException' DynamoDB error, the error
+%% response has the form `{error, {<<"TransactionCanceledException">>, {Message, CancellationReasons}}}'
+%% where `Message' is a binary string and `CancellationReasons' is an ordered list in the form
+%% `[{Code, Message}]', where `Code' is the status code of the result and `Message' is the cancellation
+%% reason message description.
+%%
 %% ===Example===
 %%
 %% Get two items in a transaction.
@@ -3347,6 +3359,12 @@ transact_write_items(RequestItems, Opts) ->
 %% @doc
 %% DynamoDB API:
 %% [https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TransactWriteItems.html]
+%%
+%% Note that in the case of a `TransactionCanceledException' DynamoDB error, the error
+%% response has the form `{error, {<<"TransactionCanceledException">>, {Message, CancellationReasons}}}'
+%% where `Message' is a binary string and `CancellationReasons' is an ordered list in the form
+%% `[{Code, Message}]', where `Code' is the status code of the result and `Message' is the cancellation
+%% reason message description.
 %%
 %% ===Example===
 %%
--- a/src/erlcloud_ddb_impl.erl
+++ b/src/erlcloud_ddb_impl.erl
@@ -240,11 +240,19 @@ client_error(Body, DDBError) ->
                         [_, Type] when
                               Type =:= <<"ProvisionedThroughputExceededException">> orelse
                               Type =:= <<"ThrottlingException">> ->
-                            DDBError#ddb2_error{error_type = ddb, 
+                            DDBError#ddb2_error{error_type = ddb,
                                                 should_retry = true,
                                                 reason = {Type, Message}};
+                        [_, Type] when
+                              Type =:= <<"TransactionCanceledException">> ->
+                            CancellationReasons0 = proplists:get_value(<<"CancellationReasons">>, Json, []),
+                            CancellationReasons = [{proplists:get_value(<<"Code">>, R),
+                                                    proplists:get_value(<<"Message">>, R, null)} || R <- CancellationReasons0],
+                            DDBError#ddb2_error{error_type = ddb,
+                                                should_retry = should_retry_canceled_transaction(CancellationReasons),
+                                                reason = {Type, {Message, CancellationReasons}}};
                         [_, Type] ->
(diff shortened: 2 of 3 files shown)


src/erlcloud_ddb_impl.erl
```diff
@@ -240,11 +240,19 @@ client_error(Body, DDBError) ->
...
+                            DDBError#ddb2_error{error_type = ddb,
                                                 should_retry = true,
                                                 reason = {Type, Message}};
+                        [_, Type] when
+                              Type =:= <<"TransactionCanceledException">> ->
+                            CancellationReasons0 = proplists:get_value(<<"CancellationReasons">>, Json, []),
+                            CancellationReasons = [{proplists:get_value(<<"Code">>, R),
+                                                    proplists:get_value(<<"Message">>, R, <<>>)} || R <- CancellationReasons0],
+                            DDBError#ddb2_error{error_type = ddb,
+                                                should_retry = false,
```
motobob: why not retry if one of the reason for transaction fails is conflicting transactions?
@hmartel-alertlogic FYI

nalundgaard: @motobob I don't think it's an OK idea. Transactions may be canceled for reasons ranging from provisioned throughput exceptions to conditional check failure.

If we want to really "magic" it, we could make a `should_retry(CancellationReasonsCodes)` that would *only* set retry to true if the CancellationReasons are all `None`,  `ProvisionedThroughputExceeded`, or `ThrottlingError`. @hmartel-alertlogic what do you think?

nalundgaard: @motobob FYI, check out [TransactWriteItems](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TransactWriteItems.html) docs, there's `ConditionalCheckFailed`, `ItemCollectionSizeLimitExceeded`, `TransactionConflict`, and `ValidationError`. *None* of these should ever be retried on.

src/erlcloud_ddb_impl.erl
```diff
@@ -240,11 +240,19 @@ client_error(Body, DDBError) ->
...
                               Type =:= <<"ThrottlingException">> ->
-                            DDBError#ddb2_error{error_type = ddb, 
+                            DDBError#ddb2_error{error_type = ddb,
                                                 should_retry = true,
                                                 reason = {Type, Message}};
+                        [_, Type] when
+                              Type =:= <<"TransactionCanceledException">> ->
+                            CancellationReasons0 = proplists:get_value(<<"CancellationReasons">>, Json, []),
+                            CancellationReasons = [{proplists:get_value(<<"Code">>, R),
+                                                    proplists:get_value(<<"Message">>, R, <<>>)} || R <- CancellationReasons0],
```
nalundgaard: shouldn't the default here be `null`?

va17: Thanks. Fixed.

src/erlcloud_ddb_impl.erl
va17: As suggested by @nalundgaard, I added a function to set `should_retry` to `true` only if cancellation reasons are either: `None`, `ThrottlingError` and/or `ProvisionedThroughputExceeded`.

src/erlcloud_ddb2.erl
```diff
@@ -61,12 +61,17 @@
...
+%% DynamoDB errors for most cases are return in the form
+%% `{error, {ErrorCode, Message}}' where `ErrorCode' and `Message' are both binary
 %% strings. List of error codes:
 %% [http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ErrorHandling.html]. So
 %% to handle conditional check failures, match `{error,
 %% {<<"ConditionalCheckFailedException">>, _}}'.
+%% Note that for `TransactionCanceledException' DynamoDB error type, it is returned
+%% in the form `{error, {<<"TransactionCanceledException">>, {Message,
+%% CancellationReasons}}}' where `Message' is a binary string and `CancellationReasons'
+%% is a proplist in the form `[{Code, Message}]` where `Code' is the status code of the result
```
nalundgaard: I think we should call this an "ordered list" or something. And use `ErrorCode` rather than `Code` for consistency.

va17: AWS calls it `Code` since it is not always an error.

va17: `ReasonCode` maybe?

nalundgaard: Good point. Never mind, leave it as is.

src/erlcloud_ddb2.erl
```diff
@@ -61,12 +61,17 @@
 %% contain typed attribute values so that they may be correctly passed
 %% to subsequent calls.
 %%
-%% DynamoDB errors are return in the form `{error, {ErrorCode,
-%% Message}}' where `ErrorCode' and 'Message' are both binary
+%% DynamoDB errors for most cases are return in the form
```
nalundgaard: s/are return/are returned

va17: Fixed. Ty.

src/erlcloud_ddb2.erl
```diff
@@ -61,12 +61,17 @@
...
+%% DynamoDB errors for most cases are returned in the form
+%% `{error, {ErrorCode, Message}}' where `ErrorCode' and `Message' are both binary
 %% strings. List of error codes:
 %% [http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ErrorHandling.html]. So
 %% to handle conditional check failures, match `{error,
 %% {<<"ConditionalCheckFailedException">>, _}}'.
+%% Note that for `TransactionCanceledException' DynamoDB error type, it is returned
+%% in the form `{error, {<<"TransactionCanceledException">>, {Message,
+%% CancellationReasons}}}' where `Message' is a binary string and `CancellationReasons'
+%% is a proplist in the form `[{Code, Message}]` where `Code' is the status code of the result
```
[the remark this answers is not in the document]
va17: Done.

src/erlcloud_ddb2.erl
```diff
@@ -3213,6 +3218,12 @@ transact_get_items(RequestItems, Opts) ->
 %% DynamoDB API:
 %% [https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TransactGetItems.html]
 %%
+%% Note that the error response in case of `TransactionCanceledException' DDB type error,
```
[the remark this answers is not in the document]
va17: Thanks for the input. Addressed.

src/erlcloud_ddb_impl.erl
hmartel: Retrying in case of conflicting transaction would break patterns such as:

- read some values
- prepare update using read values
- update only if read values have not changed

The "prepare update using read values" needs to be redone. We would need to pass a fun to do that work... goes way beyond what erlclould_ddb2 should do.

nalundgaard: RIght, @hmartel-alertlogic auto-retrying on a conflicting transaction is a non-starter, but the question is whether auto-retrying on exclusively throttling/provisioned throughput exceeded errors is a good idea. That is what @va17 has implemented here. Let me know if you think that is an issue.

nalundgaard:
I'm OK with this. I think given the placement of the "incompatibility" (replacing `Message` with `{Message, CancellationReasons}`) is an OK pattern to use, since consumers have always been warned to ignore `Message` with a match like `{error, {Reason, _Message}}`. Having these reasons is really helpful in being able to return right error codes to clients in your own APIs, (e.g., 400 Bad Request vs. 409 Conflict vs. 500 for throttling, etc.). An optional thing, but this is needed for us to be able to do it.

motobob:
yes. i;'m okay with the break here. 👍  :shipit:

nalundgaard:
Epic 🤦‍♀ at AWS for making a `TransactionCanceledException` with a Message, "Transaction **cancelled**, please refer cancellation reasons for specific reasons..."

Why not both?

PR #610, opened 2019-07-31
Adding the CustomVerificationTemplate API

stephb9959:
First time I contribute - so be gentle...

Just added support for the whole CustomVerificationTemplate API. This includes Create, Update, List, Delete, Get as well as SendCustomVerificationEmail. Added corresponding error codes. Tried to add specs & types. I tested this as much as I could. No actual tests were added.

This is the [API](https://docs.aws.amazon.com/ses/latest/APIReference/API_GetCustomVerificationEmailTemplate.htm)

Thanks

### DIFF
--- a/src/erlcloud_ses.erl
+++ b/src/erlcloud_ses.erl
@@ -43,8 +43,21 @@
 -export([verify_email_identity/1, verify_email_identity/2]).
 -export([verify_domain_identity/1, verify_domain_identity/2]).
 
--include("erlcloud.hrl").
--include("erlcloud_aws.hrl").
+-export([ create_custom_verification_email_template/6,
+          create_custom_verification_email_template/7,
+          update_custom_verification_email_template/2,
+          update_custom_verification_email_template/3,
+          send_custom_verification_email/3,
+          delete_custom_verification_email_template/1,
+          delete_custom_verification_email_template/2,
+          get_custom_verification_email_template/2,
+          get_custom_verification_email_template/1,
+          list_custom_verification_email_templates/0,
+          list_custom_verification_email_templates/1]).
+
+
+-include("../include/erlcloud.hrl").
+-include("../include/erlcloud_aws.hrl").
 
 -define(API_VERSION, "2010-12-01").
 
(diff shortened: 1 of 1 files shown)


src/erlcloud_ses.erl
```diff
@@ -132,6 +145,147 @@ delete_identity(Identity, Config) ->
...
+  Params = encode_params(
+            [   { template_name , TemplateName} ,
+                { from_email_address , FromEmailAddress } ,
+                { template_subject , TemplateSubject } ,
+                { template_content , TemplateContent } ,
+                { success_redirect_url , SuccessRedirectionURL} ,
+                { failure_redirect_url , FailureRedirectionURL} ] ),
+  case ses_request(Config, "CreateCustomVerificationEmailTemplate", Params) of
+    {ok, _Doc} ->
+      ok;
```
motobob: stylish: either both on new line or both on the same line

src/erlcloud_ses.erl
```diff
@@ -794,8 +948,23 @@ encode_params([{source, Source} | T], Acc) when is_list(Source); is_binary(Sourc
     encode_params(T, [{"Source", Source} | Acc]);
 encode_params([{subject, Subject} | T], Acc) ->
     encode_params(T, encode_content("Message.Subject", Subject, Acc));
+encode_params([{template_name, TemplateName} | T], Acc) when is_list(TemplateName); is_binary(TemplateName) ->
```
motobob: stylish:
```
fun(...)
  when ... ->
    stuff;
```
is better then line wraps and definitely better then wrapping in the middle of the guards

src/erlcloud_ses.erl
```diff
@@ -794,8 +948,23 @@ encode_params([{source, Source} | T], Acc) when is_list(Source); is_binary(Sourc
...
+encode_params([{template_content, TemplateContent} | T], Acc) when is_list(TemplateContent); is_binary(TemplateContent) ->
+  encode_params(T, [{"TemplateContent", TemplateContent} | Acc]);
+encode_params([{success_redirect_url, SuccessRedirectionURL} | T], Acc) when is_list(SuccessRedirectionURL); is_binary(SuccessRedirectionURL) ->
+  encode_params(T, [{"SuccessRedirectionURL", SuccessRedirectionURL} | Acc]);
+encode_params([{failure_redirect_url, FailureRedirectionURL} | T], Acc) when is_list(FailureRedirectionURL); is_binary(FailureRedirectionURL) ->
+  encode_params(T, [{"FailureRedirectionURL", FailureRedirectionURL} | Acc]);
 encode_params([Option | _], _Acc) ->
-    error({erlcloud_ses, {invalid_parameter, Option}}).
+  error({erlcloud_ses, {invalid_parameter, Option}}).
+
```
motobob: 👾 no need for these lines

src/erlcloud_ses.erl
```diff
@@ -43,6 +43,19 @@
 -export([verify_email_identity/1, verify_email_identity/2]).
 -export([verify_domain_identity/1, verify_domain_identity/2]).
 
+-export([ create_custom_verification_email_template/6,
```
nalundgaard: style nit: can you please regroup the exports like the rest of the module? The intent is for the export order to mimic the API documentation.

src/erlcloud_ses.erl
```diff
@@ -132,6 +145,147 @@ delete_identity(Identity, Config) ->
...
+%%  send_custom_verification_email
+%%  delete_custom_verification_email_template
+%%  get_custom_verification_email_template
+%%  list_custom_verification_email_templates
+%%
+%% @end
+%%------------------------------------------------------------------------------
+
+-type custom_template_attribute_names() :: template_name | from_email_address | template_subject | template_content | success_redirect_url | failure_redirect_url .
+-type custom_template_attributes() :: [ { custom_template_attribute_names() , string() } ].
```
nalundgaard: I think that this should be incorporated into the `Common types` section near the top of the module.

src/erlcloud_ses.erl
```diff
@@ -132,6 +145,147 @@ delete_identity(Identity, Config) ->
...
+
+-type create_custom_verification_email_template_result() :: ok | { error , term() }.
+
+-spec create_custom_verification_email_template( string() , string() ,string() ,string() ,string() ,string() ) -> create_custom_verification_email_template_result().
```
nalundgaard: a couple of other style nits:

* the spacing between args, commas, and braces here (and in a number of other places in the code) is inconsistent to itself and the rest of the module. Some of the lines are also really, really wide. I prefer if you stick with
  ```erlang
  -spec my_fun(string(), string(), string) -> result().
  my_fun(Arg1, Arg2, Arg3) ->
      MyProplist = [{v1, Arg1},
                    {v2, Arg2},
                    {v3, Arg3}],
      do_stuff(MyProplist).
  
  
  -spec my_really_long_function_name(string(), string(), string(), aws_config()) ->
          my_really_long_function_name_result().
  ```
* the ordering of these functions in the module is inconsistent to the others here. It would be better if they were placed in alphabetical order into the module, consistent with how the others have been contributed.

src/erlcloud_ses.erl
stephb9959: Agreed. Done in next commit.

src/erlcloud_ses.erl
stephb9959: Agreed. Done in next check in.

src/erlcloud_ses.erl
stephb9959: I grouped and ordered them alphabetically. Next checkin.

src/erlcloud_ses.erl
stephb9959: OK. So I stuck now with what I have seen. Short stays on the same line, long, 2+ lines. Next checkin.

src/erlcloud_ses.erl
stephb9959: Got it. Fixed.

src/erlcloud_ses.erl
stephb9959: Done.

src/erlcloud_ses.erl
```diff
@@ -43,6 +43,13 @@
 -export([verify_email_identity/1, verify_email_identity/2]).
 -export([verify_domain_identity/1, verify_domain_identity/2]).
 
+-export([create_custom_verification_email_template/6, create_custom_verification_email_template/7]).
```
nalundgaard: Just to clarify on grouping, I was suggesting that `create_custom_verification_email_template` go above `delete_identity`, that `update_custom_verification_email_template` go between `set_identity_notification_topic` and `verify_domain_dkim`, etc. In other words, to me, the cadre of functions you've contributed should be interspersed in the module in function name order.

src/erlcloud_ses.erl
```diff
@@ -132,6 +144,136 @@ delete_identity(Identity, Config) ->
         {error, Reason} -> {error, Reason}
     end.
 
+%%------------------------------------------------------------------------------
+%% @doc
+%% Custom Verification Templates
+%% SES API:
+%% [https://docs.aws.amazon.com/ses/latest/APIReference/API_GetCustomVerificationEmailTemplate.html]
```
nalundgaard: One other thing: the rest of the module has documentation for each function, with a direct link to the AWS documentation for that API call and a succinct example. I think it would be best to use that same pattern here.

src/erlcloud_ses.erl
stephb9959: OK. Now I see what you mean. Personally, it's a way I have never like. Alphabetically grouping functions make using an API harder. Grouping by object type (i.e. all the CRUD ops for a specific object together for example) makes it easier to search when looking at the source. That said, I will follow the style of the API as it is currently and intersperse the functions. Now it's very clear to me.

src/erlcloud_ses.erl
stephb9959: If I move the functions in alphabetical order, then absolutely, each function should have a link to the actual API.

src/erlcloud_ses.erl
nalundgaard: I understand the preference, but IMO consistency (especially within a module) is a lot more valuable than just having each contributor organize the code as they see fit. I'd be open to regrouping the module as a whole separately from extending it, but as-is, contributions should be consistent with existing patterns.

Looks like your update of the code blocks is consistent, and just remains to intersperse the exports to be interspersed in symmetric order here at the top.

Last question is, will you make the unit tests, or should we just merge this and raise an issue to do so in the future?

motobob:
thank you for contribution

nalundgaard:
+1, thanks!

PR #614, opened 2019-08-16
Add function clause dynamize_local_secondary_indexes

raghavkarol:
Handle the case where we are passing in a HaskKey only.


motobob:
but SecodIndx always has a sort key, no? so the current behaviour is correct.

@nalundgaard

raghavkarol:
Yes that is correct and something I did not know. The table must have a key if a local secondary needs to be added. With this change `erlcloud_ddb2` returns 

```
{error,{<<"ValidationException">>,
                 <<"Local Secondary indices are not allowed on hash tables, only hash and range tables">>}}
```

which IMO is better than a function clause error without any hint about the failure.

motobob:
@raghavkarol rebase and merge?

PR #617, opened 2019-08-19
upgrade meck version

brunorafa:


motobob:
need lock file update too

PR #619, opened 2019-09-04
Fix unknown type

paulo-ferraz-oliveira:
There's a bunch of other dialyzer -related warnings (in the past I'd solved a bunch of them, but they keep showing up 😢), but this one is what is showing up in our analysis. I'll keep fixing them as I find them, for our specific use case, but do believe that **good -spec().s** help everybody to code better.


motobob:
sorry, we are not keeping that..probably we need to update the travis robot to check for dialyzer. 
i'm not really in favour of it but i value +100500 the specs.

paulo-ferraz-oliveira:
If this passes continuous-integration checks could we have a tag for it, please, e.g. `3.2.12`?

motobob:
sure thing. cc @nalundgaard

paulo-ferraz-oliveira:
> sorry, we are not keeping that..probably we need to update the travis robot to check for dialyzer.
> i'm not really in favour of it but i value +100500 the specs.

I didn't get that last part 😃. You find value in the `-spec()`s but wouldn't want to enforce it on the lib.'s collaborators, is that it? It's understandable, but some might be as strict as "warnings as errors".

nalundgaard:
I see the same problem with dialyzer: it's modestly valuable, but super finicky to get right. the wall of errors that are present now make it hard to start "partially" enforcing it on "clean" modules. If a contributor would like to clean it up and then start enforcing, or tune the dialyzer output to ignore issues in broken modules, I would support that.

The main issue is that I don't see the value of getting dialyzer to be "clean" being proportional to the cost of doing so.

paulo-ferraz-oliveira:
Sure. Thanks for the tag. 😉

PR #623, opened 2019-11-06
allow raw options to be set and MFA use

motobob:
@nalundgaard @kkuzmin take a look.

this is raw and unfinished but want to get feedback before processing.

### DIFF
--- a/src/erlcloud_sts.erl
+++ b/src/erlcloud_sts.erl
@@ -9,6 +9,7 @@
 -endif.
 
 -export([assume_role/4, assume_role/5,
+         assume_role/2,
          get_caller_identity/1,
          get_federation_token/3,
          get_federation_token/4]).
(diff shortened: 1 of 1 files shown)


src/erlcloud_sts.erl
```diff
@@ -9,6 +9,7 @@
 -endif.
 
 -export([assume_role/4, assume_role/5,
+         assume_role/2,
```
nalundgaard: seems like this should precede `assume_role/4, assume_role/5`

src/erlcloud_sts.erl
```diff
@@ -17,6 +18,18 @@
 -define(UTC_TO_GREGORIAN, 62167219200).
 -define(EXTERNAL_ID_MAX_LEN, 1224).
 
+% See http://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html
+% Options version of the API allowing MFA and etc access
+%% {"RoleArn", RoleArn},
```
nalundgaard: I like the idea of a /2 arity of `assume_role` that takes options, but I think I'd prefer to pass those as underscore-case atoms (`role_arn`, `role_session_name`, etc.) to be symmetric to the response format. Seems like maybe the `assume_role/2` could extract the options and call `assume_role/7` much as @kkuzmin suggested.

PR #625, opened 2019-12-06
add cloudformation create stack function

daniel-greening:
## Change
Add support for cloudformation service `create_stack` api: https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_CreateStack.html

Example Usage:
```erlang
create_my_stack() ->
    {ok, AwsConfig} = erlcloud_aws:profile(),
    Spec = #cloudformation_create_stack_input{
        stack_name="TestStack",
        template_url="https://example-domain.net/cloudformation-template.json"
    },
    {ok, StackId} = erlcloud_cloudformation:create_stack(Spec, AwsConfig).
```

### DIFF
--- a/src/erlcloud_cloudformation.erl
+++ b/src/erlcloud_cloudformation.erl
@@ -170,17 +170,13 @@ delete_stack(Spec = #cloudformation_delete_stack_input{}) ->
     delete_stack(Spec, default_config()).
 
 -spec delete_stack(cloudformation_delete_stack_input(), aws_config()) ->
-    {ok, string()} | {error, error_reason()}.
+    ok | {error, error_reason()}.
 delete_stack(Spec = #cloudformation_delete_stack_input{}, Config = #aws_config{}) ->
 
     Params = delete_stack_input_to_params(Spec),
     case cloudformation_request(Config, "DeleteStack", Params) of
-        {ok, XmlNode} ->
-            RequestId = erlcloud_xml:get_text(
-                    "/DeleteStackResponse/DeleteStackResult/RequestId",
-                    XmlNode,
-                    undefined),
-            {ok, RequestId};
+        {ok, _XmlNode} ->
+            ok;
         {error, Error} ->
             {error, Error}
     end.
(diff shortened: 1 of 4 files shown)


src/erlcloud_cloudformation.erl
```diff
@@ -659,6 +683,113 @@ extract_account_limit(XmlNode) ->
...
+                FieldFun(Parameter)
+            ) | Acc]
+        end,
+        [],
+        List
+    ));
+member_object_array(_Prefix, _Fun, _) ->
+    [].
+
+member_list(Prefix, List) when is_list(List) ->
```
kkuzmin: we can probably reuse `erlcloud_util:encode_list()`.
And also worth moving the rest of `member...()` function into `erlcloud_util.erl`

src/erlcloud_cloudformation.erl
```diff
@@ -659,6 +725,137 @@ extract_account_limit(XmlNode) ->
...
+        { "EnableTerminationProtection", Spec#cloudformation_create_stack_input.enable_termination_protection },
+        { "OnFailure", Spec#cloudformation_create_stack_input.on_failure },
+        { "RoleARN", Spec#cloudformation_create_stack_input.role_arn },
+        { "StackName", Spec#cloudformation_create_stack_input.stack_name },
+        { "StackPolicyBody", Spec#cloudformation_create_stack_input.stack_policy_body },
+        { "StackPolicyURL", Spec#cloudformation_create_stack_input.stack_policy_url },
+        { "TemplateBody", Spec#cloudformation_create_stack_input.template_body },
+        { "TemplateURL", Spec#cloudformation_create_stack_input.template_url },
+        { "TimeoutInMinutes", Spec#cloudformation_create_stack_input.timeout_in_minutes }
+    ].
```
kkuzmin: Do we need to call `filter_undefined()` here as well?

src/erlcloud_cloudformation.erl
daniel-greening: undefined get filtered out here https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_aws.erl#L1055 we just need to worry about filtering out nested fields like parameters, capabilities etc

src/erlcloud_cloudformation.erl
daniel-greening: undefined get filtered out here https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_aws.erl#L1055 we just need to worry about filtering out nested fields like parameters, capabilities etc

PR #627, opened 2019-12-31
DynamoDB: do not retry requests unless it's safe

g-andrade:
Otherwise, non-idempotent operations like counter increments may result in unwanted side effects when retried.

Besides that, avoid retrying upon protocol, socket or network errors that are very unlikely to see recovery in the short term, like failed SSL negotiation (as was suggested in a previous TODO comment, now removed.)

I haven't yet had the chance to fully test the changes, but they pass the unit tests.

### DIFF
--- a/src/erlcloud_ddb1.erl
+++ b/src/erlcloud_ddb1.erl
@@ -14,6 +14,13 @@
 -include("erlcloud.hrl").
 -include("erlcloud_aws.hrl").
 
+-ifdef(TEST).
+% Ensure we always evaluate the checks when running the test cases.
+-define(COSTLY_IDEMPOTENCY_CHECK(Check), (Check)).
+-else.
+-define(COSTLY_IDEMPOTENCY_CHECK(Check), (fun () -> (Check) end)).
+-endif.
+
 %% DDB API Functions
 -export([batch_get_item/1, batch_get_item/2,
          batch_write_item/1, batch_write_item/2,
@@ -93,7 +100,7 @@ batch_get_item(RequestItems) ->
 -spec batch_get_item([batch_get_item_request_item()], aws_config()) -> json_return().
 batch_get_item(RequestItems, Config) ->
     Json = [{<<"RequestItems">>, [batch_get_item_request_item_json(R) || R <- RequestItems]}],
-    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.BatchGetItem", Json).
+    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.BatchGetItem", Json, [idempotent]).
 
 -type batch_write_item_put() :: {put, item()}.
 -type batch_write_item_delete() :: {delete, key()}.
@@ -117,7 +124,7 @@ batch_write_item(RequestItems) ->
 -spec batch_write_item([batch_write_item_request_item()], aws_config()) -> json_return().
 batch_write_item(RequestItems, Config) ->
     Json = [{<<"RequestItems">>, [batch_write_item_request_item_json(R) || R <- RequestItems]}],
-    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.BatchWriteItem", Json).
+    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.BatchWriteItem", Json, [idempotent]).
 
 
 -spec key_schema_value_json(key_schema_value()) -> jsx:json_term().
@@ -141,7 +148,7 @@ create_table(Table, KeySchema, ReadUnits, WriteUnits, Config) ->
             key_schema_json(KeySchema),
             {<<"ProvisionedThroughput">>, [{<<"ReadCapacityUnits">>, ReadUnits},
                                            {<<"WriteCapacityUnits">>, WriteUnits}]}],
-    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.CreateTable", Json).
+    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.CreateTable", Json, [{idempotent, false}]).
 
     
 -spec delete_item(table_name(), key()) -> json_return().
@@ -157,7 +164,8 @@ delete_item(Table, Key, Opts, Config) ->
     Json = [{<<"TableName">>, Table},
             key_json(Key)] 
         ++ Opts,
-    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.DeleteItem", Json).
+    IsIdempotent = do_opts_keep_idempotency(Opts),
+    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.DeleteItem", Json, [{idempotent, IsIdempotent}]).
 
     
 -spec delete_table(table_name()) -> json_return().
@@ -167,7 +175,7 @@ delete_table(Table) ->
 -spec delete_table(table_name(), aws_config()) -> json_return().
 delete_table(Table, Config) ->
     Json = [{<<"TableName">>, Table}],
-    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.DeleteTable", Json).
+    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.DeleteTable", Json, [{idempotent, false}]).
 
     
 -spec describe_table(table_name()) -> json_return().
@@ -177,7 +185,7 @@ describe_table(Table) ->
 -spec describe_table(table_name(), aws_config()) -> json_return().
 describe_table(Table, Config) ->
     Json = [{<<"TableName">>, Table}],
-    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.DescribeTable", Json).
+    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.DescribeTable", Json, [idempotent]).
 
 
 -spec get_item(table_name(), key()) -> json_return().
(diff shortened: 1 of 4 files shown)


g-andrade:
(Bumping this PR in case it might have been missed/forgotten due to holiday season, and since #628, opened yesterday, might conflict with this.)

src/erlcloud_ddb1.erl
```diff
@@ -141,7 +148,7 @@ create_table(Table, KeySchema, ReadUnits, WriteUnits, Config) ->
             key_schema_json(KeySchema),
             {<<"ProvisionedThroughput">>, [{<<"ReadCapacityUnits">>, ReadUnits},
                                            {<<"WriteCapacityUnits">>, WriteUnits}]}],
-    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.CreateTable", Json).
+    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.CreateTable", Json, [{idempotent, false}]).
```
nalundgaard: isn't this idempotent?

src/erlcloud_ddb1.erl
```diff
@@ -167,7 +175,7 @@ delete_table(Table) ->
 -spec delete_table(table_name(), aws_config()) -> json_return().
 delete_table(Table, Config) ->
     Json = [{<<"TableName">>, Table}],
-    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.DeleteTable", Json).
+    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.DeleteTable", Json, [{idempotent, false}]).
```
nalundgaard: Why isn't this idempotent?

src/erlcloud_ddb1.erl
```diff
@@ -302,6 +313,42 @@ retry(Attempt, Reason) when Attempt >= ?NUM_ATTEMPTS ->
...
+do_opts_keep_idempotency(OptsMap) when is_map(OptsMap) ->
+    Opts = maps:to_list(OptsMap),
+    do_opts_keep_idempotency(Opts);
+do_opts_keep_idempotency([{}]) ->
+    true;
+do_opts_keep_idempotency(Opts) ->
+    % * https://docs.amazonaws.cn/en_us/amazondynamodb/latest/developerguide/API_DeleteItem_v20111205.html#API_DeleteItem_RequestParameters
+    % * https://docs.amazonaws.cn/en_us/amazondynamodb/latest/developerguide/API_PutItem_v20111205.html#API_PutItem_RequestParameters
+    % * https://docs.amazonaws.cn/en_us/amazondynamodb/latest/developerguide/API_UpdateItem_v20111205.html#API_UpdateItem_RequestParameters
+    not lists:keymember(<<"Expected">>, 1, Opts).
```
nalundgaard: This seems wrong. wouldn't setting an expectation (condition) on a request make it more likely to be idempotent?

src/erlcloud_ddb2.erl
```diff
@@ -2792,15 +2830,28 @@ put_item(Table, Item, Opts) ->
...
         #ddb2_put_item.attributes, {ok, []}).
 
+-spec do_put_item_opts_keep_idempotency(put_item_opts()) -> boolean().
+do_put_item_opts_keep_idempotency(Opts) ->
+    lists:all(fun does_put_item_opt_keep_idempotency/1, Opts).
+
+-spec does_put_item_opt_keep_idempotency(put_item_opt()) -> boolean().
+does_put_item_opt_keep_idempotency({OptName, _}) ->
+    % https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_PutItem.html
+    not lists:member(OptName, [condition_expression,
```
nalundgaard: again, here, it seems like conditions make a request more likely to be idempotent.

src/erlcloud_ddb1.erl
g-andrade: Well, not really; if for some reason a previous attempt got to the other end but the response failed to arrive / timed out, should you retry? 
It's not like there's any harm in retrying within a short time window, sure (the table will still be in the state of being created, or even already created) - but getting an "immediate" response along the lines of "Table being created" or "Table already created" sounds, to me, more confusing than getting a network error, retrying, and getting that same error - one can more easily conclude from the latter the reason for the failure is that the request succeeded, only you didn't get the response telling you so on the first attempt.
So, my intention here is mostly about simplifying the interpretation of the response. But you're right in the sense that there's no harm in retrying - for plausible scenarios, anyway.

src/erlcloud_ddb1.erl
g-andrade: Same reasoning as for `CreateTable` :-)

src/erlcloud_ddb1.erl
g-andrade: It safeguards you against committing unwanted changes, of course, but it's still a bit weird to get an "expectation failed" sort of error because of a retry - wouldn't it be better to spit the network error back to the consumer, and let them decide? Otherwise it's much more obscure whether to say if the expectation failed because your initial assumptions are wrong (say, because your in-memory cache of the data is outdated) rather than because there was a retry.

src/erlcloud_ddb2.erl
g-andrade: Same rationale as for the `Expected` condition I previously commented :-)

g-andrade:
Your reply gave me some food for thought - so I'll describe the concrete problem I'm trying to solve (maybe I should have done so to start with.)
I work with (and maintain) some systems that deal with DynamoDB using `erlcoud`; now, in the past I've noticed some non-idempotent operations either unexpectedly failing due to violated constraints that really shouldn't happen or (worse!) silently running update expressions more than once, like those involving counter increments.
Well, my first thought once I looked into the retry strategy was - well, screw that, let's just not do retries, right? After all, it doesn't make any sense, it's a freaking database, what were those guys thinking?
I tried not doing retries, and was quickly shown why retries are performed by default - DynamoDB fails _all the time_. Like, random, unexpected HTTP 5xx errors. Well, there's a lesson in humility. But surely we can limit retries to cases whether it's safe to do so?
And that's what got me to this PR - it's still very far from ideal in my mind, since `lhttpc` is rather shy about telling why it timed out in some cases. But it still looks to me like an improvement in general - if you don't know whether retrying is safe, spit the error back to the consumer and let them decide for themselves. It just seems safer as a default strategy.
What do you think?

g-andrade:
> I'd be a lot more open to just making a common DDB option to allow clients to specify that their requests are idempotent or not, with some documentation suggesting when that might be an issue. 

I'm all up for that, especially since some cases are very hard to cover (like determining the idempotency of an update expression, which under my changes would not be considered safe, even if it actually is.)

> It really needs to be up to the client to give this data more explicitly, and I would be open on ideas for elegant ways to do that, but failing that, I think the easiest thing is to just set the retry count to 0 in the aws_config for non-idempotent requests.

Well, I partly agree, while adding that the best option would be to not allow for retries by default unless the consumer says it's safe. But most users will not be thinking about that, because, frankly, it's a pain in the ass - to describe my own case, I didn't consider this particular problem for _years_.
A random user just wants to read from and write to a database - it's a uncanny that _every so often_ their updates will be applied twice (or thrice, or ...). 
So I think that making for a safer but usable default it's best, while optimizing for the cases where retries can be performed as to reduce the impact of DynamoDB temperament.

src/erlcloud_ddb1.erl
nalundgaard: Indeed, there are 2 forms of idempotency here.

CreateTable is idempotent in the sense that you can call it as many times as you like, it will only create it once. In the response sense, it is very much not idempotent—the response when the table exists is an exception which the caller happens.

On the other hand, I very much think the library should let this happen, because you will have to address that error one way or the other, and TBH we have a pretty 'sophisticated' library internally that uses create_table as-is and does exactly this. I think it's robust enough to deal with your library change, but I'm not sure that it's that good in this case.

One of the terrible aspects of this API, for stuff like CreateTable (and even more so, the global table functions), is that the default "first request" timeout is 1 second, unless you explicitly set the timeout in the configuration. @rmpalomino is dealing with that with the November 2019 global tables API updates right now.

src/erlcloud_ddb1.erl
nalundgaard: Fair enough.

nalundgaard:
Maybe it would be better to make this PR a little narrower in scope—only setting `[{idempotent, false}]` for cases that are almost certainly not idempotent and leaving it `unknown` for most cases (for clarity in reading the code if nothing else). Having said that, though, I think this feature is unacceptable unless it can be explicitly disabled/overridden by the client, either by an aws_config property or by options in the erlcloud_ddb* functions where it might be `false` and prevent the automatic retries the client wants to be able to take advantage of.

@motobob one kind of wonders if this feature shouldn't be more generic—I am sure DynamoDB isn't the only AWS API that doesn't have idempotency issues that are snarled by generic retry logic.

src/erlcloud_ddb1.erl
g-andrade: > One of the terrible aspects of this API, for stuff like CreateTable (and even more so, the global table functions), is that the default "first request" timeout is 1 second [...]

Wow, I didn't know that! Yes, it looks way too low.

g-andrade:
> only setting [{idempotent, false}] for cases that are almost certainly not idempotent and leaving it unknown for most cases (for clarity in reading the code if nothing else)

But should the default behaviour, upon `lhttpc` errors, be kept when the idempotency is unknown? (That default being "perform retry.")

> Having said that, though, I think this feature is unacceptable unless it can be explicitly disabled/overridden by the client

Fair enough; I'll give it another go when I find the chance.

g-andrade:
Mmmh, I don't think I'll come back to this anytime soon. 

Closing the issue but I'll leave the branch around in case anyone wishes to pick it up.

PR #630, opened 2020-01-30
Get rid of "term() | no_return()" specs

### LINKED ISSUE #559: Fix dialyzer specs
In several places (erlcloud_s3 in particular, maybe somewhere else) there are typespecs that look like:

`-spec foo() -> something() | no_return().`

In success typing they don't make sense [1] [2], as `none()` type indicates that the function _never_ returns a value (e.g. loops forever, _always_ crashes or shuts down BEAM emulator). Annotating functions that have at least one code path returning a term with `no_return()` 1) can hide genuine dialyzer errors 2) is viral, as all code calling the function becomes `no_return()`, hiding more errors 

[1] http://erlang.org/doc/reference_manual/typespec.html
[2] https://learnyousomeerlang.com/dialyzer

I can create a PR to clean this stuff up.

k32:
Fixes #559

### DIFF
--- a/src/erlcloud_elb.erl
+++ b/src/erlcloud_elb.erl
@@ -40,13 +40,13 @@
 -define(DEFAULT_MAX_RECORDS, 400).
 
 % xpath for elb descriptions used in describe_groups functions:
--define(DESCRIBE_ELBS_PATH, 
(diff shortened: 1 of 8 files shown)


k32:
Thanks!

PR #632, opened 2020-02-13
Properly encode SQS queue name path

dsrosario:
localstack queue url is in the form of
http://localhost:4576/queue/<queue name>. In order to use erlcloud
with localstack, queue path should not urlencode '/' if this is
present at the QueueName.


dsrosario:
found out that I can use the queue url directly at receive_message in order to cope with localstack.

PR #633, opened 2020-02-18
Fixes for dialyzer (all warnings)

paulo-ferraz-oliveira:
I pulled `master`, ran `make check` and fixed all dialyzer issues.

I'm comfortable with all commits except for the last one, 9c93a76, where I explicitly hand pick certain `dialyzer` warnings to have them ignored, for the sake of a "clean" run.

This means that if these functions are changed, in the future, to something that might cause those issues, they will still be ignored (they are all caused by invalid - as per documentation - calls to `application:`).

In the end, I ran `make eunit` which caused the "Revert previous change..." commit.

### DIFF
--- a/src/erlcloud_application_autoscaler.erl
+++ b/src/erlcloud_application_autoscaler.erl
@@ -58,20 +58,20 @@
 %% Type Definitions
 %% ------------------------------------------------------------------
 
--type response_attribute() :: binary().
+-type response_attribute() :: string() | integer().
 -type response_key() :: atom().
--type response() :: [{response_key(), response_attribute()} | proplist:proplist()].
+-type response() :: [{response_key(), response_attribute()}].
 
 -type aws_aas_request_body() :: [proplist:proplist()].
 
--spec extract_alarm(J :: binary()) -> response().
+-spec extract_alarm(J :: proplist()) -> response().
 extract_alarm(J) ->
     erlcloud_json:decode([
         {alarm_name, <<"AlarmName">>, optional_string},
         {alarm_arn, <<"AlarmARN">>, optional_string}
     ], J).
 
--spec extract_step_adjustments(J :: binary) -> response().
+-spec extract_step_adjustments(J :: proplist()) -> response().
 extract_step_adjustments(J) ->
     erlcloud_json:decode([
         {metric_interval_lower_bound, <<"MetricIntervalLowerBound">>, optional_integer},
@@ -79,7 +79,7 @@ extract_step_adjustments(J) ->
         {scaling_adjustment, <<"ScalingAdjustment">>, optional_integer}
     ], J).
 
--spec extract_step_scaling_policy(J :: binary) -> response().
+-spec extract_step_scaling_policy(J :: proplist()) -> response().
 extract_step_scaling_policy(J) ->
     Scaling = erlcloud_json:decode([
         {adjustment_type, <<"AdjustmentType">>, optional_string},
@@ -94,21 +94,21 @@ extract_step_scaling_policy(J) ->
             [{step_adjustments,  [ extract_step_adjustments(Step) || Step <- Steps]} | Scaling]
     end.
 
--spec extract_dimensions(J :: binary) -> response().
+-spec extract_dimensions(J :: proplist()) -> response().
 extract_dimensions(J) ->
     erlcloud_json:decode([
         {name, <<"Name">>, optional_string},
         {value, <<"Value">>, optional_string}
     ], J).
 
--spec extract_predefined_metric_specifications(J :: binary) -> response().
+-spec extract_predefined_metric_specifications(J :: proplist()) -> response().
 extract_predefined_metric_specifications(J) ->
     erlcloud_json:decode([
         {predefined_metric_type, <<"PredefinedMetricType">>, optional_string},
         {resource_label, <<"ResourceLabel">>, optional_string}
     ], J).
 
--spec extract_customized_metric_specification(J :: binary) -> response().
+-spec extract_customized_metric_specification(J :: proplist()) -> response().
 extract_customized_metric_specification(J) ->
     CustomizedMetricSpecification = erlcloud_json:decode([
         {metric_name, <<"MetricName">>, optional_string},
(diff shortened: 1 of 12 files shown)


src/erlcloud_application_autoscaler.erl
```diff
@@ -318,7 +318,9 @@ delete_scaling_policy(Configuration, BodyConfiguration) ->
                             ScalableDimension :: binary(),
                             ScheduledActionName :: binary(),
                             ServiceNamespace :: binary()
-                        ) -> response().
+                        ) -> {ok, jsx:json_term()} | {error, metadata_not_available
```
nalundgaard: seems like it would save a lot of copy-pasta to coin an `application_autoscaler_response()` type, no?

paulo-ferraz-oliveira: It would. Do you prefer it?

**Note**: there was already a `response()` type defined, and I didn't want to introduce further ambiguity.

src/erlcloud_ddb2.erl
```diff
@@ -3677,7 +3678,6 @@ transact_write_items(TransactItems, Opts, Config) ->
     case out(Return,
              fun(Json, UOpts) -> undynamize_record(transact_write_items_record(), Json, UOpts) end, DdbOpts,
              #ddb2_transact_write_items.attributes, {ok, []}) of
-        {simple, Record} -> {ok, Record};
```
nalundgaard: is this right?

paulo-ferraz-oliveira: It's further explained in commit 078fae2's message: 

> `out/5` never returns `{simple, _}` which is why it was removed from the `case` in `transact_write_items/3`

That function is even `-spec(_)`ed to return `-> ok_return(term())`.

src/erlcloud_states.erl
```diff
@@ -539,7 +539,7 @@ start_execution(StateMachineArn, Options) ->
     start_execution(StateMachineArn, Options, default_config()).
 
 -spec start_execution(StateMachineArn   :: binary(),
-                      Options           :: list(),
+                      Options           :: list() | map(),
```
nalundgaard: It looks like it's just "map" only, no?

src/erlcloud_states.erl
```diff
@@ -573,7 +573,7 @@ stop_execution(ExecutionArn, Options) ->
     stop_execution(ExecutionArn, Options, default_config()).
 
 -spec stop_execution(ExecutionArn   :: binary(),
-                     Options           :: list(),
+                     Options           :: list() | map(),
```
nalundgaard: Same here, `list()` looks like just a bad type that should be removed

src/erlcloud_guardduty.erl
```diff
@@ -123,11 +123,7 @@ guardduty_request_no_update(Config, Method, Path, Body, QParam) ->
     end.
 
 encode_body(undefined) ->
-    <<>>;
```
nalundgaard: This removal seems iffy. Why?

src/erlcloud_states.erl
paulo-ferraz-oliveira: Yup, you're right. I saw the function was exported and though I didn't want to break the external interface, but it does guard against maps, though.

src/erlcloud_states.erl
paulo-ferraz-oliveira: Yes

src/erlcloud_ddb2.erl
nalundgaard: Perfect, thanks. Sorry, was looking at the unified view.

src/erlcloud_guardduty.erl
paulo-ferraz-oliveira: It's explained in commit eb0e45a's message:

> `encode_body/1`
> is only ever called locally via
> `guardduty_request_no_update/5`
> 
> `guardduty_request_no_update/5`
> is only ever called locally via
> `guardduty_request/5`
> 
> `guardduty_request/5`
> is only ever called locally with
> `Body =:= undefined`
> 
> (maybe it was intended, initially, to cover more use cases, but undefined ended up being the only one present)

It also took me a while to find out why `dialyzer` was complaining.

src/erlcloud_states.erl
paulo-ferraz-oliveira: Fixed in 067807c.

src/erlcloud_application_autoscaler.erl
nalundgaard: Good point, what about an `ok_response()` type to distinguish the response format from the ones that return naked values and raise exceptions on failure?

src/erlcloud_states.erl
paulo-ferraz-oliveira: Fixed in 067807c.

src/erlcloud_guardduty.erl
nalundgaard: Interesting. I guess it probably was planned for some future use. I guess that's fine with me.

src/erlcloud_application_autoscaler.erl
paulo-ferraz-oliveira: You mean turn some of those `response()`s into `ok_response()` and then have `application_autoscaler_response() :: ok_response() | {error, _}`?

nalundgaard: I mean something more like 

```erlang
-type ok_response() :: {ok, jsx:json_term()}
                     | {error, metadata_not_available
                             | container_credentials_unavailable
                             | erlcloud_aws:httpc_result_error()}.
```

maybe better to call it `ok_error_response()`? I don't care too much about naming, I'd just prefer to avoid copying the same ok/error response spec everywhere.

paulo-ferraz-oliveira: Sure, I can do that.

paulo-ferraz-oliveira: Fixed in 7b52338.

paulo-ferraz-oliveira:
Sure, it would be nice to improve the CI checks with clean dialyzer. Want me to add it to this PR? Or will you do it after the merge?

I think I handled all your comments 😄

nalundgaard:
on incorporating into CI, I think it would be fine to include it in this PR, but I would wait for @motobob's agreement before making the change. What do you think, @motobob?

motobob:
i'm fine to put it in. we seems to have enough of users to care about it

motobob:
i suggest merging and sending a follow on dializer ci check integration

nalundgaard:
👍, merging it then.

nalundgaard:
I'll hold off on cutting a new tag/hex package; I think we'll have some additional new change soon in #634.

PR #634, opened 2020-03-03
Extend VPC description with cidr_block_association_set

alexo-spb:
#### Problem
VPC description has missing information about CIDR block associations. Amazon VPC allows customers to expand their VPCs by adding secondary IPv4 address ranges to their VPCs. https://aws.amazon.com/about-aws/whats-new/2017/08/amazon-virtual-private-cloud-vpc-now-allows-customers-to-expand-their-existing-vpcs/

#### Solution
Extract the information about CIDR block associations based on https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-vpcs.html

### DIFF
--- a/src/erlcloud_ec2.erl
+++ b/src/erlcloud_ec2.erl
@@ -2459,10 +2459,30 @@ extract_vpc(Node) ->
       {dhcp_options_id, get_text("dhcpOptionsId", Node)},
       {instance_tenancy, get_text("instanceTenancy", Node)},
       {is_default, get_bool("isDefault", Node)},
-      {tag_set, 
+      {cidr_block_association_set, extract_cidr_block_association_set(Node)},
+      {tag_set,
         [extract_tag_item(Item)
          || Item <- xmerl_xpath:string("tagSet/item", Node)]}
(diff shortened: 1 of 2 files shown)


src/erlcloud_ec2.erl
```diff
@@ -2459,10 +2459,30 @@ extract_vpc(Node) ->
...
+    Items = xmerl_xpath:string("cidrBlockAssociationSet/item", Node),
+    [extract_cidr_block_association_item(Item) || Item <- Items].
+
+extract_cidr_block_association_item(Node) ->
+    [
+         {cidr_block, get_text("cidrBlock", Node)},
+         {association_id, get_text("associationId", Node)},
+         {cidr_block_state,
+            [{state, get_text("cidrBlockState/state", Node)}] ++
+            case get_text("cidrBlockState/statusMessage", Node, undefined) of
```
nalundgaard: Can you add example with a statusMessage into the unit tests?

alexo-spb: @nalundgaard the example has been updated

PR #636, opened 2020-04-06
More kinesis headers

codyf221:
Some InternalFailures are being seen from kinesis. Initial investigation targeted 400-499 response headers, but that is not the case we are in . 
Log more response headers from kinesis to find InternalFailure request ids for AWS

### DIFF
--- a/src/erlcloud_kinesis_impl.erl
+++ b/src/erlcloud_kinesis_impl.erl
@@ -129,13 +129,16 @@ request_and_retry(Config, Headers, Body, ShouldDecode, {attempt, Attempt}) ->
                     {error, Reason}
             end;
 
-        {ok, {{Status, StatusLine}, _, RespBody}} when Status >= 500 ->
+        {ok, {{Status, StatusLine}, RespHeaders, RespBody}} when Status >= 500 ->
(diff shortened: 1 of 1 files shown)


codyf221:
Wrong location for this PR

src/erlcloud_kinesis_impl.erl
```diff
@@ -113,28 +113,32 @@ request_and_retry(Config, Headers, Body, ShouldDecode, {attempt, Attempt}) ->
...
+        {ok, {{200, _}, RespHeaders, RespBody}} ->
             Result = case ShouldDecode of
                          true  -> decode(RespBody);
                          false -> RespBody
                      end,
             {ok, Result};
 
-        {ok, {{Status, StatusLine}, _, RespBody}} when Status >= 400 andalso Status < 500 ->
+        {ok, {{Status, StatusLine}, RespHeaders, RespBody}} when Status >= 400 andalso Status < 500 ->
+            error_logger:error_msg("Kinesis response headers 400-499 : ~p", [RespHeaders]),
```
motobob, line 124: in spite it's closed, do not log from the library directly make top level app do logging.

PR #637, opened 2020-04-13
allow using configured VPCe

motobob:
problem: interface VPCe do not mitigate cross-AZ traffic in EC2/ECS environments.

solution: allow top level app to configure it. 
deliberately not doing ecs2 describe-vpe-endpoints nor ip magic in /etc/hosts since it has it's flaws with large number of instances. 
instead configure it ones at startup on application level from application env

cc @kkuzmin @nalundgaard 

might add add helper function to configure that via AZ.

### DIFF
--- a/src/erlcloud.app.src
+++ b/src/erlcloud.app.src
@@ -19,6 +19,7 @@
                   lhttpc]},
   {modules, []},
   {env, [
+      % Example [{<<"kinesis">>, <<"myAZdnsentry.amazonaws.com">>}]
       {services_vpc_endpoints, []}
   ]},
   {licenses, ["MIT"]},
--- a/src/erlcloud.app.src
+++ b/src/erlcloud.app.src
@@ -18,7 +18,9 @@
                   %% hackney,
                   lhttpc]},
   {modules, []},
-  {env, []},
+  {env, [
+      {services_vpc_endpoints, []}
+  ]},
   {licenses, ["MIT"]},
   {links, [{"Github", "https://github.com/erlcloud/erlcloud"}]}
  ]
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -14,7 +14,7 @@
          default_config_region/2, default_config_override/1,
          update_config/1,clear_config/1, clear_expired_configs/0,
          service_config/3, service_host/2,
-         configure_vpce/2, get_vpce/0,
+         configure_vpc_endpoints/2, get_vpc_endpoints/0,
          configure/1, format_timestamp/1,
          http_headers_body/1,
          http_body/1,
@@ -812,9 +812,9 @@ service_host( Service, Region ) when is_binary(Service) ->
         application:get_env(erlcloud, services_vpc_endpoints, []),
             Default).
 
--spec configure_vpce(binary(), list(binary())) -> ok.
+-spec configure_vpc_endpoints(binary(), list(binary())) -> ok.
 % take the list of possible endpoints and pick the one suites from our AZ.
-configure_vpce(Service, Endpoints) when is_binary(Service) ->
+configure_vpc_endpoints(Service, Endpoints) when is_binary(Service) ->
     %% default config is fine. no IAM is used
     case erlcloud_ec2_meta:get_instance_metadata("placement/availability-zone", erlcloud_aws:default_config()) of
         {ok, AZ} ->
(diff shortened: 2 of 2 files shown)


src/erlcloud_aws.erl
```diff
@@ -796,14 +797,41 @@ service_host( <<"s3">>, Region ) ->
...
 service_host( Service, <<"cn-north-1">> = Region ) when is_binary(Service) ->
     binary_to_list( <<Service/binary, $., Region/binary, ".amazonaws.com.cn">> );
 service_host( Service, <<"cn-northwest-1">> =Region ) when is_binary(Service) ->
     binary_to_list( <<Service/binary, $., Region/binary, ".amazonaws.com.cn">> );
+% services which can have VPCe configured to mitigate cross-AZ traffic
+% it's application level decision to use VPCe and to match AZ and pick correct DNS name
+% if configured use it, otherwise use default
+% magic can be done vi EC3 describe VpcEndpoints/filter by VPC/filter by AZ,
+% however permissions and describe* API throttling 
 service_host( Service, Region ) when is_binary(Service) ->
```
kkuzmin, line 809: please add `is_binary(Region)` guard

src/erlcloud.app.src
```diff
@@ -18,7 +18,9 @@
                   %% hackney,
                   lhttpc]},
   {modules, []},
-  {env, []},
+  {env, [
+      {services_vpc_endpoints, []}
```
kkuzmin, line 22: please add an example of one VPC endpoint entry.

src/erlcloud_aws.erl
```diff
@@ -796,14 +797,41 @@ service_host( <<"s3">>, Region ) ->
...
-    binary_to_list( <<Service/binary, $., Region/binary, ".amazonaws.com">> ).
+    Default = binary_to_list( <<Service/binary, $., Region/binary, ".amazonaws.com">> ),
+    proplists:get_value(Service,
+        application:get_env(erlcloud, services_vpc_endpoints, []),
+            Default).
+
+-spec configure_vpce(binary(), list(binary())) -> ok.
+% take the list of possible endpoints and pick the one suites from our AZ.
+configure_vpce(Service, Endpoints) when is_binary(Service) ->
```
kkuzmin, line 817: feels like this function belongs to `erlcloud_ec2.erl` or `erllcoud_util.erl`

src/erlcloud_aws.erl
```diff
@@ -14,6 +14,7 @@
          default_config_region/2, default_config_override/1,
          update_config/1,clear_config/1, clear_expired_configs/0,
          service_config/3, service_host/2,
+         configure_vpce/2, get_vpce/0,
```
kkuzmin, line 17: I'd prefer full names `configure_vpc_endpoints`

src/erlcloud_aws.erl
```diff
@@ -796,14 +797,41 @@ service_host( <<"s3">>, Region ) ->
...
+% magic can be done vi EC3 describe VpcEndpoints/filter by VPC/filter by AZ,
+% however permissions and describe* API throttling 
 service_host( Service, Region ) when is_binary(Service) ->
-    binary_to_list( <<Service/binary, $., Region/binary, ".amazonaws.com">> ).
+    Default = binary_to_list( <<Service/binary, $., Region/binary, ".amazonaws.com">> ),
+    proplists:get_value(Service,
+        application:get_env(erlcloud, services_vpc_endpoints, []),
+            Default).
+
+-spec configure_vpce(binary(), list(binary())) -> ok.
```
kkuzmin, line 815: pleas fix the spec.

src/erlcloud_aws.erl
motobob: not really. i would like to keep it nearby the the thing is used

src/erlcloud.app.src
motobob: added ^

src/erlcloud_aws.erl
```diff
@@ -796,14 +797,41 @@ service_host( <<"s3">>, Region ) ->
...
+
+-spec configure_vpc_endpoints(binary(), list(binary())) -> ok | {error, httpc_result_error()}.
+% take the list of possible endpoints and pick the one suites from our AZ.
+configure_vpc_endpoints(Service, Endpoints) when is_binary(Service) ->
+    %% default config is fine. no IAM is used
+    case erlcloud_ec2_meta:get_instance_metadata("placement/availability-zone", default_config()) of
+        {ok, AZ} ->
+            case [{binary:match(AZ, E), E} || E <- Endpoints, is_binary(E)] of
+                [] ->  ok;
+                [{_, Endpoint}] ->
```
kkuzmin, line 823: Looks like we need to check if `binary:match()` returns `nomatch`.
What if there more than 1 endpoint matching AZ? Is it possible?
Lets also add default case clause here and return an error

motobob: fixed. and made it pick the first one, even though there could only one

motobob:
let me rework the setting from ENV to simplify it even further

motobob:
w

PR #638, opened 2020-04-13
allow to use pre-configured VPC endpoints

motobob:
problem: interface VPCe do not mitigate cross-AZ traffic in EC2/ECS environments.
original #637 
problem: some time VPCe need to be using in single AZ mode and not to pay for cross-AZ overhead. 

solution: allow top level app to configure it.
deliberately **not** doing ecs2 describe-vpe-endpoints nor ip magic in /etc/hosts since it has it's flaws with large number of instances.
instead configure it ones at startup on application level from application env

cc @kkuzmin @nalundgaard

might add add helper function to configure that via AZ.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,8 +157,24 @@ Example of such app.config:
 ].
 ```
 
+### VPC endpoints
+If you want to utilise AZ affinity for VPC endpoints you can configure those in application config via:
+```erlang
+{erlcloud, [
+    {services_vpc_endpoints, [
+        {<<"sqs">>, [<<"myAZ1.sqs-dns.amazonaws.com">>, <<"myAZ2.sqs-dns.amazonaws.com">>]},
+        {<<"kinesis">>, {env, "KINESIS_VPC_ENDPOINTS"}}
+    ]}
+]}
+```
+Two options supported:
+ - explicit lists of route53 AZ endpoints
+ - OS environment variable, comes handy for ECS deployments.
+ 
+Upon config generation, `erlcloud` will check the AZ of the deployment and match it to one of the pre-configured DNS records to use.
 
-### Basic use ###
+
+## Basic use ##
 Then you can start making API calls, like:
 ```
 erlcloud_ec2:describe_images().
--- a/README.md
+++ b/README.md
@@ -169,9 +169,9 @@ If you want to utilise AZ affinity for VPC endpoints you can configure those in 
 ```
 Two options supported:
  - explicit lists of route53 AZ endpoints
- - OS environment variable, comes handy for ECS deployments.
+ - OS environment variable, comes handy for ECS deployments. Env should be of comma separated string like:`"myAZ1.sqs-dns.amazonaws.com,myAZ2.sqs-dns.amazonaws.com"`
  
-Upon config generation, `erlcloud` will check the AZ of the deployment and match it to one of the pre-configured DNS records to use.
+Upon config generation, `erlcloud` will check the AZ of the deployment and match it to the first one of the pre-configured DNS records to use.
 
 
 ## Basic use ##
--- a/README.md
+++ b/README.md
@@ -171,7 +171,8 @@ Two options supported:
  - explicit lists of route53 AZ endpoints
  - OS environment variable, comes handy for ECS deployments. Env should be of comma separated string like:`"myAZ1.sqs-dns.amazonaws.com,myAZ2.sqs-dns.amazonaws.com"`
  
-Upon config generation, `erlcloud` will check the AZ of the deployment and match it to the first one of the pre-configured DNS records to use.
+Upon config generation, `erlcloud` will check the AZ of the deployment 
+and match it to one of the pre-configured DNS records. First match is used and if not match found default is used. 
 
 
 ## Basic use ##
--- a/src/erlcloud.app.src
+++ b/src/erlcloud.app.src
@@ -18,7 +18,11 @@
                   %% hackney,
                   lhttpc]},
   {modules, []},
-  {env, []},
+  {env, [
+      % Example   [{<<"kinesis">>, [<<"myAZ1.amazonaws.com">>, <<"myAZ2.amazonaws.com">>]}]
+      % or vi ENV [{<<"kinesis">>, {env, "KINESIS_VPC_ENDPOINTS"}]
+      {services_vpc_endpoints, []}
+  ]},
   {licenses, ["MIT"]},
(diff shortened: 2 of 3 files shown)


src/erlcloud.app.src
```diff
@@ -18,7 +18,11 @@
                   %% hackney,
                   lhttpc]},
   {modules, []},
-  {env, []},
+  {env, [
+      % Example   [{<<"kinesis">>, [<<"myAZ1.amazonaws.com">>, <<"myAZ2.amazonaws.com">>]}]
+      % or vi ENV [{<<"kinesis">>, {env, "KINESIS_VPC_ENDPOINTS"}]
```
kkuzmin, line 23: typo, `via`

src/erlcloud_aws.erl
```diff
@@ -796,14 +797,55 @@ service_host( <<"s3">>, Region ) ->
...
+    %% resolve through ENV if any
+    Endpoints = case ConfiguredEndpoints of
+        {env, EnvVarName} when is_list(EnvVarName) ->
+            Es = string:split(os:getenv(EnvVarName, ""), ",", all),
+            [list_to_binary(E) || E <- Es];
+        EndpointsList when is_list(EndpointsList) ->
+            EndpointsList
+    end,
+    % now get our AZ and match
+    case erlcloud_ec2_meta:get_instance_metadata("placement/availability-zone", default_config()) of
```
kkuzmin, line 827: let's avoid calling instance metadata if endpoints are not configured.

src/erlcloud.app.src
```diff
@@ -18,7 +18,11 @@
                   %% hackney,
                   lhttpc]},
   {modules, []},
-  {env, []},
+  {env, [
+      % Example   [{<<"kinesis">>, [<<"myAZ1.amazonaws.com">>, <<"myAZ2.amazonaws.com">>]}]
```
nalundgaard, line 22: Seems like this could/should really have a brief section in the README

src/erlcloud_aws.erl
```diff
@@ -796,14 +797,59 @@ service_host( <<"s3">>, Region ) ->
...
+% however permissions and describe* API throttling is not what we want to deal with here.
+service_host( Service, Region ) when is_binary(Service) andalso is_binary(Region) ->
+    Default = binary_to_list( <<Service/binary, $., Region/binary, ".amazonaws.com">> ),
+    get_host_vpc_endpoint(Service, Default).
+
+-spec get_host_vpc_endpoint(binary(), binary()) -> binary().
+%% take the list of possibly configured endpoints
+%% and pick the one which suites from our AZ.
+get_host_vpc_endpoint(Service, Default) when is_binary(Service) ->
+    ConfiguredEndpoints = proplists:get_value(Service,
```
nalundgaard, line 816: Split this into 2 calls, IMO

```erlang
VPCEndpointsByService = application:get_env(erlcloud, services_vpc_endpoints, []),
ConfiguredEndpoints = proplists:get_value(Service, ConfiguredEndpoints, []),
```

As you have it, it looks like it can return `undefined` and reach a case clause?

src/erlcloud_aws.erl
```diff
@@ -796,14 +797,59 @@ service_host( <<"s3">>, Region ) ->
...
 service_host( Service, <<"cn-north-1">> = Region ) when is_binary(Service) ->
     binary_to_list( <<Service/binary, $., Region/binary, ".amazonaws.com.cn">> );
-service_host( Service, <<"cn-northwest-1">> =Region ) when is_binary(Service) ->
+service_host( Service, <<"cn-northwest-1">> = Region ) when is_binary(Service) ->
     binary_to_list( <<Service/binary, $., Region/binary, ".amazonaws.com.cn">> );
-service_host( Service, Region ) when is_binary(Service) ->
-    binary_to_list( <<Service/binary, $., Region/binary, ".amazonaws.com">> ).
+% services which can have VPCe configured to mitigate cross-AZ traffic.
+% It's application level decision to use VPCe and configure those
+% magic can be done vi EC2 DescribeVpcEndpoints/filter by VPC/filter by AZ,
```
nalundgaard, line 806: s/vi/via

src/erlcloud_aws.erl
```diff
@@ -796,14 +797,59 @@ service_host( <<"s3">>, Region ) ->
...
-service_host( <<"states">>, Region ) ->
-    binary_to_list( <<"states.", Region/binary, ".amazonaws.com">> );
 service_host( Service, <<"cn-north-1">> = Region ) when is_binary(Service) ->
     binary_to_list( <<Service/binary, $., Region/binary, ".amazonaws.com.cn">> );
-service_host( Service, <<"cn-northwest-1">> =Region ) when is_binary(Service) ->
+service_host( Service, <<"cn-northwest-1">> = Region ) when is_binary(Service) ->
     binary_to_list( <<Service/binary, $., Region/binary, ".amazonaws.com.cn">> );
-service_host( Service, Region ) when is_binary(Service) ->
-    binary_to_list( <<Service/binary, $., Region/binary, ".amazonaws.com">> ).
+% services which can have VPCe configured to mitigate cross-AZ traffic.
```
nalundgaard, line 804: I think I would just put this comment at the top of the function head, and also in a short README segment.

src/erlcloud_aws.erl
nalundgaard: Is this worth caching? I feel like it should never change without an application restart, is intrinsically application-wide, and therefore could be encapsulated into an ephemeral application env setting. Do we need to spam and parse instance metadata every time we want the service host?

README.md
```diff
@@ -157,8 +157,24 @@ Example of such app.config:
...
+{erlcloud, [
+    {services_vpc_endpoints, [
+        {<<"sqs">>, [<<"myAZ1.sqs-dns.amazonaws.com">>, <<"myAZ2.sqs-dns.amazonaws.com">>]},
+        {<<"kinesis">>, {env, "KINESIS_VPC_ENDPOINTS"}}
+    ]}
+]}
+```
+Two options supported:
+ - explicit lists of route53 AZ endpoints
+ - OS environment variable, comes handy for ECS deployments.
```
nalundgaard, line 172: We should note in the docs that the OS env variable is parsed as a CSV of VPC endpoints.

motobob:
do not merge still

motobob:
@nalundgaard please take a final look. confirmed this is now working on both EC2 and ECS environments

PR #642, opened 2020-05-26
Prepare for the future - OTP 23 (http_uri:parse/1)

paulo-ferraz-oliveira:
This PR fixes one of the classes of issues (the use of `http_uri:parse/1`) that I stumbled upon while `dialyzer`ing in OTP 23.

Some other "issues" persist and I'll get around to them if/when I have time.

I've added a very basic test, and `make eunit` is without error.

At any rate, `.travis.yml`, is not set for OTP 22 or 23.

### DIFF
--- a/src/erlcloud_s3.erl
+++ b/src/erlcloud_s3.erl
@@ -1738,7 +1738,7 @@ create_notification_param_xml({cloud_function, CF}, Acc) -> [{'CloudFunction', [
 -spec get_bucket_and_key(string()) -> {string(), string()}.
 
 get_bucket_and_key(Uri) ->
-  {ok, Parsed} = http_uri:parse(Uri),
+  {ok, Parsed} = erlcloud_util:uri_parse(Uri),
   {Host, Path} = extract_host_and_path(Parsed),
   extract_location_fields(Host, Path).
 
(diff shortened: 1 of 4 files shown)


paulo-ferraz-oliveira:
> Makes sense to me. Any reason not to just put 22.3/23.0 into travis?

You mean in the current PR? I didn't think about putting 23.0- since it was out-of-scope. On the other hand, if I include 23.0 now it should compile with warnings - at least. But I'm OK with that, if you are.

My preference would go toward solving all the other 23.0 issues, first, though, and then add the target. Oh, and on the subject of issues, `lhttpc` will "soon" become one, since there's also breaking/deprecation changes in the OTP API that affect it.

nalundgaard:
Good point on 23, I guess best to keep it out till it’s ready. I’ve had updating our lhttpc on my to do list for awhile, I’m sure 23 will splay into that later this year.

paulo-ferraz-oliveira:
This `lhttpc` fork, https://github.com/MiniclipPortugal/lhttpc, is maintained and compatible with `erlcloud` since it's what we use internally at Miniclip, but I wouldn't know about breaking differences from yours, though. If you want to, I can, in a new open issue, enumerate what it would mean to change to it.

paulo-ferraz-oliveira:
@nalundgaard: do you think this will get merged soon? Thanks.

nalundgaard:
Yes! Sorry I spaced on this, I’ll give it a final check in the morning and look at your other PRs. Thanks for the reminder!

PR #643, opened 2020-06-07
Add CloudWatch Log Groups and Streams creation and deletion functions

noxdafox:
Greetings,

I am building an Erlang Lager backend for CloudWatch and I found these functions to be of use for this case.


noxdafox:
Anyone who could review this PR?

PR #645, opened 2020-06-29
Make use of jsx compatibe with 3.x

nalundgaard:
[jsx](https://github.com/talentdeficit/jsx) moved to [v3.0.0](https://github.com/talentdeficit/jsx/releases/tag/v3.0.0) on Friday, 26 June 2020. See [jsx#139](https://github.com/talentdeficit/jsx/pull/139). The big change is that the default option is that `return_maps` now defaults to `true`, meaning that you will get maps output **unless** you explicitly pass `[{return_maps, false}]` when decoding JSON.

This change updates internal use of jsx to *always* pass `return_maps` explicitly, true or false, so that it is compatible with jsx [v3.0.0](https://github.com/talentdeficit/jsx/releases/tag/v3.0.0). I also updated the pin of jsx to [v2.11.0](https://github.com/talentdeficit/jsx/releases/tag/v2.11.0), since I don't want to force 3.x onto clients just yet.


paulo-ferraz-oliveira:
I found one use of `jsx` without `return_maps`, in `erlcloud_sm`.

nalundgaard:
@paulo-ferraz-oliveira looks like I missed it in the later https://github.com/erlcloud/erlcloud/pull/654. Will PR a fix. Thanks.

PR #647, opened 2020-06-30
Add dialyzer checks to Travis

paulo-ferraz-oliveira:
This Pull Request is an attempt at trying to prevent further dialyzer -related warnings/issues from creeping into the project.

I had to update the `Makefile` for rebar 2 to run with the current Travis matrix, only to find another issue: the use of `string:split` in OTP 19.3 (introduced in 90b7ee2, from Apr.04.20). This function was only [added in OTP 20.0](https://erlang.org/doc/man/string.html#split-2). This issue was fixed, since it derived from the original fix's context.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,23 +57,32 @@ else
 	@$(REBAR) eunit
 endif
 
+.dialyzer_plt:
+	dialyzer --build_plt -r deps \
+		--apps erts kernel stdlib inets crypto public_key ssl xmerl \
+		--fullpath \
+		--output_plt .dialyzer_plt
+
 check:
 ifeq ($(REBAR_VSN),2)
 	@$(REBAR) compile
-	dialyzer --verbose --no_check_plt --no_native --fullpath \
+	$(MAKE) .dialyzer_plt
+	dialyzer --no_check_plt --fullpath \
 		$(CHECK_FILES) \
-		-Wunmatched_returns \
-		-Werror_handling
+		-I include \
+		--plt .dialyzer_plt
 else
 	@$(REBAR) dialyzer
 endif
 
 check-eunit: eunit
 ifeq ($(REBAR_VSN),2)
-	dialyzer --verbose --no_check_plt --no_native --fullpath \
+	@$(REBAR) compile
+	$(MAKE) .dialyzer_plt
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -57,23 +57,32 @@ else
...
+		--output_plt .dialyzer_plt
+
 check:
 ifeq ($(REBAR_VSN),2)
 	@$(REBAR) compile
-	dialyzer --verbose --no_check_plt --no_native --fullpath \
+	$(MAKE) .dialyzer_plt
+	dialyzer --no_check_plt --fullpath \
 		$(CHECK_FILES) \
-		-Wunmatched_returns \
```
paulo-ferraz-oliveira, line 65: These two changes (replicated further below) were performed so that rebar 2 -based analysis is done with the same configuration as rebar3 -based analysis, thus yielding the same results.

Makefile
```diff
@@ -57,23 +57,32 @@ else
 	@$(REBAR) eunit
 endif
 
+.dialyzer_plt:
```
nalundgaard, line 60: should we add a rebar.config hook to clean this when cleaning?

paulo-ferraz-oliveira: Sure, if it makes sense for you. I just tweaked the `rebar 2` "options" to make what was there work properly, not because I'm using rebar 2 still. Do you want me to add that "clean" statement?

paulo-ferraz-oliveira: Actually, I'm going to go ahead and assume you want the change, so I'll perform it; otherwise you wouldn't have asked the question :)

paulo-ferraz-oliveira: Actually, now that I read your question properly, you talk about a _hook_. I'll have to look up how to do that in `rebar 2`, since I don't think I've ever done it previously...

paulo-ferraz-oliveira: I did it using a "post" hook. Though it works for `rebar3` also it'll have no affect since (to my knowledge) the PLT is usually generated inside `_build`.

nalundgaard: yeah, I think I got the 2 rebars confused, probably was fine to leave as-is, but it's certainly harmless. I can just merge as-is.

PR #649, opened 2020-07-01
Prepare for the future - OTP 25 (http_uri:decode/1 + http_uri:encode/1)

paulo-ferraz-oliveira:
This PR fixes the last class of issues (the use of `http_uri:decode/1` + `http_uri:encode/1`) that I stumbled upon while compiling on OTP 23.

If this PR is accepted after #647, #642 and #648 (it'll probably present conflicts in `erlcloud_util.erl`) it should be safe to add OTP 23.0 to `.travis.yml` for a more complete validation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ install:
 script:
   - make check_warnings
   - make check
+  - make check-eunit
   - make eunit
 deploy:
   skip_cleanup: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,6 @@ install:
 script:
   - make check_warnings
   - make check
-  - make check-eunit
   - make eunit
 deploy:
   skip_cleanup: true
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,7 @@ ifeq ($(REBAR_VSN),2)
 		-I include \
 		--plt .dialyzer_plt
 else
-	@$(REBAR) as eunit_check dialyzer
+	@$(REBAR) dialyzer
 endif
 
 doc:
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,7 @@ ifeq ($(REBAR_VSN),2)
 		-I include \
 		--plt .dialyzer_plt
 else
-	@$(REBAR) dialyzer
+	@$(REBAR) as eunit_check dialyzer
 endif
 
 doc:
--- a/rebar.config
+++ b/rebar.config
@@ -33,7 +33,6 @@
 
 {profiles, [
             {test, [{deps, [{meck, "0.8.13"}]}]}
-           ,{eunit_check, [{extra_src_dirs, ["test"]}]}
            ,{warnings, [{erl_opts, [warnings_as_errors]}]}
            ]}.
 
--- a/rebar.config
+++ b/rebar.config
@@ -33,6 +33,7 @@
 
 {profiles, [
             {test, [{deps, [{meck, "0.8.13"}]}]}
+           ,{eunit_check, [{extra_src_dirs, ["test"]}]}
            ,{warnings, [{erl_opts, [warnings_as_errors]}]}
            ]}.
 
--- a/src/erlcloud_iam.erl
+++ b/src/erlcloud_iam.erl
@@ -1055,7 +1055,7 @@ data_fun("Boolean") -> {erlcloud_xml, get_bool};
 data_fun("Uri") -> {?MODULE, get_uri}.
 
 get_uri(Key, Item) ->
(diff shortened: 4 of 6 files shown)


.travis.yml
```diff
@@ -19,6 +19,7 @@ install:
   - make travis-install
 script:
   - make check_warnings
```
paulo-ferraz-oliveira, line 21: A counterpart for this could be added for the `eunit` tests to compile with "warnings as errors", but I didn't want to go so far.

nalundgaard: Yeah, that's probably just vexing for people who want to use the application on future erlang versions that we haven't adjusted for just yet.

nalundgaard:
Looks like this needs a rebase from merging #642, #648

paulo-ferraz-oliveira:
> Looks like this needs a rebase from merging #642, #648

I've rebased it onto master.

.travis.yml
paulo-ferraz-oliveira: Hm... maybe I didn't express myself correctly. I was stating that, e.g., `rebar3` profile 'test' could be "enhanced" with option `{erl_opts, [warnings_as_errors]}` (which I did locally for testing) and that wouldn't break what exists "too much".

In any case, if you're interested, this is what happens when you do that change.

1. compilation will initially stop (first clean run) with
```erlang
test/erlcloud_sdb_tests.erl:4: export_all flag enabled - all functions will be exported
```

2. after that issue is solved compilation will proceed with
```erlang
test/erlcloud_aws_tests.erl:71: this clause cannot match because a previous clause at line 71 always matches
test/erlcloud_aws_tests.erl:72: this clause cannot match because a previous clause at line 72 always matches
test/erlcloud_aws_tests.erl:73: this clause cannot match because a previous clause at line 73 always matches
```

These are the only two issues I'm aware of, for that particular change.

paulo-ferraz-oliveira: I'll deal with this in a separate Pull Request.

nalundgaard: Makes sense, I didn't realize that you meant only to use warnings_as_errors for unit testing. I am +1 for that change, and I like #650.

nalundgaard:
looks like there are a few failures in travis:

https://travis-ci.org/github/erlcloud/erlcloud/jobs/704137559
```
429 dialyzer --verbose --no_check_plt --no_native --fullpath \
430	ebin/*.beam .eunit/*.beam \
431	-Wunmatched_returns \
432	-Werror_handling
433
434 dialyzer: Could not read PLT file /home/travis/.dialyzer_plt: no_such_file
435
436 Makefile:73: recipe for target 'check-eunit' failed
437 make: *** [check-eunit] Error 1
438 The command "make check-eunit" exited with 2.
```

https://travis-ci.org/github/erlcloud/erlcloud/jobs/704137566
```
278 ===> Analyzing 55 files with "/home/travis/build/erlcloud/erlcloud/_build/eunit_check/rebar3_22.3_plt"...
279
280 src/erlcloud_cloudwatch_logs.erl
281  157: Function create_log_group/1 has no local return
282  165: Function create_log_group/2 has no local return
283  166: The call erlcloud_cloudwatch_logs:create_log_group(LogGroupName::any(), 'undefined', 'undefined', Config::any()) breaks the contract (log_group_name(), [tag()], kms_key_id(), aws_config()) -> 'ok' | error_result()
284 ===> Warnings written to /home/travis/build/erlcloud/erlcloud/_build/eunit_check/22.3.dialyzer_warnings
285 ===> Warnings occurred running dialyzer: 3
286 Makefile:78: recipe for target 'check-eunit' failed
287 make: *** [check-eunit] Error 1
288 The command "make check-eunit" exited with 2.
```

Should we just disable dialyzer for rebar2, at least in travis? seems it might be an issue with the environment.

paulo-ferraz-oliveira:
> Looks like there are a few failures in travis

They'll go away if I rebase this on top of `master` (which now includes `feature/dialyzer_in_travis`) which already presented that issue. I'll do that later.

paulo-ferraz-oliveira:
> They'll go away if I rebase this on top of `master` (which now includes `feature/dialyzer_in_travis`) which already presented that issue. I'll do that later.

Since this became a code-only Pull Request (as of the latest rebase) there should be no reason for the tests to fail. I can't follow the link to travis-ci.org, though, since something is broken with the certificate chain.

I'll check back later.

paulo-ferraz-oliveira:
Nice, thanks. I forgot to add `23.0` to `.travis.yml`, though. Maybe you could do it directly in `master`?

PR #499, opened 2018-04-11
Add support for DDB Backups in the erlcloud_ddb2 module

nanshakov:
Implemented API:
1. ListBackups
2. CreateBackup
3. DeleteBackup
4. DescribeBackup
5. DescribeContinuousBackups
6. RestoreTableFromBackup
And tests.

PR #195, opened 2015-07-31
Add DynamoDB Streams support

smilerlee:
@nalundgaard and I have finished the implementation, more discuss can be found in #189.
I prefer to preserve the history instead of squashing all these commit together, since not all the commits are made by me.
Anyway, if we need to keep a clean code base history, I can make a squashed PR later.

PR #650, opened 2020-07-02
Be stricter on eunit compilation-based warnings

paulo-ferraz-oliveira:
We stop compilation with error the same we would for normal code (`warnings_as_errors`).

We apply this to the Travis checks.

We fix the issues surfaced by this increased strictness.

### DIFF
--- a/include/erlcloud_ddb2.hrl
+++ b/include/erlcloud_ddb2.hrl
@@ -279,16 +279,16 @@
 -record(ddb2_q, 
         {consumed_capacity :: undefined | #ddb2_consumed_capacity{},
          count :: undefined | non_neg_integer(),
-         items :: undefined | [erlcloud_ddb2:out_item()],
-         last_evaluated_key :: undefined | erlcloud_ddb2:key(),
+         items :: undefined | [erlcloud_ddb2:out_item()] | [binary()],
+         last_evaluated_key :: undefined | erlcloud_ddb2:key() | binary(),
          scanned_count :: undefined | non_neg_integer()
         }).
 
 -record(ddb2_scan, 
         {consumed_capacity :: undefined | #ddb2_consumed_capacity{},
          count :: undefined | non_neg_integer(),
-         items :: undefined | [erlcloud_ddb2:out_item()],
-         last_evaluated_key :: undefined | erlcloud_ddb2:key(),
+         items :: undefined | [erlcloud_ddb2:out_item()] | [binary()],
+         last_evaluated_key :: undefined | erlcloud_ddb2:key() | binary(),
          scanned_count :: undefined | non_neg_integer()
         }).
 
--- a/include/erlcloud_ddb2.hrl
+++ b/include/erlcloud_ddb2.hrl
@@ -361,7 +361,7 @@
          provisioned_throughput :: undefined | #ddb2_provisioned_throughput{},
          table_arn :: undefined | binary(),
          table_creation_date_time :: undefined | number(),
-         table_id :: undefined | number() | binary(),
+         table_id :: undefined | binary(),
          table_name :: undefined | binary(),
          table_size_bytes :: undefined | integer()
         }).
--- a/include/erlcloud_ddb2.hrl
+++ b/include/erlcloud_ddb2.hrl
@@ -279,7 +279,7 @@
 -record(ddb2_q, 
         {consumed_capacity :: undefined | #ddb2_consumed_capacity{},
          count :: undefined | non_neg_integer(),
-         items :: undefined | [erlcloud_ddb2:out_item()] | [binary()],
+         items :: undefined | [erlcloud_ddb2:out_item()],
          last_evaluated_key :: undefined | erlcloud_ddb2:key() | binary(),
          scanned_count :: undefined | non_neg_integer()
         }).
@@ -287,7 +287,7 @@
 -record(ddb2_scan, 
         {consumed_capacity :: undefined | #ddb2_consumed_capacity{},
          count :: undefined | non_neg_integer(),
-         items :: undefined | [erlcloud_ddb2:out_item()] | [binary()],
+         items :: undefined | [erlcloud_ddb2:out_item()],
          last_evaluated_key :: undefined | erlcloud_ddb2:key() | binary(),
          scanned_count :: undefined | non_neg_integer()
         }).
--- a/include/erlcloud_ddb2.hrl
+++ b/include/erlcloud_ddb2.hrl
@@ -113,8 +113,8 @@
         {last_decrease_date_time :: undefined | date_time(),
          last_increase_date_time :: undefined | date_time(),
          number_of_decreases_today :: undefined | integer(),
-         read_capacity_units :: undefined | non_neg_integer(),
-         write_capacity_units :: undefined | non_neg_integer()
+         read_capacity_units :: undefined | pos_integer(),
+         write_capacity_units :: undefined | pos_integer()
         }).
 
 -record(ddb2_global_secondary_index_description,
--- a/include/erlcloud_ddb2.hrl
+++ b/include/erlcloud_ddb2.hrl
@@ -280,7 +280,7 @@
         {consumed_capacity :: undefined | #ddb2_consumed_capacity{},
          count :: undefined | non_neg_integer(),
          items :: undefined | [erlcloud_ddb2:out_item()],
-         last_evaluated_key :: undefined | erlcloud_ddb2:key() | binary(),
+         last_evaluated_key :: undefined | erlcloud_ddb2:key(),
          scanned_count :: undefined | non_neg_integer()
         }).
 
@@ -288,7 +288,7 @@
         {consumed_capacity :: undefined | #ddb2_consumed_capacity{},
          count :: undefined | non_neg_integer(),
          items :: undefined | [erlcloud_ddb2:out_item()],
-         last_evaluated_key :: undefined | erlcloud_ddb2:key() | binary(),
+         last_evaluated_key :: undefined | erlcloud_ddb2:key(),
          scanned_count :: undefined | non_neg_integer()
         }).
 
--- a/include/erlcloud_ddb2.hrl
+++ b/include/erlcloud_ddb2.hrl
@@ -47,8 +47,14 @@
         {index_name :: undefined | binary(),
          provisioned_throughput_override :: undefined | #ddb2_provisioned_throughput_override{}}).
 
+-record(ddb2_replica_global_secondary_index_auto_scaling_description,
+        {index_name :: undefined | binary(),
+         index_status :: undefined | index_status(),
+         provisioned_read_capacity_auto_scaling_settings :: undefined | #ddb2_auto_scaling_settings_description{},
+         provisioned_write_capacity_auto_scaling_settings :: undefined | #ddb2_auto_scaling_settings_description{}}).
+
 -record(ddb2_replica_auto_scaling_description,
-        {global_secondary_indexes :: undefined | [#ddb2_replica_global_secondary_index_description{}],
+        {global_secondary_indexes :: undefined | [#ddb2_replica_global_secondary_index_auto_scaling_description{}],
          region_name :: undefined | binary(),
          replica_provisioned_read_capacity_auto_scaling_settings :: undefined | #ddb2_auto_scaling_settings_description{},
          replica_provisioned_write_capacity_auto_scaling_settings :: undefined | #ddb2_auto_scaling_settings_description{},
@@ -72,15 +78,13 @@
          provisioned_write_capacity_auto_scaling_settings :: undefined | #ddb2_auto_scaling_settings_description{},
          provisioned_write_capacity_units :: undefined | pos_integer()}).
 
--record(ddb2_replica_global_secondary_index_auto_scaling_description,
-        {index_name :: undefined | binary(),
-         index_status :: undefined | binary(),
-         provisioned_read_capacity_auto_scaling_settings :: undefined | #ddb2_auto_scaling_settings_description{},
-         provisioned_write_capacity_auto_scaling_settings :: undefined | #ddb2_auto_scaling_settings_description{}}).
+-record(ddb2_billing_mode_summary,
+        {billing_mode :: undefined | erlcloud_ddb2:billing_mode(),
+        last_update_to_pay_per_request_date_time :: undefined | date_time()}).
 
 -record(ddb2_replica_settings_description,
         {region_name :: undefined | binary(),
-         replica_billing_mode_summary :: undefined | binary(),
+         replica_billing_mode_summary :: undefined | #ddb2_billing_mode_summary{},
          replica_global_secondary_index_settings :: undefined | [#ddb2_replica_global_secondary_index_settings_description{}],
          replica_provisioned_read_capacity_auto_scaling_settings :: undefined | #ddb2_auto_scaling_settings_description{},
          replica_provisioned_read_capacity_units :: undefined | number(),
@@ -105,16 +109,12 @@
          replication_group :: undefined | [#ddb2_replica{}]
         }).
 
--record(ddb2_billing_mode_summary,
-       {billing_mode :: undefined | erlcloud_ddb2:billing_mode(),
-        last_update_to_pay_per_request_date_time :: undefined | date_time()}).
-
 -record(ddb2_provisioned_throughput_description,
         {last_decrease_date_time :: undefined | date_time(),
          last_increase_date_time :: undefined | date_time(),
          number_of_decreases_today :: undefined | integer(),
-         read_capacity_units :: undefined | pos_integer(),
-         write_capacity_units :: undefined | pos_integer()
+         read_capacity_units :: undefined | non_neg_integer(),
+         write_capacity_units :: undefined | non_neg_integer()
         }).
 
 -record(ddb2_global_secondary_index_description,
@@ -361,7 +361,7 @@
          provisioned_throughput :: undefined | #ddb2_provisioned_throughput{},
          table_arn :: undefined | binary(),
          table_creation_date_time :: undefined | number(),
-         table_id :: undefined | number(),
+         table_id :: undefined | number() | binary(),
          table_name :: undefined | binary(),
          table_size_bytes :: undefined | integer()
         }).
@@ -388,7 +388,7 @@
         {global_secondary_indexes :: undefined | [#ddb2_global_secondary_index_info{}],
          local_secondary_indexes  :: undefined | [#ddb2_local_secondary_index_info{}],
          sse_description :: undefined | erlcloud_ddb2:sse_description(),
-         stream_description :: undefined | #ddb2_stream_description{},
+         stream_description :: undefined | #ddb2_stream_description{} | {boolean(), erlcloud_ddb2:stream_view_type()},
          time_to_live_description :: undefined | #ddb2_time_to_live_description{}
         }).
 
--- a/include/erlcloud_ecs.hrl
+++ b/include/erlcloud_ecs.hrl
@@ -38,14 +38,14 @@
 }).
 
 -record(ecs_deployment, {
-    created_at :: undefined | pos_integer(),
+    created_at :: undefined | pos_integer() | float(),
     desired_count :: undefined | pos_integer(),
     id :: undefined | binary(),
-    pending_count :: undefined | pos_integer(),
-    running_count :: undefined | pos_integer(),
+    pending_count :: undefined | non_neg_integer(),
+    running_count :: undefined | non_neg_integer(),
     status :: undefined | binary(),
     task_definition:: undefined | binary(),
-    updated_at :: undefined | pos_integer()
+    updated_at :: undefined | float()
 }).
 
 -record(ecs_event, {
@@ -62,12 +62,12 @@
 }).
 
 -record(ecs_cluster, {
-    active_services_count :: undefined | pos_integer(),
+    active_services_count :: undefined | non_neg_integer(),
     cluster_arn :: undefined | binary(),
     cluster_name :: undefined | binary(),
-    pending_tasks_count :: undefined | pos_integer(),
-    registered_container_instances_count :: undefined | pos_integer(),
-    running_tasks_count :: undefined | pos_integer(),
+    pending_tasks_count :: undefined | non_neg_integer(),
+    registered_container_instances_count :: undefined | non_neg_integer(),
+    running_tasks_count :: undefined | non_neg_integer(),
     status :: undefined | binary()
 }).
 
@@ -79,9 +79,9 @@
     desired_count :: undefined | pos_integer(),
     events :: undefined | [#ecs_event{}],
     load_balancers :: undefined | [#ecs_load_balancer{}],
-    pending_count :: undefined | pos_integer(),
+    pending_count :: undefined | non_neg_integer(),
     role_arn :: undefined | binary(),
-    running_count :: undefined | pos_integer(),
+    running_count :: undefined | non_neg_integer(),
     service_arn :: undefined | binary(),
     service_name :: undefined | binary(),
     status :: undefined | binary(),
@@ -89,11 +89,11 @@
 }).
 
 -record(ecs_resource, {
-   double_value :: undefined | pos_integer(),
-   integer_value :: undefined | pos_integer(),
-   long_value :: undefined | pos_integer(),
+   double_value :: undefined | non_neg_integer(),
+   integer_value :: undefined | non_neg_integer(),
+   long_value :: undefined | non_neg_integer(),
    name :: undefined | binary(),
-   string_set_value :: undefined | binary(),
+   string_set_value :: undefined | [binary()],
    type :: undefined | binary()
 }).
 
@@ -109,10 +109,10 @@
     attributes :: undefined | [#ecs_attribute{}],
     container_instance_arn :: undefined | binary(),
     ec2_instance_id :: undefined | binary(),
-    pending_tasks_count :: undefined | pos_integer(),
+    pending_tasks_count :: undefined | non_neg_integer(),
     registered_resources :: undefined | [#ecs_resource{}],
     remaining_resources :: undefined | [#ecs_resource{}],
-    running_tasks_count :: undefined | pos_integer(),
+    running_tasks_count :: undefined | non_neg_integer(),
     status :: undefined | binary(),
     version_info :: undefined | #ecs_version_info{}
 }).
@@ -210,7 +210,7 @@
 -record(ecs_task_definition, {
     container_definitions :: undefined | [#ecs_container_definition{}],
     family :: undefined | binary(),
-    network_mode :: undefined | binary(),
+    network_mode :: undefined | ecs_network_mode(),
     requires_attributes :: undefined | [#ecs_attribute{}],
     revision :: undefined | pos_integer(),
     status :: undefined | binary(),
--- a/include/erlcloud_ecs.hrl
+++ b/include/erlcloud_ecs.hrl
@@ -38,14 +38,14 @@
 }).
 
 -record(ecs_deployment, {
-    created_at :: undefined | pos_integer() | float(),
+    created_at :: undefined | pos_integer(),
     desired_count :: undefined | pos_integer(),
     id :: undefined | binary(),
     pending_count :: undefined | non_neg_integer(),
     running_count :: undefined | non_neg_integer(),
     status :: undefined | binary(),
     task_definition:: undefined | binary(),
-    updated_at :: undefined | float()
+    updated_at :: undefined | pos_integer()
 }).
 
 -record(ecs_event, {
--- a/include/erlcloud_ecs.hrl
+++ b/include/erlcloud_ecs.hrl
@@ -38,18 +38,18 @@
 }).
 
 -record(ecs_deployment, {
-    created_at :: undefined | pos_integer(),
+    created_at :: undefined | number(),
     desired_count :: undefined | pos_integer(),
     id :: undefined | binary(),
     pending_count :: undefined | non_neg_integer(),
     running_count :: undefined | non_neg_integer(),
     status :: undefined | binary(),
     task_definition:: undefined | binary(),
-    updated_at :: undefined | pos_integer()
+    updated_at :: undefined | number()
 }).
 
 -record(ecs_event, {
-    created_at :: undefined | pos_integer(),
+    created_at :: undefined | number(),
     id :: undefined | binary(),
     message :: undefined | binary()
 }).
@@ -73,7 +73,7 @@
 
 -record(ecs_service, {
     cluster_arn :: undefined | binary(),
-    created_at :: undefined | pos_integer(),
+    created_at :: undefined | number(),
     deployment_configuration :: undefined | #ecs_deployment_configuration{},
     deployments :: undefined | [#ecs_deployment{}],
     desired_count :: undefined | pos_integer(),
@@ -251,7 +251,7 @@
     cluster_arn :: undefined | binary(),
     container_instance_arn :: undefined | binary(),
     containers :: undefined | [#ecs_container{}],
-    created_at :: undefined | pos_integer(),
+    created_at :: undefined | number(),
     desired_status:: undefined | binary(),
     last_status :: undefined | binary(),
     overrides :: undefined | #ecs_task_override{},
--- a/rebar.config
+++ b/rebar.config
@@ -32,7 +32,7 @@
 
 
 {profiles, [
-            {test, [{deps, [{meck, "0.8.13"}]}]}
+            {test, [{deps, [{meck, "0.8.13"}]}, {erl_opts, [warnings_as_errors]}]}
            ,{warnings, [{erl_opts, [warnings_as_errors]}]}
            ]}.
 
--- a/rebar.config
+++ b/rebar.config
@@ -32,7 +32,7 @@
 
 
 {profiles, [
-            {test, [{deps, [{meck, "0.8.13"}]}, {erl_opts, [warnings_as_errors]}]}
+            {test, [{deps, [{meck, "0.9.0"}]}, {erl_opts, [warnings_as_errors]}]}
            ,{warnings, [{erl_opts, [warnings_as_errors]}]}
            ]}.
 
--- a/src/erlcloud_cloudwatch_logs.erl
+++ b/src/erlcloud_cloudwatch_logs.erl
@@ -465,7 +465,7 @@ log_stream_order_by(last_event_time) -> <<"LastEventTime">>.
     log_stream_name(),
     seq_token(),
     events()
-) -> datum:either( seq_token() ).
+) -> {ok, seq_token()} | {error, erlcloud_aws:httpc_result_error()}.
 
 put_logs_events(LogGroup, LogStream, SeqToken, Events) ->
     put_logs_events(LogGroup, LogStream, SeqToken, Events, default_config()).
@@ -477,7 +477,7 @@ put_logs_events(LogGroup, LogStream, SeqToken, Events) ->
     seq_token(),
     events(),
     aws_config()
-) -> datum:either( seq_token() ).
+) -> {ok, seq_token()} | {error, erlcloud_aws:httpc_result_error()}.
 
 put_logs_events(LogGroup, LogStream, SeqToken, Events, Config) ->
     case
--- a/src/erlcloud_ddb2.erl
+++ b/src/erlcloud_ddb2.erl
@@ -1874,9 +1874,9 @@ dynamize_sse_specification({enabled, Enabled}) when is_boolean(Enabled) ->
 -type create_table_opt() :: {billing_mode, billing_mode()} |
                             {local_secondary_indexes, local_secondary_indexes()} |
                             {global_secondary_indexes, global_secondary_indexes()} |
+                            {provisioned_throughput, {read_units(), write_units()}} |
                             {sse_specification, sse_specification()} |
-                            {stream_specification, stream_specification()} |
-                            {provisioned_throughput, {read_units(), write_units()}}.
+                            {stream_specification, stream_specification()}.
 -type create_table_opts() :: [create_table_opt()].
 
 -spec create_table_opts(key_schema()) -> opt_table().
--- a/src/erlcloud_ddb2.erl
+++ b/src/erlcloud_ddb2.erl
@@ -1875,7 +1875,8 @@ dynamize_sse_specification({enabled, Enabled}) when is_boolean(Enabled) ->
                             {local_secondary_indexes, local_secondary_indexes()} |
                             {global_secondary_indexes, global_secondary_indexes()} |
                             {sse_specification, sse_specification()} |
-                            {stream_specification, stream_specification()}.
+                            {stream_specification, stream_specification()} |
+                            {provisioned_throughput, {read_units(), write_units()}}.
 -type create_table_opts() :: [create_table_opt()].
 
 -spec create_table_opts(key_schema()) -> opt_table().
@@ -3443,7 +3444,9 @@ tag_resource(ResourceArn, Tags, Config) ->
 %%%-----------------------------------------------------------------------------
 
 -type transact_get_items_transact_item_opts() :: expression_attribute_names_opt() |
-                           projection_expression_opt().
+                           projection_expression_opt() |
+                           return_consumed_capacity_opt() |
+                           {out, record}.
 -type transact_get_items_opts() :: [transact_get_items_transact_item_opts()].
 
 -type transact_get_items_get_item() :: {table_name(), key()}
@@ -3456,7 +3459,7 @@ tag_resource(ResourceArn, Tags, Config) ->
 
 -type transact_get_items_return() :: ddb_return(#ddb2_transact_get_items{}, out_item()).
 
--spec dynamize_transact_get_items_transact_items(transact_write_items_transact_items())
+-spec dynamize_transact_get_items_transact_items(transact_get_items_transact_items())
                                           -> [jsx:json_term()].
 dynamize_transact_get_items_transact_items(TransactItems) ->
     dynamize_maybe_list(fun dynamize_transact_get_items_transact_item/1, TransactItems).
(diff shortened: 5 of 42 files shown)


rebar.config
```diff
@@ -32,6 +32,6 @@
 
 
 {profiles, [
-            {test, [{deps, [{meck, "0.8.13"}]}]}
+            {test, [{deps, [{meck, "0.8.13"}]}, {erl_opts, [warnings_as_errors]}]}
```
nalundgaard, line 35: What do you think about adding `export_all, nowarn_export_all` here? that would seem convenient and still trip up on any other warnings as we want.

I am good either way, just wondering what you think on this.

paulo-ferraz-oliveira: I prefer to not allow `export_all` in my projects, but it's your call. That's the default, any way, in which sense we wouldn't have to update anything (if that's what you decide, I mean).

paulo-ferraz-oliveira:
I've rebased this onto `master`. There's still the question about allowing/not allowing `all_exports`.

paulo-ferraz-oliveira:
I ended up adding two extra commits (ported from #649) so that the Pull Request's content is more consistent and to try and avoid future merge conflicts.

nalundgaard:
Looks like it's clean on rebar3 now, only rebar2 giving new trouble:

Maybe `$CHECK_EUNIT_FILES` shouldn't contain `$CHECK_FILES`?

```
509make[1]: Leaving directory '/home/travis/build/erlcloud/erlcloud'
510dialyzer --no_check_plt --fullpath \
511	ebin/*.beam .eunit/*.beam \
512	-I include \
513	--plt .dialyzer_plt
514  Compiling some key modules to native code... done in 0m0.17s
515  Proceeding with analysis...
516dialyzer: Analysis failed with error:
517Duplicate modules: [["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_sns.beam",
518                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_sns.beam"],
519                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_as.beam",
520                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_as.beam"],
521                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_redshift.beam",
522                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_redshift.beam"],
523                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_lambda.beam",
524                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_lambda.beam"],
525                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_athena.beam",
526                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_athena.beam"],
527                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_kinesis.beam",
528                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_kinesis.beam"],
529                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_config.beam",
530                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_config.beam"],
531                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_application_autoscaler.beam",
532                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_application_autoscaler.beam"],
533                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_ddb_impl.beam",
534                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_ddb_impl.beam"],
535                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_http.beam",
536                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_http.beam"],
537                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_ec2_meta.beam",
538                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_ec2_meta.beam"],
539                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_cloudfront.beam",
540                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_cloudfront.beam"],
541                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_mms.beam",
542                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_mms.beam"],
543                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_glue.beam",
544                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_glue.beam"],
545                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_directconnect.beam",
546                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_directconnect.beam"],
547                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_mon.beam",
548                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_mon.beam"],
549                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_ecs_container_credentials.beam",
550                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_ecs_container_credentials.beam"],
551                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_ec2.beam",
552                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_ec2.beam"],
553                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_cur.beam",
554                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_cur.beam"],
555                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_autoscaling.beam",
556                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_autoscaling.beam"],
557                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_sqs.beam",
558                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_sqs.beam"],
559                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_kinesis_impl.beam",
560                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_kinesis_impl.beam"],
561                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_iam.beam",
562                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_iam.beam"],
563                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_ddb_util.beam",
564                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_ddb_util.beam"],
565                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_emr.beam",
566                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_emr.beam"],
567                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_xml.beam",
568                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_xml.beam"],
569                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_sdb.beam",
570                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_sdb.beam"],
571                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_rds.beam",
572                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_rds.beam"],
573                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud.beam",
574                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud.beam"],
575                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_waf.beam",
576                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_waf.beam"],
577                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_util.beam",
578                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_util.beam"],
579                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_s3.beam",
580                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_s3.beam"],
581                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_cloudformation.beam",
582                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_cloudformation.beam"],
583                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_ses.beam",
584                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_ses.beam"],
585                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_mturk.beam",
586                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_mturk.beam"],
587                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_kms.beam",
588                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_kms.beam"],
589                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_httpc.beam",
590                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_httpc.beam"],
591                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_elb.beam",
592                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_elb.beam"],
593                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_ecs.beam",
594                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_ecs.beam"],
595                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_ddb1.beam",
596                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_ddb1.beam"],
597                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_cloudtrail.beam",
598                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_cloudtrail.beam"],
599                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_aws.beam",
600                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_aws.beam"],
601                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_ddb_streams.beam",
602                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_ddb_streams.beam"],
603                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_mes.beam",
604                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_mes.beam"],
605                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_cloudsearch.beam",
606                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_cloudsearch.beam"],
607                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_sts.beam",
608                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_sts.beam"],
609                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_json.beam",
610                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_json.beam"],
611                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_ddb.beam",
612                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_ddb.beam"],
613                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_cloudwatch_logs.beam",
614                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_cloudwatch_logs.beam"],
615                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_route53.beam",
616                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_route53.beam"],
617                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_inspector.beam",
618                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_inspector.beam"],
619                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_states.beam",
620                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_states.beam"],
621                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_ddb2.beam",
622                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_ddb2.beam"],
623                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_guardduty.beam",
624                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_guardduty.beam"],
625                    ["/home/travis/build/erlcloud/erlcloud/.eunit/erlcloud_retry.beam",
626                     "/home/travis/build/erlcloud/erlcloud/ebin/erlcloud_retry.beam"]]
627Last messages in the log cache:
628  Reading files and computing callgraph... 
629Makefile:87: recipe for target 'check-eunit' failed
630make: *** [check-eunit] Error 1
631The command "make check-eunit" exited with 2.
```

paulo-ferraz-oliveira:
> Looks like it's clean on rebar3 now, only rebar2 giving new trouble:

Let me check...

paulo-ferraz-oliveira:
> Maybe `$CHECK_EUNIT_FILES` shouldn't contain `$CHECK_FILES`?

This makes sense, since we would effectively duplicate the files for analysis. I'm testing locally and pushing a commit at the same time (again, I rebased it onto `master`).

nalundgaard:
Here's the most recent output on rebar2:

[log.txt](https://github.com/erlcloud/erlcloud/files/4867062/log.txt)

Quite the wall of issues, all on the test beams AFAICT. I am thinking rebar3 must just not dialyze those modules, as I can't imagine it working on rebar3 and not on rebar2 any other way.

paulo-ferraz-oliveira:
> Quite the wall of issues, all on the test beams AFAICT. I am thinking rebar3 must just not dialyze those modules, as I can't imagine it working on rebar3 and not on rebar2 any other way.

What's happening is that `rebar.config` was not correct (which is why `rebar3` wasn't picking up those issues). It'll take a while longer, but I think I can tackle it... it's a nice exercise, in any case :D

paulo-ferraz-oliveira:
> What's happening is that `rebar.config` was not correct (which is why `rebar3` wasn't picking up those issues). It'll take a while longer, but I think I can tackle it... it's a nice exercise, in any case :D

I further looked into this and it seems the newer `rebar3` version (not yet released as `3.14.x`) already supports `rebar3 as test dialyzer` (https://github.com/erlang/rebar3/pull/2190). Since the way things are setup (in this branch) already shows the errors on `rebar 2` I'll go ahead and:
1. continue with the fixes (locally, using `rebar3` from `master`),
2. make sure `rebar.config` is updated as it should for when `rebar3` comes out.

What this means is that, for a while, dialyzer for `rebar 2` (`src` + `test`) and `rebar3` (`src` only) will run for different files. Is that an issue for you?

**Note**: I already found (via this analysis) `-spec(_).` errors on the main interfaces.

nalundgaard:
> What this means is that, for a while, dialyzer for `rebar 2` (`src` + `test`) and `rebar3` (`src` only) will run for different files. Is that an issue for you?

Not at all—not like it's something we can control, and it's good that it's being patched in rebar3 soon. Thanks for doing this!

paulo-ferraz-oliveira:
Some notes regarding the recent changes:
1. if a test **can be** changed in order to respect the public API - and still pass -, I _prefer to do that_ rather than changing the public API,
2. if a test **targets the public API** explicitly (e.g. wrong parameters), I prefer to signal that with `-dialyzer` than to tweak the public API (whenever possible I still try to fix as per `dialyzer` warnings),
3. `dialyzer`'ing the public API will only show warnings when code is called (e.g. not all public API elements will issue warnings if not tested),
4. some elements that I fixed might not have been so according to the API's expectations (I tried not to break the `-spec(_).` unless it was very evident it was wrong) - the API expectations (as far as I know) aren't declared anywhere but in their usage,
5. I'm unsure option `{out, record}` is part of an old API, but it's called by the tests quite a few times, so I preferred to "fix" the public API,
6. all the `undefined` in the `.hrl` files are only possible to observe when the API is called, which it is for many tests,
7. I put more emphasis on fixing tests than the public API which I assumed is the better working part of the tool,
8. the way some tests are organized (with a centralized output function **EXPECTED**/**RESULT**, not making use of the `eunit` macros) makes it hard to find the origin of their misbehavior,
9. after I made sure all tests were passing, I revisited `make checks-eunit` in order to be clear of warnings (which lead to some API interface spec. changes that weren't performed before, but were not compulsory, as proved by those tests).

paulo-ferraz-oliveira:
I think the checks might be somewhat redundant. Shouldn't we just run `compile` (with "warnings as errors"), `dialyze` and `eunit`?
(sure, they'll run in Travis CI, but maybe there's no need for such repetitiveness)

src/erlcloud_ddb_util.erl
```diff
@@ -51,7 +51,7 @@
 -type range_key_name() :: erlcloud_ddb2:range_key_name().
 -type table_name() :: erlcloud_ddb2:table_name().
 
--type items_return() :: {ok, [out_item()]}
+-type items_return() :: {ok, [out_item() | binary()]}
```
nalundgaard, line 54: Curious—where do these binary values come from in the client? I didn't realize DDB supported returning a list of strings.

src/erlcloud_inspector.erl
```diff
@@ -134,16 +134,16 @@ inspector_result_fun(#aws_request{response_type = error, error_type = aws} = Req
 -type attribute() :: proplists:proplist().
 
 -spec add_attributes_to_findings
-          (Attributes :: [attribute()],
-           FindingArns :: [string()]) ->
+          (Attributes :: attribute(),
```
nalundgaard, line 137: Is this type just mis-named? I feel like it should be `attributes()` in this case.

include/erlcloud_ddb2.hrl
```diff
@@ -105,16 +109,12 @@
...
-       {billing_mode :: undefined | erlcloud_ddb2:billing_mode(),
-        last_update_to_pay_per_request_date_time :: undefined | date_time()}).
-
 -record(ddb2_provisioned_throughput_description,
         {last_decrease_date_time :: undefined | date_time(),
          last_increase_date_time :: undefined | date_time(),
          number_of_decreases_today :: undefined | integer(),
-         read_capacity_units :: undefined | pos_integer(),
-         write_capacity_units :: undefined | pos_integer()
+         read_capacity_units :: undefined | non_neg_integer(),
```
paulo-ferraz-oliveira, line 116: I don't know if this is preferable to tweaking the test file. AWS console, for example, doesn't allow for IOPS of 0.

src/erlcloud_ddb_util.erl
paulo-ferraz-oliveira: In this specific case, they come from tests `test/erlcloud_ddb_util_tests.erl:665` and `test/erlcloud_ddb_util_tests.erl:812`. In goes a list of `binary()` out comes another one. Do you think the test is wrong (if I change the type `dialyzer` starts complaining again).

src/erlcloud_cloudwatch_logs.erl
```diff
@@ -465,7 +465,7 @@ log_stream_order_by(last_event_time) -> <<"LastEventTime">>.
     log_stream_name(),
     seq_token(),
     events()
-) -> datum:either( seq_token() ).
```
paulo-ferraz-oliveira, line 468: I don't know where this comes from, but I "fixed" it to what I could analyze by looking at the code and tests.

src/erlcloud_cloudwatch_logs.erl
```diff
@@ -477,7 +477,7 @@ put_logs_events(LogGroup, LogStream, SeqToken, Events) ->
     seq_token(),
     events(),
     aws_config()
-) -> datum:either( seq_token() ).
```
paulo-ferraz-oliveira, line 480: The same as before.

src/erlcloud_ecs.erl
```diff
@@ -1177,7 +1178,8 @@ decode_container_overrides_list(V, Opts) ->
 %%%------------------------------------------------------------------------------
 %% CreateCluster
 %%%------------------------------------------------------------------------------
--type create_cluster_opt() :: {cluster_name, string_param()}.
+-type create_cluster_opt() :: {cluster_name, string_param()} |
+                              {out, record}.
```
paulo-ferraz-oliveira, line 1182: I don't know if this is a proper option, or not, but it's referenced in the tests in a lot of different places.

src/erlcloud_inspector.erl
paulo-ferraz-oliveira: This refers to a `proplist`. I didn't want to make considerations on naming, since it's out of the scope of the PR, but can change it if you prefer `attributes()` (mind you that, if exported, it'll break "interface").

src/erlcloud_inspector.erl
```diff
@@ -134,16 +134,16 @@ inspector_result_fun(#aws_request{response_type = error, error_type = aws} = Req
...
+           FindingArns :: [string() | binary()]) ->
```
paulo-ferraz-oliveira, line 138: All of these `| binary()` respect the tests (I didn't want to break interface) but don't know what is preferable to keep or let go.

test/erlcloud_aws_tests.erl
```diff
@@ -29,7 +29,7 @@ config() ->
                 retry_num = 3}.
 
 request_default_test(_) ->
-    ok = erlcloud_aws:aws_request(get, "host", "/", [], "id", "key"),
+    _Body = erlcloud_aws:aws_request(get, "host", "/", [], "id", "key"),
```
paulo-ferraz-oliveira, line 32: I don't understand how these could have been passing, since I don't see `:aws_request` ever returning `ok`.

paulo-ferraz-oliveira:
In my Travis CI experiments it seems to be failing `rebar 2` for lack of `meck`. `rebar3` is using a profile. How should we make this work for `rebar 2`?

nalundgaard:
> In my Travis CI experiments it seems to be failing `rebar 2` for lack of `meck`. `rebar3` is using a profile. How should we make this work for `rebar 2`?

It should be present: https://github.com/erlcloud/erlcloud/blob/e4d903c/rebar.config.script#L8

Here's the actual output I see:

[log.txt](https://github.com/erlcloud/erlcloud/files/4887460/log.txt)

src/erlcloud_inspector.erl
nalundgaard: Nah, just leave it, as you say—the interface is what it is.

src/erlcloud_ddb_util.erl
nalundgaard: Those tests both look like lazy BS garbage-in/garbage out for test cases that test functions that treat those values opaquely in the layer tested. Could you just change this to

```erlang
    Item1 = [{<<"key">>, <<"item_1">>}],
    Item2 = [{<<"key">>, <<"item_2">>}],
```

in both cases? Thanks.

include/erlcloud_ddb2.hrl
nalundgaard: Better to change tests if there is a case here—you really can't set this to 0.

src/erlcloud_cloudwatch_logs.erl
nalundgaard: +1 to what you have, I have no idea what this means.

src/erlcloud_inspector.erl
nalundgaard: What you have looks correct, all the stuff calls `to_binary/1` which allows strings or binary strings. Just incomplete spec. Thanks.

test/erlcloud_aws_tests.erl
nalundgaard: that is... very weird.

nalundgaard: I wonder if these are working right at all? ending your test function with _test triggers special treatment with eunit, and I kind of wonder if that is short-circuiting something in eunit. maybe it would be better to make these functions in the `request_test_` group _prefixed_ with `test_` rather than _suffixed_ with `_test`.

src/erlcloud_ecs.erl
nalundgaard: looks like there is a type, `out_opt()`, that should be in all these.

src/erlcloud_ddb_util.erl
paulo-ferraz-oliveira: > Could you just change this

Sure, here it is: a76e395. Tests are passing.

Also, since now (?) `dialyzer` is complaining about the non-opaqueness of `meck:expect`'s 4th argument, I decided to bump it to 0.9.0 (latest stable release), that solves that issue with https://github.com/eproxus/meck/issues/212. Here it is: 6933ee7.

This surfaced a new issue that I solved in afe0067.

Tests are passing.

include/erlcloud_ddb2.hrl
paulo-ferraz-oliveira: Fixed in b66b410.

src/erlcloud_ecs.erl
paulo-ferraz-oliveira: Fixed in 18a08be. I think I might have fixed previous occurrences also, but since I was touching it... (I guess no harm done)

rebar.config
```diff
@@ -32,7 +32,7 @@
 
 
 {profiles, [
-            {test, [{deps, [{meck, "0.8.13"}]}]}
+            {test, [{deps, [{meck, "0.9.0"}]}, {erl_opts, [warnings_as_errors]}]}
```
nalundgaard, line 35: Note—I think you also need to fix this in the [`rebar.config.script`](https://github.com/erlcloud/erlcloud/blob/afe0067c864c72bf59b3309a61ba72730398a23b/rebar.config.script#L8) if you want rebar2 to use it.

include/erlcloud_ddb2.hrl
```diff
@@ -280,15 +280,15 @@
         {consumed_capacity :: undefined | #ddb2_consumed_capacity{},
          count :: undefined | non_neg_integer(),
          items :: undefined | [erlcloud_ddb2:out_item()],
-         last_evaluated_key :: undefined | erlcloud_ddb2:key(),
+         last_evaluated_key :: undefined | erlcloud_ddb2:key() | binary(),
```
nalundgaard, line 283: I think this must be some more lazy eunit mocking. This should not be a binary.

include/erlcloud_ddb2.hrl
```diff
@@ -280,15 +280,15 @@
...
          scanned_count :: undefined | non_neg_integer()
         }).
 
 -record(ddb2_scan, 
         {consumed_capacity :: undefined | #ddb2_consumed_capacity{},
          count :: undefined | non_neg_integer(),
          items :: undefined | [erlcloud_ddb2:out_item()],
-         last_evaluated_key :: undefined | erlcloud_ddb2:key(),
+         last_evaluated_key :: undefined | erlcloud_ddb2:key() | binary(),
```
nalundgaard, line 291: I think this must be some more lazy eunit mocking. This should not be a binary.

include/erlcloud_ddb2.hrl
nalundgaard: Are you sure? I still see it as `non_neg_integer()` in the diff here.

src/erlcloud_ddb2.erl
```diff
@@ -1875,7 +1875,8 @@ dynamize_sse_specification({enabled, Enabled}) when is_boolean(Enabled) ->
                             {local_secondary_indexes, local_secondary_indexes()} |
                             {global_secondary_indexes, global_secondary_indexes()} |
                             {sse_specification, sse_specification()} |
-                            {stream_specification, stream_specification()}.
+                            {stream_specification, stream_specification()} |
+                            {provisioned_throughput, {read_units(), write_units()}}.
```
nalundgaard, line 1879: picky, but can you put this between `global_secondary_indexes` and `sse_specification`? I think these are ordered lexically—or intended to be.

test/erlcloud_aws_tests.erl
paulo-ferraz-oliveira: Done in 7f77b78.

I tried to revert to `ok`, prior to the change, and tests pass.
With `ok` but performing the function name changes the tests pass.

I'll look further into it, since we are getting ok as a result, for sure.

paulo-ferraz-oliveira: It's coming from line 19:
```
meck:expect(erlcloud_httpc, request, fun(_,_,_,_,_,_) -> {ok, {{200, "OK"}, [], ok}} end)
                                                                                ^ this!
```

nalundgaard: Hah! wow. Ever feel like.... you're being mocked? 😉

include/erlcloud_ddb2.hrl
paulo-ferraz-oliveira: I hadn't pushed it yet (I didn't know you'd be looking at this now :D). Sorry about that. It's pushed.

test/erlcloud_aws_tests.erl
paulo-ferraz-oliveira: I made it more visible (easier to find) in c187237. I used example <<"OkBody">> since it was already present elsewhere.

paulo-ferraz-oliveira: > Hah! wow. Ever feel like.... you're being mocked? 😉

Many times :D

rebar.config
paulo-ferraz-oliveira: > want rebar2 to use it

I don't :D but I understand you might. I'll get on it.

paulo-ferraz-oliveira: Pushed in 15984c6

paulo-ferraz-oliveira:
> It should be present: https://github.com/erlcloud/erlcloud/blob/e4d903c/rebar.config.script#L8

It seems to be behaving better now. Let's wait on Travis CI.

include/erlcloud_ddb2.hrl
paulo-ferraz-oliveira: On it.

paulo-ferraz-oliveira: Check eb92f35

include/erlcloud_ddb2.hrl
paulo-ferraz-oliveira: Same as previous comment reply.

src/erlcloud_ddb2.erl
paulo-ferraz-oliveira: Sure. There's another occurrence in that file that is not at all ordered alphabetically, though.

paulo-ferraz-oliveira: 09bbab1

paulo-ferraz-oliveira:
> I think the current checks might be somewhat redundant. Shouldn't we just run `compile` (with "warnings as errors"), `dialyze` and `eunit`? (in `rebar3` the last two already do `compile`)
> (sure, they'll run in Travis CI, but maybe there's no need for such repetitiveness)

Any thoughts on this?

include/erlcloud_ddb2.hrl
```diff
@@ -361,7 +361,7 @@
          provisioned_throughput :: undefined | #ddb2_provisioned_throughput{},
          table_arn :: undefined | binary(),
          table_creation_date_time :: undefined | number(),
-         table_id :: undefined | number(),
+         table_id :: undefined | number() | binary(),
```
paulo-ferraz-oliveira, line 364: Should `table_id` ever be a number?

include/erlcloud_ddb2.hrl
```diff
@@ -388,7 +388,7 @@
         {global_secondary_indexes :: undefined | [#ddb2_global_secondary_index_info{}],
          local_secondary_indexes  :: undefined | [#ddb2_local_secondary_index_info{}],
          sse_description :: undefined | erlcloud_ddb2:sse_description(),
-         stream_description :: undefined | #ddb2_stream_description{},
+         stream_description :: undefined | #ddb2_stream_description{} | {boolean(), erlcloud_ddb2:stream_view_type()},
```
paulo-ferraz-oliveira, line 391: Does this seem OK? I adapted it for tests to dialyze properly, but am having seconds thoughts.

include/erlcloud_ecs.hrl
```diff
@@ -38,14 +38,14 @@
 }).
 
 -record(ecs_deployment, {
-    created_at :: undefined | pos_integer(),
+    created_at :: undefined | pos_integer() | float(),
```
paulo-ferraz-oliveira, line 41: This seems wrong, and I guess I did it to have tests not complain.

include/erlcloud_ddb2.hrl
nalundgaard: I think this is an acceptable input, but I don't think it's correct output.

nalundgaard: (probably 1 more lazy meck, and that one may actually be _my_ fault).

paulo-ferraz-oliveira: I'm going to take a closer look at this.

paulo-ferraz-oliveira: I'll change the output (it'll be a breaking change, in any case, because the interface will change) to respect "I don't think it's correct output". Once you check it, we can probably revert it if you don't like the change. e08033d

include/erlcloud_ecs.hrl
paulo-ferraz-oliveira: I'm going to fix the tests for this. It'll revert an API change that I did, but it's not a breaking change, since the expectation was already previously a `pos_integer()`.

paulo-ferraz-oliveira: Done in ba02c27

include/erlcloud_ddb2.hrl
paulo-ferraz-oliveira: Done in 882fe1f

paulo-ferraz-oliveira:
> > I think the current checks might be somewhat redundant. Shouldn't we just run `compile` (with "warnings as errors"), `dialyze` and `eunit`? (in `rebar3` the last two already do `compile`)
> > (sure, they'll run in Travis CI, but maybe there's no need for such repetitiveness)
> 
> Any thoughts on this?

Reduction to essentials was done in d048675

src/erlcloud_ddb2.erl
```diff
@@ -862,13 +870,15 @@ undynamize_stream_view_type(<<"NEW_IMAGE">>, _) -> new_image;
 undynamize_stream_view_type(<<"OLD_IMAGE">>, _) -> old_image;
 undynamize_stream_view_type(<<"NEW_AND_OLD_IMAGES">>, _) -> new_and_old_images.
 
--spec undynamize_stream_specification(jsx:json_term(), undynamize_opts()) -> stream_specification().
+-spec undynamize_stream_specification(jsx:json_term(), undynamize_opts()) -> stream_specification_out().
 undynamize_stream_specification(Json, Opts) ->
     case proplists:get_value(<<"StreamEnabled">>, Json, false) of
         false ->
-            false;
```
nalundgaard, line 869: Ah, this one is actually for backward compatibility, as the initial implementation of the interface was naive, see https://github.com/erlcloud/erlcloud/pull/195 ([here](https://github.com/erlcloud/erlcloud/pull/195/files#diff-7e436c0c4b4861e9228d1dfd9bdf109eR54)). It looks like `#ddb2_stream_description` was introduced in https://github.com/erlcloud/erlcloud/pull/499 for DDB backup APIs ([here](https://github.com/erlcloud/erlcloud/commit/b70efdcdbac1f8723ae920b1731d512dd8eb9618#diff-7e436c0c4b4861e9228d1dfd9bdf109eR294-R303)). I think we should probably not touch the core describe_table API, I know for sure it would break our code internally. Sorry for not investigating further before.

include/erlcloud_ddb2.hrl
nalundgaard: Answered [here](https://github.com/erlcloud/erlcloud/pull/650#discussion_r452184237), sorry about that.

nalundgaard:
> > I think the current checks might be somewhat redundant. Shouldn't we just run `compile` (with "warnings as errors"), `dialyze` and `eunit`? (in `rebar3` the last two already do `compile`)
> > (sure, they'll run in Travis CI, but maybe there's no need for such repetitiveness)
> 
> Any thoughts on this?

I endorse reducing pointless repetition.

src/erlcloud_ddb2.erl
paulo-ferraz-oliveira: I'll revert the change to the change, since it seems acceptable.

paulo-ferraz-oliveira: I deleted the commit (it's as if it never existed... poof!).

paulo-ferraz-oliveira:
I think I'm done. If this Travis CI run passes, I'll leave it up to you to decide on merging this right away or not.

(also, if there's no impediment to it, could you release a new version so we can benefit from these changes, especially with all the adaptations to OTP 23? Thanks much)

nalundgaard:
Thanks so much @paulo-ferraz-oliveira! This is a really cool improvement, and I'm really impressed by how much inconsistency/copy-pasta had gotten through without this strictness. Definitely evidence of its value. I'll give this a final (I hope) look later today, and I'll also ask @motobob and @kkuzmin if they want to take a glance at this. It is a bit large of a change for me to just merge alone, but I agree we should get this in ASAP due to its scope.

test/erlcloud_ecs_tests.erl
```diff
@@ -174,14 +174,14 @@ create_service_input_tests(_) ->
         },
         \"deployments\": [
           {
-            \"createdAt\": 1430326887.362,
+            \"createdAt\": 1430326887362,
```
nalundgaard, line 177: It looks like these should be left as-is. Amazon is real weird with these, and they do it with a number of APIs. probably the spec should be `number()` rather than `float()`, etc.

https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_CreateService.html#API_CreateService_Example_2_Response

paulo-ferraz-oliveira: Sure, it seems difficult to rely on `erlcloud`'s reply though, since UNIX timestamp 1430326887.362 means `Sat Jan 17 1970 14:18:46`, while 1430326887362 means `Wed Apr 29 2015 18:01:27`. Is there a function to convert this to a readable format, somewhere in `erlcloud`'s API?

paulo-ferraz-oliveira: In any case, I'm reverting the change.

paulo-ferraz-oliveira: 541f36d

nalundgaard: > Sure, it seems difficult to rely on `erlcloud`'s reply though, since UNIX timestamp 1430326887.362 means `Sat Jan 17 1970 14:18:46`, while 1430326887362 means `Wed Apr 29 2015 18:01:27`. Is there a function to convert this to a readable format, somewhere in `erlcloud`'s API?

I am a little confused here—or rather, I think you might be confusing epoch seconds and epoch milliseconds.

I got this from [epochconverter.com](https://www.epochconverter.com/batch):

Epoch date | Human-readable date (GMT)
-- | --
1430326887 | Wed, 29 Apr 2015 17:01:27 +0000
1430326887362 | 03 April 47295 03:16:02 GMT-5

(note: had to use a different converter for `1430326887362` since it refuses to convert it)

I don't think there should be any need to have a utility in erlcloud—it looks like you get back an epoch timestamp (in seconds) as a float, with 2-3 decimal places of precision if you want to account for that. It's OK as an exercise left for the reader, IMO,

paulo-ferraz-oliveira: You're more or less right about "you might be confusing epoch seconds and epoch milliseconds"; I know the _epoch_ concept, but I messed up converting the value from the test files.

I did use `new Date(_)` from the JavaScript console, but forgot the base unit for that is milliseconds, so should have multiplied by 1000 [first error]. Since my original change to the (test) code was simply removing the `.` I further got convinced that AWS was wrong, which it kind of is, depending on the "point of view" [second error].

The correct way is
```
> new Date(1430326887000)
< Wed Apr 29 2015 18:01:27 GMT+0100
```

In any case, UNIX timestamp (unless AWS redefines it - I can't find references to how they define the **Timestamp** type) is a concept in seconds[1]. I've never heard about _epoch milliseconds_, but am willing to accept it in the scope of this Pull Request.

[1] for lack of a better alternative, here is the definition I'm used to: https://en.wikipedia.org/wiki/Unix_time ("It is the number of seconds...").

To be even more specific, the reference I use is, in Linux,
```
date -j -f "%a %b %d %T %Z %Y" "`date`" "+%s"
```

> It's OK as an exercise left for the reader, IMO.



Would it be acceptable to introduce a comment stating what is obvious for you but might not be to the consumer (e.g. `% epoch timestamp, in seconds`), or do you prefer to leave it as is?

nalundgaard: Sure, a comment is fine, but I think the intent is for the documentation to [link to the AWS API Reference](https://github.com/erlcloud/erlcloud/blob/541f36d0f5b6c9ac4c81ecccfe483bd9a469923a/src/erlcloud_ecs.erl#L1264) corresponding to each function, so that you can see what values AWS gives you.

paulo-ferraz-oliveira:
Sorry about the late commit. I just remembered we wanted to start testing on Erlang/OTP 23, also.

nalundgaard:
Chatted with @motobob separately, he is OK with my approval, so I am merging this. I don't think I'll release a hex package for this unless you see a need for that. Thanks again!

paulo-ferraz-oliveira:
> I don't think I'll release a hex package for this unless you see a need for that. Thanks again!

I do have a need for it :) [not specifically this PR but the 3 previous ones] I'm updating an internal lib. to run on OTP 23 and would be grateful if a tag were available.

nalundgaard:
OK, published [3.3.3](https://github.com/erlcloud/erlcloud/releases/tag/3.3.3) ([hex](https://hex.pm/packages/erlcloud/3.3.3)).

Issue #652, opened 2020-07-17
Support SES requests with Signature v4


enerick:
AWS has announced its ending of support for SES APIs using Signature v3 on September 30, 2020.
https://docs.aws.amazon.com/ses/latest/DeveloperGuide/using-ses-api-authentication.html

It seems that erlcound_ses is still using Signature v3, so we should migrate to v4.
https://github.com/erlcloud/erlcloud/blob/0f0540041ad6b55700715d1d3629853bf3158791/src/erlcloud_ses.erl#L1256-L1260

nalundgaard:
Thanks for the note! I am surprised that our org hadn't noticed that yet via AWS notice. I'll look at fixing it soon.

enerick:
Thanks!

PR #655, opened 2020-07-27
Switch SES API to Signature v4

gproskurin:
Fixes https://github.com/erlcloud/erlcloud/issues/652

### DIFF
--- a/src/erlcloud_ses.erl
+++ b/src/erlcloud_ses.erl
@@ -1242,38 +1242,10 @@ decode_error(Doc) ->
 %%%------------------------------------------------------------------------------
 
 ses_request(Config, Action, Params) ->
-    case erlcloud_aws:update_config(Config) of
-        {ok, Config1} ->
-            ses_request_no_update(Config1, Action, Params);
-        {error, Reason} ->
-            {error, Reason}
-    end.
-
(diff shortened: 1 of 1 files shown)


src/erlcloud_ses.erl
```diff
@@ -1242,38 +1242,10 @@ decode_error(Doc) ->
 %%%------------------------------------------------------------------------------
 
 ses_request(Config, Action, Params) ->
-    case erlcloud_aws:update_config(Config) of
```
nalundgaard, line 1245: Please don't delete this.

gproskurin: You mean, keep call to `update_config`? It is called in `aws_request4`

https://github.com/erlcloud/erlcloud/blob/0f0540041ad6b55700715d1d3629853bf3158791/src/erlcloud_aws.erl#L184


And this is similar to what SQS code does:
https://github.com/erlcloud/erlcloud/blob/0f0540041ad6b55700715d1d3629853bf3158791/src/erlcloud_sqs.erl#L677-L678

nalundgaard: Oh, nice! I didn't realize that. Thanks

gproskurin:
Please see links in discussion above. `update_config` is done in `aws_request4`

PR #656, opened 2020-09-02
Use NIF based `base64` and JSON decoders for better performance

danilagamma:
Intent of this PR is to improve performance of base64 and JSON encoding/decoding operations while keeping same interface.

Big downside of this is that `erlcloud` would now have non-pure `erlang` dependencies which might now work on all platforms.

Unfortunately there's still a huge gap between pure-erlang decoders and NIF based ones, so alternative parser libraries would not yield that amount of performance benefit.


posilva:
For a PR like this one that may have a huge impact in the community using a NIF instead a pure-erlang I would ask for:
- Good benchmarks to assess what are we gaining 
- Think about an alternative  solution that will allow the consumer to select the json parser encode/decode  (callback, macro with define at compilation time, persistent_term configuration to switch between NIF and JSX or JSONE)
- A good explanation of the use case that found the need to optimize the json parser

nalundgaard:
I think that this PR is being made for the purpose of facilitating some real-world performance analysis, so I assume we can look forward to that. In our experience using these libraries in other contexts, I expect the benefits to be quite substantial for a number of use cases around DynamoDB and Kinesis.

That said, I do kind of suspect that, given the concerns with NIF, we should probably figure out a way to allow BYO decoders (within the limits of basic compatibility with JSX's weird interface) using one of the methods that @posilva suggested.

danilagamma:
I agree that optional compilation or other means of enabling NIF-decoders would be great. 

Is there a convenient way for doing something like that? Lets say I want to keep old behaviour of erlang-native `jsx` and `base64` decoders as it was, but leave an option for my application that pulls `erlcloud` as dependency to enable it via additional compilation flag for example, is there a mechanism in `rebar` for doing something like that?

nalundgaard:
The closest thing I can think of is our [hackney support](https://github.com/erlcloud/erlcloud/blob/b30e563/include/erlcloud_aws.hrl#L144-L148), which is baked into the `#aws_config{}` record. That seems like the most consistent with the library "paradigm."

danilagamma:
@nalundgaard @posilva I've pushed a commit that makes the whole thing conditional.
If `ERLCLOUD_NIF_DECODERS` env variable is set to `true` - then `erl_opts` would be updated with `ENABLE_NIF_DECODERS` declaration and deps for `jiffy` and `b64fast` would be added. Do you guys think this is viable solution?

nalundgaard:
It's an interesting implementation, but what about the app.src? Does this also need an [app.src.script](https://www.rebar3.org/docs/dynamic_configuration/) file to fix the OTP deps? That would effectively make it rebar3-only, although that's probably OK-ish, since AFAIK it's really only rebar3/relx that are real sticklers about using OTP deps to define what is included in an OTP release.

This still seems a little complicated/fragile, but does look optimal in terms of run-time performance. Do you have an example of how to configure this when including it as a dep in a downstream project?

src/erlcloud_json.erl
```diff
@@ -8,6 +13,52 @@
...
+    {[ {K, jsx_to_jiffy(V)} || {K, V} <- Obj ]};
+jsx_to_jiffy(List) when is_list(List) ->
+    [ jsx_to_jiffy(V) || V <- List ];
+jsx_to_jiffy(V) ->
+    V.
+
+-else.
+
+decode_bin(Binary) ->
+    jsx:decode(Binary).
```
paulo-ferraz-oliveira, line 52: For compatibility with `jsx` 3.0 (and libraries that consume it, "higher up") this should actually be `[{return_maps, false}]`, otherwise (since 3.0 broke interface explicitly) this will be a real issue for consumers.

motobob:
given Erlang 24/25 and etc closing this and will re-evaluate further

PR #657, opened 2020-09-09
Add support for sending raw emails via SES

carlisom:
Adding support for sending raw emails to allow for erlcloud to send attachments via SES. 

Keep the implemented function options the same as other ses functions and did not add options for the following:
* ConfigurationSetName
* FromArn
* ReturnPathArn
* SourceArn
* Tags.member.N

### DIFF
--- a/src/erlcloud_ses.erl
+++ b/src/erlcloud_ses.erl
@@ -12,7 +12,6 @@
 %%  * ListIdentityPolicies
 %%  * ListVerifiedEmailAddresses (deprecated; use ListIdentities)
 %%  * PutIdentityPolicy
-%%  * SendRawEmail
 %%  * VerifyEmailAddress (deprecated; use VerifyEmailIdentity)
 %%
 %% @end
@@ -43,6 +42,8 @@
 -export([send_custom_verification_email/2, send_custom_verification_email/3]).
 
 -export([send_email/4, send_email/5, send_email/6]).
+
+-export([send_raw_email/1, send_raw_email/2, send_raw_email/3]).
 
 -export([set_identity_dkim_enabled/2, set_identity_dkim_enabled/3]).
 -export([set_identity_feedback_forwarding_enabled/2, set_identity_feedback_forwarding_enabled/3]).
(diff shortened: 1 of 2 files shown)


src/erlcloud_ses.erl
```diff
@@ -1093,6 +1160,12 @@ encode_destination(ToAddress, Acc) when is_list(ToAddress); is_binary(ToAddress)
     %% Single entry
     encode_destination_pairs([{to_addresses, [ToAddress]}], Acc).
 
+encode_destinations({destination_addresses, Destinations}, Acc) ->
```
nalundgaard, line 1163: why `destination_addresses` and not `destinations`?

src/erlcloud_ses.erl
```diff
@@ -732,6 +733,68 @@ send_email(Destination, Body, Subject, Source, Opts, Config) ->
...
+%% {ok, [{message_id, "00000131d51d2292-159ad6eb-077c-46e6-ad09-ae7c05925ed4-000000"}]} =
+%%     erlcloud_ses:send_raw_email(<<"From: b@from.com\nTo: a@to.com\nSubject: Subject\nMIME-Version: 1.0\nContent-type: Multipart/Mixed; boundary=\"NextPart\"\n\n--NextPart\nContent-Type: text/plain\n\nEmail Body\n\n--NextPart--">>, []).
+%% '
+%%
+%% All supported inputs.
+%%
+%% `
+%%  {ok, [{message_id, "00000131d51d2292-159ad6eb-077c-46e6-ad09-ae7c05925ed4-000000"}]} =
+%% erlcloud_ses:send_email(<<"To: d@to.com\nCC: c@cc.com\nBCC: a@bcc.com, b@bcc.com\nSubject: Subject\nMIME-Version: 1.0\nContent-type: Multipart/Mixed; boundary=\"NextPart\"\n\n--NextPart\nContent-Type: text/plain\n\nEmail Body\n\n--NextPart--">>,
+%%                         [{destinations,{destination_addressed, [<<"a@bcc.com">>, "b@bcc.com", <<"c@cc.com">>, "d@to.com"]}},
```
nalundgaard, line 779: destination_addressed looks wrong

src/erlcloud_ses.erl
carlisom: I have `destinations` as the top level option and used `destination_addresses` to be similar to `to_addresses, cc_addresses, bcc_addresses` in the send_email functionality.

I could rework the function to accept either a single email address or a list of email addresses passed in via the tuple `{destinations, [emails()]}`. The main issues I am trying to solve for is knowing a list of addresses vs a string address.

src/erlcloud_ses.erl
```diff
@@ -1128,6 +1201,12 @@ encode_opts([{reply_to_addresses, List} | T], Acc) ->
 encode_opts([{return_path, ReturnPath} | T], Acc) ->
     encode_opts(T, [{"ReturnPath", ReturnPath} | Acc]).
 
+encode_raw_opts([], Acc) ->
+    Acc;
+encode_raw_opts([{source, Source} | T], Acc) when is_list(Source); is_binary(Source) ->
+    encode_raw_opts(T, [{"Source", Source} | Acc]);
+encode_raw_opts([{destinations, Destination} | T], Acc) ->
+    encode_raw_opts(T, encode_destinations(Destination, Acc)).
```
nalundgaard, line 1209: Why do you pass Acc into `encode_destinations(Destination, Acc)`?

src/erlcloud_ses.erl
carlisom: The issues is [here](https://github.com/erlcloud/erlcloud/pull/657/files#diff-c45cdf8adff42f1b80f94fd8c2e2be6eR742-R745).
```-type send_raw_email_destinations() :: email() | {destination_addresses, emails()}.

-type send_raw_email_opt() :: {source, email()} |
                              {destinations, send_raw_email_destinations()}.
```
supporting `send_raw_email_destinations()` to be `email() | {destination_addresses, emails()}` or `email() | {destination_addresses, [emails()]}`

src/erlcloud_ses.erl
nalundgaard: NVM, I see the point there.

carlisom: So the values can be added to the current `Acc` but I will rework this to have the values returned from `encode_destinations` to be the head of the `Acc` and return `[encode_destinations(Destination) | Acc]`

src/erlcloud_ses.erl
nalundgaard: I get that `email()` can be string or binary, which sucks, but it's not too hard to pattern match string/binary vs. list of string/binary:

```erlang
encode_destinations([Dest|_T] = Destinations, Acc) when is_list(Dest); is_binary(Dest) ->
    encode_list("Destinations", Destinations, Acc);
encode_destinations(ToAddress, Acc) when is_list(ToAddress); is_binary(ToAddress) ->
    encode_destinations([ToAddress], Acc).
```

Seems like it's just awkward to write `[{destinations, {destination_addresses, [...]}}]` when it's pretty straightforward to discriminate between the 2 classes. If we supported `iolist()` that would be another story, but I don't think there is a need there.

src/erlcloud_ses.erl
carlisom: I can do this but this breaks from the style of the module where functions always are passes the `Acc`
Example [here](https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_ses.erl#L1012-L1019)
```erlang
encode_params([], Acc) ->
    lists:reverse(Acc);
encode_params([{body, Body} | T], Acc) ->
    encode_params(T, encode_body(Body, Acc));
encode_params([{destination, Destination} | T], Acc) ->
    encode_params(T, encode_destination(Destination, Acc));
encode_params([{email_address, EmailAddress} | T], Acc) when is_list(EmailAddress); is_binary(EmailAddress) ->
    encode_params(T, [{"EmailAddress", EmailAddress} | Acc]);
```

When a function is called, the current `Acc` is currently passed in vs when adding a sing value to the `Acc`

nalundgaard (approved):
LGTM

Issue #659, opened 2020-09-18
SES configuration sets


wellknownmelodies:
Doesn't look like it's possible to specify [configuration sets](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/using-configuration-sets.html) or associated tags with erlcloud_ses:send_email. Happy to submit a pull request adding these options to `send_email_opt`; are there guidelines for contribution beyond what's in the README?

thanks!

nalundgaard:
PRs welcome, of course! I generally prefer to keep option specifications in lexical order, and I think that is mostly if not all the case in the `erlcloud_ses` module. You will also need to make sure that it's encoded properly, any you probably also want to handle errors like `ConfigurationSetDoesNotExist` and `ConfigurationSetSendingPaused` in [`decode_error`](https://github.com/erlcloud/erlcloud/blob/be597d9/src/erlcloud_ses.erl#L1286-L1312) as well.

Are you just planning to add support for the `ConfigurationSetName` option from the [Request Parameters](https://docs.aws.amazon.com/ses/latest/APIReference/API_SendEmail.html#API_SendEmail_RequestParameters)? That's fine by me, I suggest to call it `configuration_set_name` in `send_email_opts()` for consistency with the other ones (titlecase -> snake case in general, as you see in the code base).

Lastly, if you add a new option, please make sure it's covered in the [unit tests](https://github.com/erlcloud/erlcloud/blob/be597d9/test/erlcloud_ses_tests.erl#L305-L351).

carlisom:
resolved by #660

PR #660, opened 2020-09-22
add support for ses config sets and tags

wellknownmelodies:
Address #659. Only including support for specifying the values in send_email calls (and not for example functions for adding and managing configuration sets) in this PR.


nalundgaard (approved):
LGTM. @carlisom can you look at this too?

carlisom (approved):
LGTM

PR #662, opened 2020-10-02
erlcloud_application_autoscaler: return error tuple when '#aws_request.response_type' is not 'ok'

va17:
Internal function `erlcloud_application_autoscaler:request_with_action/3` currently ignores any error response from AWS other than config update failures. It currently returns `{ok, {error, _}}` and throws function clause error in some actions like [here](https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_application_autoscaler.erl#L531-L533). This change updates the code to check `#aws_request.response_type` and return `{ok, _}`/`{error, _}` appropriately.

Update (10-08-2020):
Updated the branch to decode auto scaling application errors and return them in the format of `{error, {Type, Message}}`

**Note**: `erlcloud_application_autoscaler` module does not have any test coverage hence why I did not update the tests for this change.

### DIFF
--- a/src/erlcloud_application_autoscaler.erl
+++ b/src/erlcloud_application_autoscaler.erl
@@ -702,7 +702,12 @@ request_with_action(Configuration, BodyConfiguration, Action) ->
             Headers = [{"content-type", "application/x-amz-json-1.1"} | HeadersPrev],
             Request = prepare_record(Config, post, Headers, Body),
             Response = erlcloud_retry:request(Config, Request, fun aas_result_fun/1),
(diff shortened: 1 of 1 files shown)


nalundgaard (commented):
@paulo-ferraz-oliveira @Santiniis @motobob  Do you have any objections to this change? tests on this module are still to-do, but we've hit this basic issue using this library, and the way it works as-is is awful when AWS returns an error to you.

nalundgaard (approved):

nalundgaard:
Going to assume yes. Please register your complaints ex post facto.

PR #663, opened 2020-10-14
erlcloud_application_autoscaler: fix error decoder by verifying if the response contains a json body

va17:
https://github.com/erlcloud/erlcloud/pull/662 introduced the error decoder but there is a bug in there where it assumes there will always be a JSON in the response body. This PR fix error decoder by verifying if the response contains a JSON body before attempting to decode it.
This PR also:
* Updates `ok_error_response` spec type accordingly
* Returns errors that are not decodable in the format of `erlcloud_aws:httpc_result_error()`
* Fix issue where errors are returned as `{error, {error, ...}}`

@nalundgaard

### DIFF
--- a/src/erlcloud_application_autoscaler.erl
+++ b/src/erlcloud_application_autoscaler.erl
@@ -61,10 +61,13 @@
 -type response_attribute() :: string() | integer().
 -type response_key() :: atom().
 -type response() :: [{response_key(), response_attribute()}].
+-type error_type() :: binary().
+-type error_message() :: binary().
 -type ok_error_response() :: {ok, jsx:json_term()}
                            | {error, metadata_not_available
                                    | container_credentials_unavailable
-                                   | erlcloud_aws:httpc_result_error()}.
+                                   | erlcloud_aws:httpc_result_error()
+                                   | {error_type(), error_message()}}.
 
 -type aws_aas_request_body() :: proplists:proplist().
 
@@ -706,7 +709,7 @@ request_with_action(Configuration, BodyConfiguration, Action) ->
                 ok ->
                     {ok, jsx:decode(Response#aws_request.response_body, [{return_maps, false}])};
                 _ ->
-                    {error, decode_error(Response)}
+                    decode_error(Response)
             end;
         {error, Reason} ->
             {error, Reason}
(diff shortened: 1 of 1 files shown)


src/erlcloud_application_autoscaler.erl
```diff
@@ -741,8 +741,13 @@ headers(Config, Operation, Body) ->
...
 %% Example: {error, {<<"ConcurrentUpdateException">>, <<"You already have a pending update to an Auto Scaling resource.">>}}
-decode_error(Response) ->
-    DecodedError = jsx:decode(Response#aws_request.response_body, [{return_maps, false}]),
-    ErrorType = proplists:get_value(<<"__type">>, DecodedError, <<>>),
-    ErrorMessage = proplists:get_value(<<"Message">>, DecodedError, <<>>),
-    {error, {ErrorType, ErrorMessage}}.
+decode_error(#aws_request{response_body = Body}) ->
+    case jsx:is_json(Body) of
+        false ->
+            {error, {<<>>, <<>>}};
```
nalundgaard, line 747: Here can you actually return `{error, {http_error, Status, StatusLine, RespBody}}`? Should be able to get those all from the `#aws_request`.

va17: Updated the PR to return the following:
```erlang
{error, {Response#aws_request.httpc_error_reason, Response#aws_request.response_status, Response#aws_request.response_status_line, Body}};
```

src/erlcloud_application_autoscaler.erl
```diff
@@ -61,10 +61,14 @@
...
 -type response_key() :: atom().
 -type response() :: [{response_key(), response_attribute()}].
+-type error_type() :: binary().
+-type error_message() :: binary().
 -type ok_error_response() :: {ok, jsx:json_term()}
                            | {error, metadata_not_available
                                    | container_credentials_unavailable
-                                   | erlcloud_aws:httpc_result_error()}.
+                                   | erlcloud_aws:httpc_result_error()}
+                           | {error, {error_type(), error_message()}}
```
nalundgaard, line 70: Please use the pattern here, not this redundant thing.

va17: Fixed.

nalundgaard (approved):

PR #654, opened 2020-07-21
Add some support for AWS Secrets Manager

kennedyjosh:
Implemented the GetSecretValue function and wrote tests for it. For now only have that one function, but a framework for adding future Secrets Manager functions exists through these changes

PR #667, opened 2020-11-13
erlcloud_sm: fix use of jsx

nalundgaard:
Follow up to https://github.com/erlcloud/erlcloud/pull/645. I missed that I didn't catch this regression when `erlcloud_sm` was introduced in https://github.com/erlcloud/erlcloud/pull/654.

This change just adds the eplicit `[{return_maps, false}]` option to the `jsx:decode(...)` call to ensure that it comes back in the expected format, regardless of which jsx version the client uses.

cc @paulo-ferraz-oliveira.


paulo-ferraz-oliveira:
Thanks.

motobob (approved):

PR #668, opened 2020-11-17
Update how VPC endpoints are calculated

motobob:
current approach was based of manual setting of env var. 
while working on CFN via `{"Fn::GetAtt": ["CWLogsEndpoint", "DnsEntries"]` gets both "ID:Name"

support that ^.

example CFN:
`Z1xxx:vpce-xxxx-427maydd.logs.us-west-2.vpce.amazonaws.com,Z1xxx:vpce-xxxxx-427maydd-us-west-2b.logs.us-west-2.vpce.amazonaws.com,Z1xxx:vpce-xxxx-427maydd-us-west-2a.logs.us-west-2.vpce.amazonaws.com,Z1xxxx:vpce-xxxx-427maydd-us-west-2c.logs.us-west-2.vpce.amazonaws.com,Z0xxxx:logs.us-west-2.amazonaws.com`

### DIFF
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -824,11 +824,14 @@ get_host_vpc_endpoint(Service, Default) when is_binary(Service) ->
         {env, EnvVarName} when is_list(EnvVarName) ->
             % ignore "" env var or ",," cases
             % also handle "zoneID:zoneName" form when it comes from CFN
-            Es = string_split(string_split(os:getenv(EnvVarName, ""), ","), ":"),
+            Es = string_split(os:getenv(EnvVarName, ""), ","),
             lists:filtermap(
-                fun ([""]) -> false;
-                    ([Name]) -> {true, list_to_binary(Name)};
-                    ([_Id, Name]) -> {true, list_to_binary(Name)}
+                fun ("") -> false;
+                    (Value) ->
+                        case string_split(Value, ":") of
+                            [_Id, Name] -> {true, list_to_binary(Name)};
+                            [Name] ->{true, list_to_binary(Name)}
+                        end
                 end,
                 Es
             );
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -822,9 +822,16 @@ get_host_vpc_endpoint(Service, Default) when is_binary(Service) ->
     %% resolve through ENV if any
     Endpoints = case ConfiguredEndpoints of
         {env, EnvVarName} when is_list(EnvVarName) ->
-            Es = string_split(os:getenv(EnvVarName, ""), ","),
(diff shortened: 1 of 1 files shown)


motobob:
test it

src/erlcloud_aws.erl
```diff
@@ -822,9 +822,16 @@ get_host_vpc_endpoint(Service, Default) when is_binary(Service) ->
     %% resolve through ENV if any
     Endpoints = case ConfiguredEndpoints of
         {env, EnvVarName} when is_list(EnvVarName) ->
-            Es = string_split(os:getenv(EnvVarName, ""), ","),
             % ignore "" env var or ",," cases
-            [list_to_binary(E) || E <- Es, E /= ""];
+            % also handle "zoneID:zoneName" form when it comes from CFN
+            Es = string_split(string_split(os:getenv(EnvVarName, ""), ","), ":"),
```
kkuzmin, line 827: we need to `split_string(':')` every string returned by `split_string(',')`.

kkuzmin (changes requested):

src/erlcloud_aws.erl
motobob: oh!

kkuzmin: I mean `string_split()` returns a list of strings and doesn't accept a list of strings as input.

motobob: fixed

src/erlcloud_aws.erl
```diff
@@ -822,9 +822,19 @@ get_host_vpc_endpoint(Service, Default) when is_binary(Service) ->
...
             % ignore "" env var or ",," cases
-            [list_to_binary(E) || E <- Es, E /= ""];
+            % also handle "zoneID:zoneName" form when it comes from CFN
+            Es = string_split(os:getenv(EnvVarName, ""), ","),
+            lists:filtermap(
+                fun ("") -> false;
+                    (Value) ->
+                        case string_split(Value, ":") of
+                            [_Id, Name] -> {true, list_to_binary(Name)};
+                            [Name] -> {true, list_to_binary(Name)}
```
kkuzmin, line 833: Do we want a default case clause here with `false`? Or throw an exception intentionally?

motobob: ? support for current/old way of just names in the list is still neded

kkuzmin: I mean `split_string` returns a list fo 3 and more elements

kkuzmin (approved):

src/erlcloud_aws.erl
motobob: case clause is a good exception

PR #671, opened 2020-12-05
Add RoleLastUsed field to erlcloud_iam:get_role response

### LINKED ISSUE #670: erlcloud_iam:get_role response does not include RoleLastUsed
Hi,

I noticed that `erlcloud_iam:get_role` method that calls the [GetRole](https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetRole.html) AWS API doesn't include the "RoleLastUsed" information. Would it be possible to add this?

The info is present in AWS's documentation and I did a spike of this and could serialize the field. I can submit a PR if that works.

Thanks

aaronlelevier:
**Summary**
This PR adds the RoleLastUsed field from the GetRole AWS API endpoint to the erlcloud_iam:get_role response

**Issue**
fixes #670

**References**
AWS API docs reference for GetRole Endpoint:

https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetRole.html#API_GetRole_Examples

### DIFF
--- a/src/erlcloud_iam.erl
+++ b/src/erlcloud_iam.erl
@@ -989,9 +989,6 @@ data_type("Role") ->
      {"RoleName", role_name, "String"},
      {"Path", path, "String"},
      {"RoleLastUsed", role_last_used, data_type("RoleLastUsed")}];
-data_type("RoleLastUsed") ->
-    [{"LastUsedDate", last_used_date, "DateTime"},
-     {"Region", region, "String"}];
 data_type("RoleDetail") ->
     [{"RolePolicyList/member", role_policy_list, data_type("PolicyDetail")},
      {"RoleName", role_name, "String"},
@@ -1001,6 +998,9 @@ data_type("RoleDetail") ->
      {"CreateDate", create_date, "DateTime"},
      {"AssumeRolePolicyDocument", assume_role_policy_document, "Uri"},
      {"Arn", arn, "String"}];
+data_type("RoleLastUsed") ->
+    [{"LastUsedDate", last_used_date, "DateTime"},
+     {"Region", region, "String"}];
 data_type("RolePolicyList") ->
     [{"PolicyDocument", policy_document, "Uri"},
      {"RoleName", role_name, "String"},
--- a/src/erlcloud_iam.erl
+++ b/src/erlcloud_iam.erl
@@ -436,7 +436,7 @@ list_roles(PathPrefix) ->
 list_roles(PathPrefix, #aws_config{} = Config)
   when is_list(PathPrefix) ->
     ItemPath = "/ListRolesResponse/ListRolesResult/Roles/member",
-    iam_query(Config, "ListRoles", [{"PathPrefix", PathPrefix}], ItemPath, data_type("Role")).
+    iam_query(Config, "ListRoles", [{"PathPrefix", PathPrefix}], ItemPath, data_type("RoleList")).
 
 -spec list_roles_all() -> {ok, proplist()} |  {error, any()}.
 list_roles_all() -> list_roles([]).
@@ -450,7 +450,7 @@ list_roles_all(PathPrefix) ->
 list_roles_all(PathPrefix, #aws_config{} = Config)
   when is_list(PathPrefix) ->
     ItemPath = "/ListRolesResponse/ListRolesResult/Roles/member",
-    iam_query_all(Config, "ListRoles", [{"PathPrefix", PathPrefix}], ItemPath, data_type("Role")).
+    iam_query_all(Config, "ListRoles", [{"PathPrefix", PathPrefix}], ItemPath, data_type("RoleList")).
 
 -spec list_role_policies(string()) -> {ok, proplist()} | {ok, proplist(), string()} |  {error, any()}.
 list_role_policies(RoleName) ->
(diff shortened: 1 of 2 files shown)


src/erlcloud_iam.erl
```diff
@@ -981,6 +981,17 @@ data_type("Role") ->
      {"RoleId", role_id, "String"},
      {"RoleName", role_name, "String"},
      {"Path", path, "String"}];
+data_type("RoleGetDetail") ->
```
nalundgaard, line 984: Can you just call this `"RoleDetail"`? Seems like it fits the pattern better.

aaronlelevier: There's already a used by `data_type("RoleDetail")` used by  `data_type("AccountAuthorizationDetails")`. 

I named it as best I could, but wasn't sure. I can change it. Lmk.

nalundgaard: Didn't realize the conflict. This should work then.

nalundgaard: Actually, looking at the API docs, it sure looks like there is only [Role](https://docs.aws.amazon.com/IAM/latest/APIReference/API_Role.html) and [RoleDetail](https://docs.aws.amazon.com/IAM/latest/APIReference/API_RoleDetail.html). It looks like both [GetRole](https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetRole.html) and [ListRoles](https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListRoles.html) use `Role`, and [GetAccountAuthorizationDetails](https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetAccountAuthorizationDetails.html) uses `RoleDetails`. It is curious that the `ListRoles` example doesn't have `RoleLastUsed` in the response, but... I bet it's there? Also, it looks like `RoleLastUsed` is part of the `RoleDetail` object as well. Maybe it would be right to just add the `RoleLastUsed` to both the `Role` and `RoleDetail` `data_type` entries. 

Do you have access to an account with IAM to check? I can do it tomorrow if not. I'd prefer to stick with symmetry to the API object structure if the docs are right.

nalundgaard (changes requested):
Couple of concerns here after reading through the API docs in IAM.

src/erlcloud_iam.erl
aaronlelevier: Yes, I have an AWS Account and checked the API response. `RoleLastUsed` isn't returned in the [ListRoles](https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListRoles.html#API_ListRoles_Examples) API response, and AWS's docs are correct.

Sorry, I think that I should have noted this originally in the PR. That is why I created a new `data_type`.

Please feel free to double check the AWS API response.

I was also a little confused by this. I first checked `boto3.get_role` and there was a lot more info, but I guess the XML API is a little more specific.

test/erlcloud_iam_tests.erl
```diff
@@ -682,6 +684,55 @@ get_login_profile_output_tests(_) ->
...
+        ],
+
+    input_tests(?GET_ROLE_RESP, Tests).
+
+get_role_output_tests(_) ->
+    Tests = [?_iam_test(
+        {"This returns the role",
+            ?GET_ROLE_RESP,
+            {ok, [{role_last_used,
+                [[{region,"us-east-1"},
```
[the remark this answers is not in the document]
aaronlelevier: Okay, the indentation should be fixed. If I still didn't fix it right, please feel free to push a commit to fix it. I am not so familiar with the indentation rules I think. 

Thank you for reviewing.

src/erlcloud_iam.erl
nalundgaard: OK, I think I have some insight at last:

1. You're right that ListRoles does not return `RoleLastUsed`. I would propose renaming the existing `data_type("Role")` entry to `data_type("RoleList")` or something like that, and making the `get_role` handle the "real" one for `data_type("Role")`. 
2. `"RoleDetail"` used in `get_account_authorization_details()` also returns `RoleLastUsed`, so we should use it there. Can you add it in?

nalundgaard (changes requested):
Other than [this](https://github.com/erlcloud/erlcloud/pull/671/files#r537905212), one more suggestion: can you put the entry for `data_type("RoleLastUsed")` _after_ `data_type("RoleDetail")`? I just find that it breaks up the heads for the data types that are "almost the same but different levels of detail.

I also found that get_account_authorization_details is totally broken for pagination, so that's nice. I'll have to open a PR at some point.

src/erlcloud_iam.erl
aaronlelevier: I did bullet point 1 here, and the [comment](https://github.com/erlcloud/erlcloud/pull/671#pullrequestreview-546620125) below.

I didn't do bullet point 2 though because `RoleLastUsed` is actually under based on the [API doc](https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetAccountAuthorizationDetails.html): 
```
<InstanceProfileList>
  <member>
    <InstanceProfileName>EC2role</InstanceProfileName>
    <Roles>
      <member>
```

It's under the `InstanceProfileList` which uses `data_type("InstanceProfile")` and that same `data_type` is linked to a couple of other API endpoints for InstanceProfile that don't have a RoleLastUsed field. I left as-is unless it's worth creating a second `data_type("InstanceProfile")` that includes `RoleLastUsed`? Also not sure the `data_type` name?

nalundgaard: Nah, I guess this is fine. This API is so bizarrely inconsistent.

nalundgaard (approved):
LGTM. @kkuzmin @motobob any objections?

kkuzmin (approved):
LGTM. Thank you @aaronlelevier @nalundgaard

PR #674, opened 2020-12-08
erlcloud_iam: fix get_account_authorization_details

nalundgaard:
When reviewing https://github.com/erlcloud/erlcloud/pull/671, I noticed that IAM [GetAccountAuthorizationDetails](https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetAccountAuthorizationDetails.html) is a paginated API, and when you use it on an account with a paginated response, it crashes rather inelegantly.

This change makes it work correctly. `get_account_authorization_details_all` is left as an exercise for some other time.


src/erlcloud_iam.erl
```diff
@@ -664,17 +664,25 @@ get_instance_profile(ProfileName, #aws_config{} = Config) ->
...
   
--spec get_account_authorization_details(aws_config()) -> {ok, proplist()} |  {error, any()}.
+-spec get_account_authorization_details(list() | aws_config()) -> {ok, proplist()} | {ok, proplist(), string()} | {error, any()}.
 get_account_authorization_details(#aws_config{} = Config) ->
+    get_account_authorization_details([], #aws_config{} = Config);
+get_account_authorization_details(Params) ->
+    get_account_authorization_details(Params, default_config()).
+
+-spec get_account_authorization_details(list(), aws_config()) -> {ok, proplist()} | {ok, proplist(), string()} | {error, any()}.
+get_account_authorization_details(Params, #aws_config{} = Config) ->
```
motobob, line 678: why not add a guard for a `when is_list(Params)` ?

motobob (approved):
suggesting a minor thing

kkuzmin (approved):
thank you @nalundgaard

src/erlcloud_iam.erl
nalundgaard: Done.

PR #676, opened 2020-12-08
Implement GitHub Actions

### LINKED ISSUE #675: Move from Travis CI to GitHub Actions?
Would you be interested in a pull request that moves CI from Travis CI to GitHub Actions, or at least adds GitHub Actions alongside Travis CI? It seems travis-ci.org will stop being useful from Dec 31 2020. I can do it, except that I think you need to create the workflow yml, first, in `master`, after which I can branch off of.

paulo-ferraz-oliveira:
This is currently a _best-effort_ implementation. Since there's no GitHub Action in place yet, it won't run until merged to `master` (unless you want to initialize GitHub Actions for this repo, in `master`, after which I can rebase this).

I'm **mostly** sure that `ci.yml` will work as-is.

`publish.yml`... I'm not so sure of, but it's not fancy code, so I'm _confident_ 😄.

Closes #675.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,25 @@
+---
+name: build
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+jobs:
+  ci:
+    name: Run checks and tests over ${{matrix.otp_vsn}} and ${{matrix.os}} (rebar2? ${{matrix.force_rebar2}})
+    runs-on: ${{matrix.os}}
+    strategy:
+      matrix:
+        otp_vsn: [19.3, 20.3, 21.3, 22.3, 23.0]
+        os: [ubuntu-latest]
+        force_rebar2: [true, false]
+    steps:
+      - uses: actions/checkout@v2
+      - run: export FORCE_REBAR2=${{matrix.force_rebar2}}
+      - run: make travis-install
+      - run: make warnings
+      - run: make check
+      - run: make eunit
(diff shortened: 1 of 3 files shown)


nalundgaard (commented):
Does `ubuntu-latest` include OTP/rebar2? It sounds like we might need [setup-erlang](https://github.com/marketplace/actions/setup-erlang) as well?

> (unless you want to initialize GitHub Actions for this repo, in `master`, after which I can rebase this)

Isn't this basically tantamount to initializing? The above question notwithstanding, I am OK if we just merge this and adapt it from there. Any objections from your side @motobob?

motobob:
no objections - let's just make it

motobob (approved):

nalundgaard:
OK, then just the question about whether the github actions will have OTP/rebar2/rebar3.

paulo-ferraz-oliveira:
> Does ubuntu-latest include OTP/rebar2? It sounds like we might need setup-erlang as well?

Not sure about `rebar`(2), but pretty confident as per the history in https://github.com/actions/starter-workflows/commits/abf7f258d1d84c79ad067c704e069c8cf7d8d2d0/ci/erlang.yml.

`setup-erlang` is **not** required, for sure.

If you merge to `master` (or initialize the CI somehow, so I can rebase this pull request on top of it and experiment there - without having broken Actions in `master`) I can try and fix it, so `rebar` (2) is OK.

nalundgaard:
Nah, let’s “initialize” with this and go from there. Thanks for the explanation.

paulo-ferraz-oliveira:
The fact that you mentioned `setup-erlang` got me thinking something was missing. It was. We're not running on top of a specific container 😄 Also, I will be removing the publish CI element, since that makes no sense (you might want to set it up later, using GitHub secrets).

PR #677, opened 2020-12-09
Return prepared erlcloud_ddb2 packet

### LINKED ISSUE #672: Export request creation functions on erlcloud_ddb2
Hi,

I'm developing a library on top of erlcloud that need packet creation functions inside **erlcloud_ddb2** to be exported.
As a matter of example, I need a function that processes the input of **erlcloud_ddb2:get_item** instead of sending the packet directly to DynamoDB to return to the caller function.

Is this possible?

Thank you for your time!

DarkGV:
This is my solution to issue #672.

I added a **DdbOpt** entry named `no_request` that defines whether the request should be or should not be sent.

Closes #672

### DIFF
--- a/include/erlcloud_ddb2.hrl
+++ b/include/erlcloud_ddb2.hrl
@@ -14,6 +14,13 @@
          response_body :: undefined | binary()
         }).
 
+-record(ddb2_request,
+       {headers :: headers(),
+        body :: jsx:json_text(),
+        json :: jsx:json_term()
+       }).
+
+-type headers() :: [{string(), string()}].
 -type date_time() :: number().
 -type global_table_status() :: creating | active | deleting | updating.
 -type replica_status() :: creating | creation_failed | updating | deleting | active.
@@ -447,9 +454,3 @@
        {table_description :: undefined | #ddb2_table_description{}
        }).
 -endif.
-
--record(ddb_request,
-       {headers :: erlcloud_ddb_impl:headers(),
-        body :: jsx:json_text(),
-        json :: jsx:json_term()
-       }).
--- a/include/erlcloud_ddb2.hrl
+++ b/include/erlcloud_ddb2.hrl
@@ -447,3 +447,9 @@
        {table_description :: undefined | #ddb2_table_description{}
        }).
 -endif.
+
+-record(ddb_request,
+       {headers :: erlcloud_ddb_impl:headers(),
+        body :: jsx:json_text(),
+        json :: jsx:json_term()
+       }).
--- a/src/erlcloud_ddb1.erl
+++ b/src/erlcloud_ddb1.erl
@@ -93,7 +93,7 @@ batch_get_item(RequestItems) ->
 -spec batch_get_item([batch_get_item_request_item()], aws_config()) -> json_return().
 batch_get_item(RequestItems, Config) ->
     Json = [{<<"RequestItems">>, [batch_get_item_request_item_json(R) || R <- RequestItems]}],
-    erlcloud_ddb_impl:request(Config, false, "DynamoDB_20111205.BatchGetItem", Json).
+    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.BatchGetItem", Json).
 
 -type batch_write_item_put() :: {put, item()}.
 -type batch_write_item_delete() :: {delete, key()}.
@@ -117,7 +117,7 @@ batch_write_item(RequestItems) ->
 -spec batch_write_item([batch_write_item_request_item()], aws_config()) -> json_return().
 batch_write_item(RequestItems, Config) ->
     Json = [{<<"RequestItems">>, [batch_write_item_request_item_json(R) || R <- RequestItems]}],
-    erlcloud_ddb_impl:request(Config, false, "DynamoDB_20111205.BatchWriteItem", Json).
+    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.BatchWriteItem", Json).
 
 
 -spec key_schema_value_json(key_schema_value()) -> jsx:json_term().
@@ -141,7 +141,7 @@ create_table(Table, KeySchema, ReadUnits, WriteUnits, Config) ->
             key_schema_json(KeySchema),
             {<<"ProvisionedThroughput">>, [{<<"ReadCapacityUnits">>, ReadUnits},
                                            {<<"WriteCapacityUnits">>, WriteUnits}]}],
-    erlcloud_ddb_impl:request(Config, false, "DynamoDB_20111205.CreateTable", Json).
+    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.CreateTable", Json).
 
     
 -spec delete_item(table_name(), key()) -> json_return().
@@ -157,7 +157,7 @@ delete_item(Table, Key, Opts, Config) ->
     Json = [{<<"TableName">>, Table},
             key_json(Key)] 
         ++ Opts,
-    erlcloud_ddb_impl:request(Config, false, "DynamoDB_20111205.DeleteItem", Json).
+    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.DeleteItem", Json).
 
     
 -spec delete_table(table_name()) -> json_return().
@@ -167,7 +167,7 @@ delete_table(Table) ->
 -spec delete_table(table_name(), aws_config()) -> json_return().
 delete_table(Table, Config) ->
     Json = [{<<"TableName">>, Table}],
-    erlcloud_ddb_impl:request(Config, false, "DynamoDB_20111205.DeleteTable", Json).
+    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.DeleteTable", Json).
 
     
 -spec describe_table(table_name()) -> json_return().
@@ -177,7 +177,7 @@ describe_table(Table) ->
 -spec describe_table(table_name(), aws_config()) -> json_return().
 describe_table(Table, Config) ->
     Json = [{<<"TableName">>, Table}],
-    erlcloud_ddb_impl:request(Config, false, "DynamoDB_20111205.DescribeTable", Json).
+    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.DescribeTable", Json).
 
 
 -spec get_item(table_name(), key()) -> json_return().
@@ -193,7 +193,7 @@ get_item(Table, Key, Opts, Config) ->
     Json = [{<<"TableName">>, Table},
             key_json(Key)] 
         ++ Opts,
-    erlcloud_ddb_impl:request(Config, false, "DynamoDB_20111205.GetItem", Json).
+    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.GetItem", Json).
 
 
 -spec list_tables() -> json_return().
@@ -206,7 +206,7 @@ list_tables(Opts) ->
 
 -spec list_tables(opts(), aws_config()) -> json_return().
 list_tables(Opts, Config) ->
-    erlcloud_ddb_impl:request(Config, false, "DynamoDB_20111205.ListTables", Opts).
+    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.ListTables", Opts).
 
     
 -spec put_item(table_name(), item()) -> json_return().
@@ -222,7 +222,7 @@ put_item(Table, Item, Opts, Config) ->
     Json = [{<<"TableName">>, Table},
             item_json(Item)] 
         ++ Opts,
-    erlcloud_ddb_impl:request(Config, false, "DynamoDB_20111205.PutItem", Json).
+    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.PutItem", Json).
 
 
 -spec q(table_name(), hash_key()) -> json_return().
@@ -238,7 +238,7 @@ q(Table, HashKey, Opts, Config) ->
     Json = [{<<"TableName">>, Table},
             hash_key_json(HashKey)] 
         ++ Opts,
-    erlcloud_ddb_impl:request(Config, false, "DynamoDB_20111205.Query", Json).
+    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.Query", Json).
 
 
 -spec scan(table_name()) -> json_return().
@@ -253,7 +253,7 @@ scan(Table, Opts) ->
 scan(Table, Opts, Config) ->
     Json = [{<<"TableName">>, Table}]
         ++ Opts,
-    erlcloud_ddb_impl:request(Config, false, "DynamoDB_20111205.Scan", Json).
+    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.Scan", Json).
 
 
 -spec update_item(table_name(), key(), updates()) -> json_return().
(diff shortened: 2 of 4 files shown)


src/erlcloud_ddb1.erl
```diff
@@ -93,7 +93,7 @@ batch_get_item(RequestItems) ->
 -spec batch_get_item([batch_get_item_request_item()], aws_config()) -> json_return().
 batch_get_item(RequestItems, Config) ->
     Json = [{<<"RequestItems">>, [batch_get_item_request_item_json(R) || R <- RequestItems]}],
-    erlcloud_ddb_impl:request(Config, "DynamoDB_20111205.BatchGetItem", Json).
+    erlcloud_ddb_impl:request(Config, false, "DynamoDB_20111205.BatchGetItem", Json).
```
paulo-ferraz-oliveira, line 96: Are all of these `false` actually required? Maybe I'm missing the _bigger picture_ here, but why not make `erlcloud_ddb_impl:request/3` use `false` by default, internally, while calling newer `erlcloud_ddb_impl:request/4` (while exposing both functions)?

src/erlcloud_ddb_impl.erl
```diff
@@ -68,23 +68,24 @@
         ]).
 
 %% Internal impl api
--export([request/3]).
+-export([request/4]).
 
 -export_type([json_return/0, attempt/0, retry_fun/0]).
 
--type json_return() :: ok | {ok, jsx:json_term()} | {error, term()}.
+-type json_return() :: ok | {ok, jsx:json_term()} | {error, term()} | list().
```
paulo-ferraz-oliveira, line 75: `list()` seems rather vague, right? Why not a tuple with an indication of what the actual return value means?

src/erlcloud_ddb2.erl
```diff
@@ -3434,7 +3496,7 @@ tag_resource(ResourceArn, Tags) ->
                   -> tag_resource_return().
 tag_resource(ResourceArn, Tags, Config) ->
     erlcloud_ddb_impl:request(
-      Config,
+      Config, false,
```
paulo-ferraz-oliveira, line 3499: This should probably call the `/3` (`false` by default) version of the implementation, if you decide to go that route.

src/erlcloud_ddb2.erl
```diff
@@ -3718,7 +3784,7 @@ untag_resource(ResourceArn, TagKeys) ->
                   -> untag_resource_return().
 untag_resource(ResourceArn, TagKeys, Config) ->
     erlcloud_ddb_impl:request(
-      Config,
+      Config, false,
```
paulo-ferraz-oliveira, line 3787: This should probably call the /3 (false by default) version of the implementation, if you decide to go that route.

src/erlcloud_ddb2.erl
```diff
@@ -1574,8 +1584,10 @@ batch_get_item(RequestItems, Opts) ->
                             batch_get_item_return().
 batch_get_item(RequestItems, Opts, Config) ->
     {AwsOpts, DdbOpts} = opts(batch_get_item_opts(), Opts),
+    NoRequest = proplists:get_value(no_request, DdbOpts, false),
```
paulo-ferraz-oliveira, line 1587: This seems a bit repetitive, but I can't figure out how to have a better pattern, since `Opts` is never passed to `request` function. This is a general comment, not a request or suggestion for change.

src/erlcloud_ddb_impl.erl
nalundgaard: why wouldn't this return `{ok, {RequestHeaders, Body}}` anyway?

nalundgaard: maybe `{ok, {request, Headers, Body}}`, for even more clarity? This seems like an OK approach, but the output format is extremely ambiguous and out of step with the style here.

src/erlcloud_ddb2.erl
nalundgaard: TBH I would suggest passing `DdbOpts` to `erlcloud_ddb_impl:request/4` as a 4th argument (and keep `erlcloud_ddb_impl:request/3` passing to `erlcloud_ddb_impl:request/4` with `DdbOpts = []`), then read the `no_request` option out there in exactly one place. That would make this substantially more concise and also flexible to do similar things in the future without changing the arity of every function call again.

src/erlcloud_ddb1.erl
DarkGV: This false was due to changes in erlcloud_ddb_impl:request function. I added a new request/4 and that request/3 will just make a call to this new function as suggested by @nalundgaard.

src/erlcloud_ddb_impl.erl
DarkGV: Yes, it makes it much more clear and goes within erlcloud style. I changed that for the sake of clarity 😄

src/erlcloud_ddb2.erl
DarkGV: DdbOpts is now checked inside erlcloud_ddb_impl:request/4

paulo-ferraz-oliveira: > TBH I would suggest passing

I was going to... but I didn't want to go that far 😄

paulo-ferraz-oliveira (approved):
Maintenance-wise, this new approach is much saner.

src/erlcloud_ddb2.erl
```diff
@@ -2977,7 +3007,7 @@ list_tags_of_resource(ResourceArn, Opts, Config) ->
     Return = erlcloud_ddb_impl:request(
                Config,
                "DynamoDB_20120810.ListTagsOfResource",
-               [{<<"ResourceArn">>, ResourceArn} | AwsOpts]),
+               [{<<"ResourceArn">>, ResourceArn} | AwsOpts], DdbOpts),
```
nalundgaard, line 3010: Can you put this one on the next line like the other calls?

```suggestion
               [{<<"ResourceArn">>, ResourceArn} | AwsOpts],
               DdbOpts),
```

src/erlcloud_ddb2.erl
```diff
@@ -3241,7 +3274,8 @@ restore_table_from_backup(BackupArn, TargetTableName, Opts, Config)
      Config,
      "DynamoDB_20120810.RestoreTableFromBackup",
      [{<<"BackupArn">>, BackupArn},
-      {<<"TargetTableName">>, TargetTableName}]),
+      {<<"TargetTableName">>, TargetTableName}],
+      DdbOpts),
```
nalundgaard, line 3278: Looks like there is a small alignment issue here, the `DdbOpts` should align to the `[` on L3276.

```suggestion
     DdbOpts),
```

src/erlcloud_ddb2.erl
```diff
@@ -3766,7 +3804,8 @@ update_continuous_backups(TableName, PointInTimeRecoveryEnabled, Opts, Config)
      Config,
      "DynamoDB_20120810.UpdateContinuousBackups",
      [{<<"TableName">>, TableName},
-      {<<"PointInTimeRecoverySpecification">>, dynamize_point_in_time_recovery_enabled(PointInTimeRecoveryEnabled)}]),
+      {<<"PointInTimeRecoverySpecification">>, dynamize_point_in_time_recovery_enabled(PointInTimeRecoveryEnabled)}],
+      DdbOpts),
```
nalundgaard, line 3808: Another small alignment problem here.
```suggestion
     DdbOpts),
```

src/erlcloud_ddb2.erl
```diff
@@ -3952,7 +3992,8 @@ update_global_table(GlobalTableName, ReplicaUpdates, Opts, Config) ->
                Config,
                "DynamoDB_20120810.UpdateGlobalTable",
                [{<<"GlobalTableName">>, GlobalTableName},
-                {<<"ReplicaUpdates">>, dynamize_maybe_list(fun dynamize_replica_update/1, ReplicaUpdates)}]),
+                {<<"ReplicaUpdates">>, dynamize_maybe_list(fun dynamize_replica_update/1, ReplicaUpdates)}],
+                DdbOpts),
```
nalundgaard, line 3996:
```suggestion
               DdbOpts),
```

src/erlcloud_ddb2.erl
```diff
@@ -4099,7 +4140,7 @@ update_global_table_settings(GlobalTableName, Opts, Config) ->
     Return = erlcloud_ddb_impl:request(
         Config,
         "DynamoDB_20120810.UpdateGlobalTableSettings",
-        [{<<"GlobalTableName">>, GlobalTableName} | AwsOpts]),
+        [{<<"GlobalTableName">>, GlobalTableName} | AwsOpts], DdbOpts),
```
nalundgaard, line 4143:
```suggestion
        [{<<"GlobalTableName">>, GlobalTableName} | AwsOpts],
        DdbOpts),
```

src/erlcloud_ddb2.erl
```diff
@@ -4414,7 +4456,7 @@ update_table_replica_auto_scaling(Table, Opts, Config) when is_list(Opts) ->
     Return = erlcloud_ddb_impl:request(
         UpdatedConfig,
         "DynamoDB_20120810.UpdateTableReplicaAutoScaling",
-        [{<<"TableName">>, Table} | AwsOpts]),
+        [{<<"TableName">>, Table} | AwsOpts], DdbOpts),
```
nalundgaard, line 4459:
```suggestion
        [{<<"TableName">>, Table} | AwsOpts],
        DdbOpts),
```

src/erlcloud_ddb_impl.erl
```diff
@@ -99,6 +107,12 @@ request(Config0, Operation, Json) ->
 %% This algorithm is similar, except that it waits a random interval up to 2^(Attempt-2)*100ms. The average
 %% wait time should be the same as boto.
 
+-spec maybe_request_and_retry(aws_config(), headers(), jsx:json_text(), {attempt, non_neg_integer()}, boolean()) -> json_return().
+maybe_request_and_retry(Config, Headers, Body, Attempt, false) ->
+    request_and_retry(Config, Headers, Body, Attempt);
+maybe_request_and_retry(_Config, Headers, Body, _Attempt, true) ->
+    {ok, {request, Headers, Body}}.
```
nalundgaard, line 114: Question here—are you sure you want encoded JSON, or would it be better to return the `jsx:json_term()` here? Not a suggested change, just wondering if it would be more useful to get the "raw" structure here.

src/erlcloud_ddb2.erl
```diff
@@ -1140,6 +1148,8 @@ out({error, Reason}, _, _) ->
     {error, Reason};
 out(ok, _, _) ->
     {error, unexpected_empty_response};
+out({ok, {request, _Headers, _Body}} = Request, _Undynamize, _Opts) ->
```
nalundgaard, line 1151: It looks like the spec for `out/3` is broken by this, because this form is not documented.

I think you also need to cover this one in `ddb_return/2` type spec.

nalundgaard (changes requested):
Looking much better, great work! Just a couple of minor concerns left.

One other thing. I don't know how comfortable you are with git rebasing, but if you can rebase this on `master`, I believe the PR will get the benefit of @paulo-ferraz-oliveira's recent #676 and #678. That would run your code through the test matrix, which would be handy.

src/erlcloud_ddb2.erl
paulo-ferraz-oliveira: After the rebase we should see this in the Action results 😄

src/erlcloud_ddb_impl.erl
DarkGV: Yes, it makes more sense to return the raw structure rather than the encoded body, also I don't think the headers are important to me.

For this situation I could create the headers and encode the body inside `maybe_request_and_retry` and return the raw structure in case `no_request` is set to true or, the easier way, I let it be the way it is and return `{ok, {request, {encoded, Headers, Body}, {raw, Json}}}`.

What do you think? For me, it is indifferent since I will only use the `{raw, Json}`.

paulo-ferraz-oliveira: Even if they're not important "to you" (and since this is a public API), do you imagine they could be useful for somebody else? If so, I'd ease the consumption for the "others" and keep as much information available, as possible, as long as it makes sense. @nalundgaard?

nalundgaard: One thing that might be handy here (and consistent with the erlcloud_ddb design) is to use a record, e.g.

```erlang
-record(ddb_request, {
    headers = [] :: headers(),
    body = <<>> :: iodata(),
    json :: jsx:json_term() | undefined
}).
```

If you write this into `erlcloud_ddb2.hrl`, you should be able to use it all over, and let clients use it. This will let us compatibly extend what is returned, if we ever care, and also make it succinct for clients to match out the parts of the record they care about. Another option, which would be what I'd do for a "new" API but not for _this one_, is use a map:

```erlang
-type ddb_request() :: #{headers := headers(), body := iodata(), json := jsx:json_term()}.
```

I prefer maps in general, but records is very much the style for this API, and I have to say that records are a lot less ambiguous 'out of the box' (because they have record tags). You can solve that by putting a property into the map (like `#{type => ddb_request}`), but anyway—I'd use a record here.

DarkGV: This now returns a record containing headers, body, and JSON.
From my point of view, this is a better approach since it doesn't break the coding style of this library and clients can take advantage of it.

include/erlcloud_ddb2.hrl
```diff
@@ -447,3 +447,9 @@
        {table_description :: undefined | #ddb2_table_description{}
        }).
 -endif.
+
+-record(ddb_request,
+       {headers :: erlcloud_ddb_impl:headers(),
```
paulo-ferraz-oliveira, line 452: Should `headers` be exported from `_impl`, rather than from `_ddb2`?

DarkGV: I checked on whether there was any type called `headers` on `_ddb2` however, I only found the type specification on `_impl` thus exported from there.

nalundgaard: I think I would actually delete that type from the `_impl` module and put it at the top of `erlcloud_ddb2.hrl` with the other types there (`date_time()`, etc.).

include/erlcloud_ddb2.hrl
```diff
@@ -447,3 +447,9 @@
        {table_description :: undefined | #ddb2_table_description{}
        }).
 -endif.
+
+-record(ddb_request,
```
nalundgaard, line 451: I think it would make the most sense to put this at the top of the module next to `ddb2_error`, and maybe call it `ddb2_request` for consistency.

src/erlcloud_ddb2.erl
```diff
@@ -3186,12 +3218,14 @@ q(Table, KeyConditionsOrExpression, Opts) ->
 -spec q(table_name(), conditions() | expression(), q_opts(), aws_config()) -> q_return().
 q(Table, KeyConditionsOrExpression, Opts, Config) ->
     {AwsOpts, DdbOpts} = opts(q_opts(), Opts),
+
```
nalundgaard, line 3221: can you remove this introduced whitespace?

nalundgaard (changes requested):
Looks like it's all good to go, just a couple of suggestions about rearranging and tweaking for consistency. Thanks!

nalundgaard (approved):

nalundgaard:
Thanks! This looks great, I hope it works for your use case. I appreciate your patience working through feedback.

PR #678, opened 2020-12-10
Fix GHA

paulo-ferraz-oliveira:
Opening this as Draft to check everything is as expected.


paulo-ferraz-oliveira:
I don't think the "execute rebar2 part is working as expected". Let me look at it closer.

paulo-ferraz-oliveira:
This seems to iron it out:
1. we build (CI) over multiple Erlang versions (a bug in the previous pull request prevented this from happening)
2. we run over `rebar` and `rebar3`

CI is only blocked because of Travis CI. Speaking of which, I'm going to remove it, [as @motobob implied should be done](https://github.com/erlcloud/erlcloud/issues/675#issuecomment-741712340).

nalundgaard (approved):
Man, this is wonderful! Thanks @paulo-ferraz-oliveira!

nalundgaard (approved):

nalundgaard:
> **Edit**: this means `deploy` from `.travis.yml` is no longer available.

I always do it by hand anyway, we can sort that out later. 🤷

paulo-ferraz-oliveira:
> I always do it by hand anyway, we can sort that out later.

Sure. I was asking around the other day (Erlang Slack) about how to automate this (publishing to Hex.pm), and was told it's not a very good practice to automate it, especially since you might not be pushing what you think you are (on the shell, at least, you have a notion of what it is).

PR #679, opened 2020-12-15
Add Ecs:ListTagsForResources support

kkuzmin:
https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_ListTagsForResource.html

I haven't found any specific reason for having `#ecs_tags{}` record with the only `tags` list field in it. The new `list_tags_for_resource()` method simply returns the list of `#ecs_tag{}` records 
```
{ok, [
    #ecs_tag{},
    #ecs_tag{},
    ...
  ]
}
```

instead of 
```
{ok, #ecs_tags{
  tags = [
    #ecs_tag{},
    #ecs_tag{},
    ...
  ]
}}
```

### DIFF
--- a/src/erlcloud_ecs.erl
+++ b/src/erlcloud_ecs.erl
@@ -1992,6 +1992,20 @@ list_services(Opts, Config) ->
         EcsOpts).
 
 %%%------------------------------------------------------------------------------
+%% ListTagsForResource
+%%%------------------------------------------------------------------------------
+-spec list_tags_for_resource(
+        Arn :: string_param()) -> ecs_return([#ecs_tag{}]).
+list_tags_for_resource(Arn) ->
+    list_tags_for_resource(Arn, [], default_config()).
+
+-spec list_tags_for_resource(
+        Arn :: string_param(),
+        Config :: aws_config()) -> ecs_return([#ecs_tag{}]).
+list_tags_for_resource(Arn, Config) when is_record(Config, aws_config) ->
+    list_tags_for_resource(Arn, [], default_config()).
+
+%%%------------------------------------------------------------------------------
 %% @doc 
 %% ECS API
 %% [https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ListTagsForResource.html]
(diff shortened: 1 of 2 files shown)


nalundgaard (changes requested):
LGTM, I think I agree on not having a wrapping record, can probably let clients get raw decoded JSON on the case that anything else is needed. Just the issue with the output specs, I guess should be `ecs_return([#ecs_tag{}]).`

kkuzmin:
@nalundgaard addressed. Thank you.

src/erlcloud_ecs.erl
```diff
@@ -1990,6 +1991,48 @@ list_services(Opts, Config) ->
     out(Return, fun(Json, UOpts) -> decode_record(list_services_record(), Json, UOpts) end,
         EcsOpts).
 
+%%%------------------------------------------------------------------------------
```
nalundgaard, line 1994: It looks like the style is to have a section heading like this at the "top"

```erlang
 %% ListTagsForResource
```

Then put this block with the full docs above the highest arity function, `list_tags_for_resource/3` here. I think that's the last change and G2G.

test/erlcloud_ecs_tests.erl
```diff
@@ -81,7 +81,9 @@ operation_test_() ->
       fun update_container_agent_input_tests/1,
       fun update_container_agent_output_tests/1,
       fun update_service_input_tests/1,
-      fun update_service_output_tests/1
+      fun update_service_output_tests/1,
+      fun list_tags_for_resource_input_tests/1,
```
nalundgaard, line 85: can you put these between `list_services_output_tests/1` and `list_task_definition_families_input_tests/1` as well?

nalundgaard (approved):

test/erlcloud_ecs_tests.erl
```diff
@@ -64,6 +64,8 @@ operation_test_() ->
       fun list_container_instances_output_tests/1,
       fun list_services_input_tests/1,
       fun list_services_output_tests/1,
+      fun list_tags_for_resource_input_tests/1,
+      fun list_tags_for_resource_output_tests/1
```
nalundgaard, line 68: looks like missing comma here causing CI to fail, actually.

PR #680, opened 2020-12-17
Rename ddb2 headers to ddb2_req_headers

cisom-al:
Fix type conflicts in downstream projects.

```
_build/default/lib/erlcloud/include/erlcloud_ddb2.hrl:23: type headers() already defined
make: *** [compile] Error 1
```


nalundgaard (approved):
Thanks for this! sorry for the super generic name.

PR #682, opened 2021-01-06
Add AWS Workspaces

kkuzmin:
Add a little bit of AWS Workspaces API:
[DescribeWorkspaces](https://docs.aws.amazon.com/workspaces/latest/api/API_DescribeWorkspaces.html)
[DescribeTags](https://docs.aws.amazon.com/workspaces/latest/api/API_DescribeTags.html)
[DescribeWorkspaceDirectories](https://docs.aws.amazon.com/workspaces/latest/api/API_DescribeWorkspaceDirectories.html)


src/erlcloud_aws.erl
```diff
@@ -781,7 +781,10 @@ service_config( <<"cur">>, Region, Config ) ->
     Config#aws_config{cur_host = Host};
 service_config( <<"application_autoscaling">>, Region, Config ) ->
     Host = service_host(<<"application_autoscaling">>, Region),
-    Config#aws_config{cur_host = Host}.
+    Config#aws_config{application_autoscaling_host = Host};
```
kkuzmin, line 784: Please note a small fix here.

motobob: wow. so it even even work?

kkuzmin: I don't think it ever worked. Probably hardly used by anyone.

nalundgaard: Yeah, we use `application_autoscaling` in our internal DynamoDB table management client, but we write it ourselves in one master configuration for the client's region (along with DynamoDB and DynamoDB Streams host). Nice catch.

src/erlcloud_workspaces.erl
```diff
@@ -0,0 +1,446 @@
...
+           end,
+    Headers = headers(Config, Operation, Payload),
+    Request = #aws_request{service = workspaces,
+                           uri = uri(Config),
+                           method = post,
+                           request_headers = Headers,
+                           request_body = Payload},
+    case erlcloud_aws:request_to_return(erlcloud_retry:request(Config, Request, fun workspaces_result_fun/1)) of
+        {ok, {_RespHeaders, <<>>}} -> {ok, []};
+        {ok, {_RespHeaders, RespBody}} -> {ok, jsx:decode(RespBody, [{return_maps, false}])};
```
nalundgaard, line 408: Any reason not to make this `true`? What's the point of adding new APIs with the legacy format?

src/erlcloud_workspaces.erl
```diff
@@ -0,0 +1,446 @@
...
+    lists:flatten([Scheme, Host, port_spec(Config)]).
+
+port_spec(#aws_config{workspaces_port=80}) ->
+    "";
+port_spec(#aws_config{workspaces_port=Port}) ->
+    [":", erlang:integer_to_list(Port)].
+
+
+to_binary(undefined) -> undefined;
+to_binary(true) -> true;
```
nalundgaard, line 439: this... isn't binary? seems kind of misleading.

include/erlcloud_workspaces.hrl
```diff
@@ -0,0 +1,64 @@
...
+-define(WORKSPACES_LIMIT, 25).
+
+-type workspaces_key_value_pair() :: {string() | binary(), string() | binary()}.
+
+%%%------------------------------------------------------------------------------
+%%
+%% Common data types
+%%
+%%%------------------------------------------------------------------------------
+-record(workspaces_pagination_options, {
```
nalundgaard, line 15: Is this used?

nalundgaard (changes requested):
I still need to finish going through the API docs to double-check things, but I noticed a couple of concerns on a first pass.

src/erlcloud_workspaces.erl
kkuzmin: I just followed the [best practices](https://github.com/erlcloud/erlcloud#best-practices). It doesn't tell anything about maps. 

> Public interfaces should normally expose proplists over records. All new modules are preferred to have both.

Also this implementation was borrowed from `erlcloud_ecs.erl`

include/erlcloud_workspaces.hrl
kkuzmin: removed, thank you

src/erlcloud_workspaces.erl
kkuzmin: Agree. This is done for json body encoding for AWS request. I'll rename it

src/erlcloud_workspaces.erl
nalundgaard: @motobob should we change this or just keep digging the 'proplist grave' for 'consistency'?

kkuzmin: @nalundgaard I've added `{out, map}` option. Please let me know what you think?

src/erlcloud_workspaces.erl
```diff
@@ -0,0 +1,588 @@
...
+-include("erlcloud_workspaces.hrl").
+
+%%% Library initialization.
+-export([configure/2, configure/3, configure/4, new/2, new/3, new/4]).
+
+-define(API_VERSION, "20150408").
+-define(OUTPUT_CHOICES, [json, record, map]).
+
+-export([
+     describe_tags/1, describe_tags/2,
```
nalundgaard, line 35: small 👾 issue here.

src/erlcloud_util.erl
```diff
@@ -246,3 +247,20 @@ http_uri_encode(URI) ->
...
+
+-spec proplists_to_map(proplists:proplist()) -> map().
+proplists_to_map(Proplist) ->
+    proplists_to_map(Proplist, #{}).
+
+-spec proplists_to_map(proplists:proplist(), map()) -> map().
+proplists_to_map([], Acc) ->
+    Acc;
+proplists_to_map([{Key, Val} | Tail], Acc) when is_list(Val) ->
+    proplists_to_map(Tail, Acc#{Key => proplists_to_map(Val)});
```
nalundgaard, line 259: This looks potentially a little problematic on the ambiguity of array vs. proplist object. I think maybe `proplists_to_map/1` might need some guard in place to prevent it from doing the wrong thing? This was covered in [jsn:from_proplist](https://github.com/nalundgaard/jsn/blob/4fe1217740e00008db94c7c6e6bfcae4537e2705/src/jsn.erl#L864-L892), although that implementation leaves out the special `[{}]` representation for empty objects that jsx uses (see [note](https://github.com/nalundgaard/jsn/tree/4fe1217740e00008db94c7c6e6bfcae4537e2705#proplist-format-concerns)). I think it might be appropriate here to map `[{}]` -> `#{}`, `[]` -> `[]`, and evaluate non-empty lists based on whether they match the pattern `[{_,_}|_]` (convert to proplist to map) or not (convert each array element to map if they are objects).

nalundgaard (changes requested):
Couple of issues, one of them I am a little concerned about. TBH I am fine if we leave out maps support if you and @motobob don't think it's worth it. Given that a robust conversion mechanism is a little more complex than what you have implemented, I'd prefer to either address the issues with conversion or just remove it.

src/erlcloud_util.erl
kkuzmin: Thanks @nalundgaard . I've added special function clauses. Please double check

src/erlcloud_workspaces.erl
nalundgaard: LGTM.

src/erlcloud_workspaces.erl
```diff
@@ -0,0 +1,588 @@
...
+    Opts1 = lists:ukeysort(1, proplists:unfold(Opts)),
+    lists:foldl(fun(Opt, A) -> opt_folder(Table, Opt, A) end, {[], []}, Opts1);
+opts(_, _) ->
+    error({erlcloud_workspaces, opts_not_list}).
+
+%%%------------------------------------------------------------------------------
+%%% Shared Decoders
+%%%------------------------------------------------------------------------------
+
+-type decode_opt() :: {typed, boolean()}.
```
nalundgaard, line 161: This doesn't seem to be used currently. Should we leave it in for future use, or extricate it now?

nalundgaard (approved):
LGTM, just a minor last question.

src/erlcloud_workspaces.erl
kkuzmin: I suggest we keep it. Removing it now and adding it later will be a fair chunk of refactoring.

PR #683, opened 2021-01-12
Allow for configurable metadata host and port

paulo-ferraz-oliveira:
I want `erlcloud_aws:auto_config()` to work locally without much hassle. I also want to not have to develop in an EC2 instance and then compare the result with the local env.

Let's say I use https://github.com/aws/amazon-ec2-metadata-mock to emulate EC2 metadata locally. With the content of this PR I can with a simple `application:set_env(erlcloud, metadata_host_port, "127.0.0.1:1338")`.

All previous behaviour remains the same. Would you like this option to be exposed somewhere (e.g. doc.)?

### DIFF
--- a/src/erlcloud_ec2_meta.erl
+++ b/src/erlcloud_ec2_meta.erl
@@ -25,12 +25,13 @@ get_instance_metadata(Config) ->
 %% @doc Retrieve the instance meta data for the instance this code is running on. Will fail if not an EC2 instance.
 %%
 %% This convenience function will retrieve the instance id from the AWS metadata available at 
-%% http://169.254.169.254/latest/meta-data/*
+%% http://<host:port>/latest/meta-data/*
 %% ItemPath allows fetching specific pieces of metadata.
+%% <host:port> defaults to 169.254.169.254
 %%
 %%
 get_instance_metadata(ItemPath, Config) ->
-    MetaDataPath = "http://169.254.169.254/latest/meta-data/" ++ ItemPath,
+    MetaDataPath = "http://" ++ ec2_meta_host_port() ++ "/latest/meta-data/" ++ ItemPath,
     erlcloud_aws:http_body(erlcloud_httpc:request(MetaDataPath, get, [], <<>>, erlcloud_aws:get_timeout(Config), Config)).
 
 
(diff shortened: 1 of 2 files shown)


src/erlcloud_ec2_meta.erl
```diff
@@ -25,12 +26,13 @@ get_instance_metadata(Config) ->
 %% @doc Retrieve the instance meta data for the instance this code is running on. Will fail if not an EC2 instance.
 %%
 %% This convenience function will retrieve the instance id from the AWS metadata available at 
-%% http://169.254.169.254/latest/meta-data/*
+%% http://<host:port>/latest/meta-data/*
```
nalundgaard, line 29: Should we document how to override this, at least the label of the application configuration parameter (`metadata_host_port`)? Also, would it be better for the label to be `ec2_meta_host_port`?

nalundgaard (commented):
Makes sense to me, but I am wondering if it wouldn't be better to just formalize `ec2_meta_host` (`169.254.169.254`)/`ec2_meta_port` (`80`) into the `#aws_config{}` record for consistency. @motobob do you have a thought here? I'm not really sure what would be more overhead, all things considered.

src/erlcloud_ec2_meta.erl
paulo-ferraz-oliveira: The first question is what I already asked in the description, basically: "Would you like this option to be exposed somewhere (e.g. doc.)?", so it might not be for me (not sure).

As for the label, I don't mind changing it; it's fairly innocuous. I let you decide.

src/erlcloud_ec2_meta.erl
```diff
@@ -66,6 +69,12 @@ get_instance_dynamic_data(Config) ->
...
-    DynamicDataPath = "http://169.254.169.254/latest/dynamic/" ++ ItemPath,
+    DynamicDataPath = "http://" ++ host_port() ++ "/latest/dynamic/" ++ ItemPath,
     erlcloud_aws:http_body(erlcloud_httpc:request(DynamicDataPath, get, [], <<>>, erlcloud_aws:get_timeout(Config), Config)).
 
+%%%------------------------------------------------------------------------------
+%%% Internal functions.
+%%%------------------------------------------------------------------------------
+
+host_port() ->
+    application:get_env(erlcloud, metadata_host_port, ?DEFAULT_METADATA_HOST_PORT).
```
motobob, line 80: I would prefer to have this in erlcloud.app.src as default. app.src is the place for all the app env settings. 
there you can also document it via comments . that way both comments will be addressed and good to go.

src/erlcloud_ec2_meta.erl
motobob: ec2_meta_host_port is preferable and as i mentioned in other comment - app.src is a good place for doc s for this.

paulo-ferraz-oliveira: 6a337be

src/erlcloud_ec2_meta.erl
paulo-ferraz-oliveira: 6a337be

paulo-ferraz-oliveira:
Just waiting for tests to finish...

motobob (approved):

PR #511, opened 2018-05-12
Add ability for Erlang applications to log events to AWS CloudWatch Logs

fogfish:
See #510

Content of PR
- [x] [DescribeLogStreams](https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_DescribeLogStreams.html)
- [x] [PutLogEvents](https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutLogEvents.html)

Note this PR is still **Work In Progress**, the purpose of such early PR is preview on the work.
 
One of proposal here is improvement of `erlcloud_cloudwatch_logs` code readability  due to excessive amount of nested `case ... of ... end`. I've been struggling with this in other Erlang projects as well, too much of boilerplate code. This problem can be solved by usage of [category pattern](https://github.com/fogfish/datum/blob/master/doc/category.md) and **either** category in particular. 

As an example

```erlang
case a() of
   {ok, A} ->
       case b(A) of
          {ok, B} ->
             c(B);
          {error, _} = Error ->
             Error
      end;
   {error, _} = Error ->
      Error
end
```

11 lines of nested cases are reduced to 1 line of pure functional composition throughout **either** category. 

```erlang
-compile({parse_transform, category}).

[either || a(), b(_), c(_)].
```

Please see more details from documentation of [category pattern](https://github.com/fogfish/datum/blob/master/doc/category.md).

Please compare the implementation of `cw_request`.  It became a sequence of pure functions. 
Let me know, your thought on this pattern.

PR #684, opened 2021-01-12
Expose erlcloud_cloudwatch_logs:start_query/stop_query/get_query_results

paulo-ferraz-oliveira:
~~I'm pushing this as a draft first (it's missing tests) to try and get early feedback.~~

~~Stuff I did:~~ This pull request:
- exposes functions `erlcloud_cloudwatch_logs:get_query_results/2,3`
- exposes functions `erlcloud_cloudwatch_logs:start_query/4,5,6`
- exposes functions `erlcloud_cloudwatch_logs:stop_query/1,2`
- exposes type `erlcloud_cloudwatch_logs:query_status/0`
- exposes type `erlcloud_cloudwatch_logs:query_results/0`
- has input/output as `string()` always, for consistency
- has output as `map()`-based (optional for `:get_query_results`), and transformed in a way that keeps a reference to the original (check doc.) but is more "native" to Erlang consumers (e.g. `atom()` and `map()`)

### DIFF
--- a/src/erlcloud_cloudwatch_logs.erl
+++ b/src/erlcloud_cloudwatch_logs.erl
@@ -123,6 +123,7 @@
     start_query/4,
     start_query/5,
     start_query/6,
+
     stop_query/1,
     stop_query/2,
 
@@ -511,8 +512,8 @@ get_query_results(QueryId, Config) ->
                       value := string() }].
 results_from_get_query_results(In) ->
     lists:map(fun (Result) ->
-                  #{ field => proplists:get_value(<<"field">>, Result),
-                     value => proplists:get_value(<<"value">>, Result) }
+                  #{ field => binary_to_list(proplists:get_value(<<"field">>, Result)),
+                     value => binary_to_list(proplists:get_value(<<"value">>, Result)) }
               end,
               In).
 
@@ -750,7 +751,7 @@ list_tags_log_group(LogGroup, Config) ->
            QueryString :: string(),
            StartTime :: non_neg_integer(),
            EndTime :: non_neg_integer(),
-           Result :: {ok, QueryId :: string()} | {error, erlcloud_aws:httpc_result_error()}.
+           Result :: {ok, #{ query_id => string() }} | {error, erlcloud_aws:httpc_result_error()}.
 start_query(LogGroupNames0, QueryString, StartTime, EndTime) ->
     start_query(LogGroupNames0, QueryString, StartTime, EndTime, _Limit = 1000).
 
@@ -760,7 +761,7 @@ start_query(LogGroupNames0, QueryString, StartTime, EndTime) ->
            StartTime :: non_neg_integer(),
            EndTime :: non_neg_integer(),
            Limit :: 1..10000,
-           Result :: {ok, QueryId :: string()} | {error, erlcloud_aws:httpc_result_error()}.
+           Result :: {ok, #{ query_id => string() }} | {error, erlcloud_aws:httpc_result_error()}.
 start_query(LogGroupNames0, QueryString, StartTime, EndTime, Limit) ->
     start_query(LogGroupNames0, QueryString, StartTime, EndTime, Limit, default_config()).
 
--- a/src/erlcloud_cloudwatch_logs.erl
+++ b/src/erlcloud_cloudwatch_logs.erl
@@ -820,7 +820,7 @@ stop_query(QueryId) ->
 -spec stop_query(QueryId, Config) -> Result
       when QueryId :: string(),
            Config :: aws_config(),
-           Result :: ok.
+           Result :: ok | {error, erlcloud_aws:httpc_result_error()}.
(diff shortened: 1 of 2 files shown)


paulo-ferraz-oliveira:
FWIW: `[{<<"rejectedLogEventsInfo">>,[{<<"tooOldLogEventEndIndex">>,1}]}]` is returned by AWS, for `:put_logs_events/5`, but hidden by the `erlcloud` API.

src/erlcloud_cloudwatch_logs.erl
```diff
@@ -437,6 +464,80 @@ log_stream_order_by(undefined) -> <<"LogStreamName">>;
...
+%% https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_GetQueryResults.html
+%%
+%% ===Example===
+%%
+%%   Returns the results from the specified query.
+%%
+%% `
+%%   application:ensure_all_started(erlcloud).
+%%   {ok, Config} = erlcloud_aws:auto_config().
+%%   {ok, Results} = erlcloud_cloudwatch_logs:get_query_results("12ab3456-12ab-123a-789e-1234567890ab").
```
nalundgaard, line 480: Would it be better to pass `Config` here explicitly?

nalundgaard (changes requested):
I like the map result format pretty well, but I'm not a huge fan of introducing it into an API that uses proplist-format JSX terms for all of the other API endpoints. Would it be better to pursue something more like #682, where there's an `{out, map}` option? I guess this would take away the nice atoms, and I guess on the input side #511 already kinda made this API somewhat inconsistent. I'm happy to approve if you prefer it as you have it, given the state of the module, just want to confirm.

paulo-ferraz-oliveira:
> I like the map result format pretty well, but I'm not a huge fan of introducing it into an API that uses proplist-format JSX terms for all of the other API endpoints. Would it be better to pursue something more like #682, where there's an `{out, map}` option? I guess this would take away the nice atoms, and I guess on the input side #511 already kinda made this API somewhat inconsistent. I'm happy to approve if you prefer it as you have it, given the state of the module, just want to confirm.

I don't mind changing this to support `{out, map}`, but would prefer to not have to change the non-map output, if that's OK.

src/erlcloud_cloudwatch_logs.erl
paulo-ferraz-oliveira: Sure: 87868d5.

paulo-ferraz-oliveira:
> I don't mind changing this to support `{out, map}`, but would prefer to not have to change the non-map output, if that's OK.

5e31836

paulo-ferraz-oliveira:
@nalundgaard, is there something you feel is blocking merging?

(if you do merge, I'd also like to request a release for inclusion as a tagged API, if that's OK)

src/erlcloud_cloudwatch_logs.erl
```diff
@@ -437,6 +464,85 @@ log_stream_order_by(undefined) -> <<"LogStreamName">>;
...
+           Options :: [{out, map}],
+           Results :: {ok, query_results()} | {error, erlcloud_aws:httpc_result_error()}.
+get_query_results(QueryId, Options) ->
+    get_query_results(QueryId, Options, default_config()).
+
+-spec get_query_results(QueryId, Options, Config) -> Results
+      when QueryId :: string(),
+           Options :: [{out, map}],
+           Config :: aws_config(),
+           Results :: {ok, query_results()} | {error, erlcloud_aws:httpc_result_error()}.
```
nalundgaard, line 497: Should this type spec be adjusted to account for the non-`query_results()` `{ok, ...}` form?

nalundgaard:
Sure, sorry I forgot to come back to this earlier in the week. I had one concern about the type spec after the out opt, otherwise LGTM, and will make a tag for you when it's merged for sure. Thanks!

src/erlcloud_cloudwatch_logs.erl
paulo-ferraz-oliveira: You are right. I was expecting the `dialyzer` to have a say, here, but it didn't :-(

nalundgaard (approved):

paulo-ferraz-oliveira:
Cool, thanks.

nalundgaard:
Thank you! I have made a release and published it to Hex. I'm kicking myself for not using squash and merge, but with time and therapy (and newer git history in the repo), I'm confident that the healing will take place.

Have a good one!

paulo-ferraz-oliveira:
😄 some people prefer a more detailed history (because there was a thought process behind the change), some people prefer a "cleaner" history, but one that shows only an increment. I have no preference for the lib.s I contribute to, and always expect the maintainers to let me know how they'd prefer it (e.g. I've had pull requests accepted into the same repo with request to merge + squash and without it, go figure 😄). I do prefer to have the merge commit, though, so I understand where the change comes from; some people don't...

PR #685, opened 2021-02-04
Properly interpret returned CloudWatch log events' list

paulo-ferraz-oliveira:
(in the map-based interface)

I mis-interpreted the AWS API, and this surfaced during tests. ~Opening as draft until my tests are finished and I'm sure I won't need to bother you again 😄 (at least until the next bug).~

**Note**: it didn't help that the tests were mocked, since they actually didn't represent the truth.


src/erlcloud_cloudwatch_logs.erl
```diff
@@ -516,14 +516,17 @@ get_query_results(QueryId, Options, Config) ->
...
+           Out :: [[#{ field := string(),
+                       value := string() }]].
 results_from_get_query_results([]) ->
     [];
-results_from_get_query_results([In]) ->
-    lists:map(fun (Result) ->
-                  #{ field => binary_to_list(proplists:get_value(<<"field">>, Result)),
-                     value => binary_to_list(proplists:get_value(<<"value">>, Result)) }
+results_from_get_query_results(In) ->
+    lists:map(fun (Results) ->
```
nalundgaard, line 524: Maybe it would be more clear with more descriptive names, like [ResultField](https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_ResultField.html), and list comprehensions:

```erlang
results_from_get_query_results(ResultRows) ->
    [ [#{ field => proplists:get_value(<<"field">>, ResultField),
          value => proplists:get_value(<<"value">>, ResultField) }
       || ResultField <- ResultRow ]
     || ResultRow <- ResultRows ].
```

The other thing that's kind of bothering me about this on closer look is, why `string()`? `binary()` seems better, and If I recall right, binary_to_list [doesn't handle unicode right](http://erlang.org/doc/apps/stdlib/unicode_usage.html#double-utf-8-encoding), which I am guessing these can be. I'd just avoid the conversion, personally, or use `unicode:characters_to_list()`

paulo-ferraz-oliveira: Sure, I can change that. But mind you the code was already there. I just fixed a bug.

paulo-ferraz-oliveira: > why string()? binary()

You mean "why is a conversion being done"? I _explained_ this in the previous pull request: "has input/output as string() always, for consistency", but I can change it (I guess I assumed it was OK since it wasn't commented on).

I mean, the API itself is already a bit confusing to me, so I'll just follow your lead.

(on a side note: if it were up to me, it'd be all `map`s and `binary`s - not `proplist`s, no `string`s, 😄)

paulo-ferraz-oliveira: 45e3764

nalundgaard: > I _explained_ this in the previous pull request: "has input/output as string() always, for consistency"

My bad, I missed this and/or didn't consider the consequences. Generally I'm more persuaded by the consistency argument, so probably best to keep it that way. As for the unicode concern, it looks like `binary_to_list/1` usage on potentially unicode binaries is a broader concern in the library, although there is some usage of `unicode:characters_to_list/1` present. So... Maybe never mind my naïve critiques! 😄

paulo-ferraz-oliveira: Well, I was **too fast** and acted upon your comments already. Want me to revert to the previous commit? I'm actually glad I changed to `binary()` since I have a lib. that wraps around that function and converts stuff to binary, so it's an extra step I can avoid 😄

nalundgaard: Since you introduced it just recently in #684 (and it was broken for lists of length >1), I'd say the binary switch is good, and looking over this module in more detail, it seems like most of the existing returns favor `binary()` rather than `string()` values (or just raw jsx objects which also have `binary()` values), although they support `string() | binary()` inputs in most cases. That makes sense to me, allowing strings in but favoring binaries out—it's just how it comes from jsx. So on consideration, I say keep it, especially if it's most convenient for you as well.

paulo-ferraz-oliveira: 👍 I'm OK with the current implementation.

paulo-ferraz-oliveira: Thanks.

nalundgaard (approved):

nalundgaard:
OK, [3.4.5](https://github.com/erlcloud/erlcloud/releases/tag/3.4.5) is tagged and [Hexed](https://hex.pm/packages/erlcloud/3.4.5)!

paulo-ferraz-oliveira:
Very cool. Thanks much.

PR #686, opened 2021-02-05
Improve SQS throttling detection

egobrain:
Some times SQS may return

```
{http_error,403,"Forbidden", <<"<?xml version=\"1.0\"?><ErrorResponse xmlns=\"http://queue.amazonaws.com/doc/2012-11-05/\"><Error><Type>Sender</Type><Code>RequestThrottled</Code><Message>Request throttled because message visibility configuration was updated too recently. Wait a short while and then retry the operation.</Message><Detail/></Error><RequestId>0e53ed64-5cf0-5d7f-8c22-b5f35143871d</RequestId></ErrorResponse>">>}}
```

which is also retriable error

### DIFF
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -251,7 +251,7 @@ do_aws_request_form_raw(Method, Scheme, Host, Port, Path, Form, Headers, Config,
                          response_status = Status} = Request) when
                 %% Retry for 400, Bad Request is needed due to Amazon
                 %% returns it in case of throttling
-                    Status == 400; Status == 429 ->
+                    Status == 400; Status == 403; Status == 429 ->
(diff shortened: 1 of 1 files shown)


src/erlcloud_aws.erl
```diff
@@ -1217,7 +1217,7 @@ is_throttling_error_response(RequestResponse) ->
          error_type = aws,
          response_body = RespBody} = RequestResponse,
 
-    case binary:match(RespBody, <<"Throttling">>) of
+    case binary:match(RespBody, <<"Throttl">>) of
```
egobrain, line 1220: This covers both `Throttled` and `Throttling` cases

motobob (approved):
interfaces are hard.

PR #688, opened 2021-02-15
Export DDB Streams record undynamizer

rmpalomino:
* Export DDB Streams record undynamizer to use with Kinesis Data Streaming for DynamoDB

After enabling  [Kinesis Data Streaming for DynamoDB](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/kds.html), DynamoDB will asynchronously export item changes to a Kinesis stream destination. The record format is a light wrapper around the existing DynamoDB Streams record, so exporting this undynamizer means that a client can do the following to make existing code compatible with Kinesis sourced records with minimal changes:

Get records from Kinesis, for each record:
1. Decode the `Data` field in the record using jsx
2. Call the exported `erlcloud_ddb_streams:undynamize_ddb_streams_record` function (1 or 2 arity, I use 2 arity with option {out, record} in the list of options)
3. Unwrap `Result` from the `{ok, Result}` tuple
4. 
Since DynamoDB sends records to the Kinesis streaming destination asynchronously, the records are not guaranteed to be in order and a record may appear more than once in the stream. The list of decoded records can be sorted using `lists:sort/2` to utilize a custom compare function and sort using the ApproximateCreationDateTime timestamp. The list should also be de-duplicated if your application does not handle repeated records.

### DIFF
--- a/src/erlcloud_ddb_streams.erl
+++ b/src/erlcloud_ddb_streams.erl
@@ -65,6 +65,9 @@
          list_streams/0, list_streams/1, list_streams/2
         ]).
 
+%%% Exported DynamoDB Streams Undynamizers
+-export([undynamize_ddb_streams_record/1, undynamize_ddb_streams_record/2]).
+
 -export_type(
    [aws_region/0,
     event_id/0,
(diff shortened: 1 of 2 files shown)


test/erlcloud_ddb_streams_tests.erl
```diff
@@ -587,3 +588,48 @@ list_streams_output_tests(_) ->
...
+                },
+                \"SequenceNumber\": \"300000000000000499659\",
+                \"SizeBytes\": 41,
+                \"StreamViewType\": \"KEYS_ONLY\"
+            },
+            \"eventID\": \"e2fd9c34eff2d779b297b26f5fef4206\",
+            \"eventName\": \"INSERT\",
+            \"eventSource\": \"aws:dynamodb\",
+            \"eventVersion\": \"1.0\"
+        }">>),
```
nalundgaard, line 612: I think you want to put `[{return_maps, false}]` explicitly here... for compatibility with newer jsx versions.

src/erlcloud_ddb_streams.erl
```diff
@@ -795,6 +798,29 @@ list_streams(Opts, Config) ->
...
+
+-spec undynamize_ddb_streams_record(json_term()) -> undynamize_ddb_streams_record_return().
+undynamize_ddb_streams_record(Return) ->
+    undynamize_ddb_streams_record(Return, []).
+
+%%------------------------------------------------------------------------------
+%% @doc Undynamizes a DynamoDB streams record.
+%% This function can be used to undynamize the jsx-decoded Data field of a
+%% record retrieved from Kinesis, after enabling Kinesis Data Streaming for
+%% a DynamoDB table.
```
nalundgaard, line 815: is there a good set of links to put in here for this use case?

nalundgaard (approved):
Other than unit test decoding issue, LGTM. Might also be worth sticking a little bit of KDS links in the function doc, but either way is good.

rmpalomino:
I added a link to the main KDS for DDB page. There really isn't anything useful for this specific use case, they don't go into detail about what the record in the Kinesis stream actually looks like.

nalundgaard (approved):

PR #693, opened 2021-03-05
Allow consumer-defined lhttpc_types -less lhttpc fork to compile without issues

### LINKED ISSUE #692: Issues with importing lhttpc_types.hrl
Hi. We have our own implementation of `lhttpc` which is a fork of the original. When we import `erlcloud` we import our own `lhttpc` (not the one `erlcloud` bundles). Now the thing is we recently removed `lhttpc_types.hrl` from our version (we put the types in `lhttpc` as we felt 1. stuff like `result()` was way to generic to not eventually cause a conflict, 2. we didn't want to keep an `.hrl` based part of the interface), for simplicity.

Would it be OK if I pull requested to get rid of the types, as imported by `erlcloud`? One strategy could be to update your `lhttpc` (I'd pull request there too) to import `lhttpc_types.hrl` to `lhttpc` and then export the types `erlcloud` requires from there. This way we don't break interface (though we kinda duplicate types).

Let me know. Thanks.

paulo-ferraz-oliveira:
Closes #692.

As discussed, in the linked issue, I present an alternative to importing `lhttpc_types.hrl` that allows consumers with a `lhttpc_types.hrl`less fork to continue compiling `erlcloud` without issues.

### DIFF
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -29,7 +29,6 @@
 
 -include("erlcloud.hrl").
 -include("erlcloud_aws.hrl").
--include_lib("lhttpc/include/lhttpc_types.hrl").
 
 -define(ERLCLOUD_RETRY_TIMEOUT, 10000).
 -define(GREGORIAN_EPOCH_OFFSET, 62167219200).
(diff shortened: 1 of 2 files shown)


nalundgaard (approved):

nalundgaard:
Looks good to me!

paulo-ferraz-oliveira:
Cool thanks.

PR #691, opened 2021-03-02
erlcloud_aws: support explicit querystring in writes

nalundgaard:
Fixes #690.

Refactors `erlcloud_aws:[do_]_aws_request_form_raw` function to take a QueryString distinctly from the Form and _always_ use that as the query string fragment of the URI, never in form body. This allows callers to explicitly force values to go into the query string on POST/PUT method calls where the Form gets stored in the body.

Also fixes the logic of `erlcloud_lambda:lambda_request_no_update` to explicitly encode the Body (Form) and QParams to be fully separate, so that QParams, which for lambda always must go in the query string fragment of the URI, is always explicitly passed there. This makes it so that the `erlcloud_lambda:invoke` call, which is a POST call, will pass its `Qualifier` parameter as a query string.

PR #694, opened 2021-03-12
Delete function API

wvj384:
While working with lambda versions and aliases found that there is no delete function api that can be useful (for example to delete old lambda versions). Maybe it will make sense to add it?
cc @motobob

### DIFF
--- a/src/erlcloud_lambda.erl
+++ b/src/erlcloud_lambda.erl
@@ -241,7 +241,7 @@ delete_function(FunctionName, Config) ->
     Qualifier    :: undefined | binary(),
     Config       :: aws_config()) -> return_val().
 delete_function(FunctionName, Qualifier, Config) ->
-    Path = base_path() ++ "functions/" ++ binary_to_list(FunctionName),
+    Path = base_path() ++ "functions/" ++ url_parameter(FunctionName),
     QParams = filter_undef([{"Qualifier", Qualifier}]),
     lambda_request(Config, delete, Path, undefined, QParams).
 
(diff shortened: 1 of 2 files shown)


motobob (approved):
cc @nalundgaard

src/erlcloud_lambda.erl
```diff
@@ -215,6 +216,34 @@ delete_event_source_mapping(Uuid, Config) ->
...
+-spec delete_function(FunctionName :: binary(),
+    Config       :: aws_config()) -> return_val().
+delete_function(FunctionName, Config) ->
+    delete_function(FunctionName, undefined, Config).
+
+-spec delete_function(FunctionName :: binary(),
+    Qualifier    :: undefined | binary(),
+    Config       :: aws_config()) -> return_val().
+delete_function(FunctionName, Qualifier, Config) ->
+    Path = base_path() ++ "functions/" ++ binary_to_list(FunctionName),
```
nalundgaard, line 244: Please use [`url_parameter(FunctionName)`](https://github.com/erlcloud/erlcloud/pull/691/files#diff-cc55208044e278022117428643ef16dba42d20c462b93fe80c5447dab2d84309R110) here.

nalundgaard (changes requested):
This PR LGTM, but it is 6 commits behind master and needs to be rebased, please. Particularly https://github.com/erlcloud/erlcloud/pull/691, and just one change below according to that PR.

src/erlcloud_lambda.erl
wvj384: Rebased and fixed

nalundgaard (approved):

nalundgaard:
LGTM! Thanks!

PR #695, opened 2021-03-23
Fix application autoscaler result fun to retry on ThrottlingException

va17:
* Fix application autoscaler result fun to retry on ThrottlingException
* Don't retry on 409 since it is not on Application Auto Scaling list of [common errors](https://docs.aws.amazon.com/autoscaling/application/APIReference/CommonErrors.html)

Please review @nalundgaard @rmpalomino


nalundgaard (changes requested):
You should also retry on [`ConcurrentUpdateException`](https://docs.aws.amazon.com/autoscaling/application/APIReference/API_RegisterScalableTarget.html#API_RegisterScalableTarget_Errors) (400) and [`ServiceUnavailable `](https://docs.aws.amazon.com/autoscaling/application/APIReference/CommonErrors.html) (503), I think. Those appear to be the intention of the previous logic, anyway. Would be nice to make it work as intended.

va17:
It already retries on 503.

src/erlcloud_application_autoscaler.erl
```diff
@@ -685,11 +685,29 @@ register_scalable_target(Configuration, BodyConfiguration) ->
...
+            end
+    catch
+        error:badarg ->
+            Request#aws_request{should_retry = false}
+    end;
+aas_result_fun(#aws_request{response_type = error,
+                            error_type = aws,
+                            response_status = Status} = Request) when
+      Status >= 500 ->
+    %% Retry on 500s including 503, SlowDown, Reduce your request rate.
```
nalundgaard, line 710: This comment should say `InternalFailure` or `ServiceUnavailable`, `SlowDown` and  `Reduce your request rate` are not documented in the API.

nalundgaard:
> It already retries on 503.

Fair, just need to fix the bad comment referenced above in that case.

nalundgaard (approved):

PR #700, opened 2021-04-28
Prevent consumers from having to -include a whole .hrl for a single type

### LINKED ISSUE #698: erlcloud:aws_config{} should be exposed via non-erlcloud_aws.hrl
This could be made as an exported type in `erlcloud.erl` and then the type exported by `erlcloud.hrl` would point to it (as to not break interface). 

**Rationale**: it's odd having to import a full file to reference a type; also, it's a good practice to keep types exported from the modules they most likely refer to (in this case, either `erlcloud.erl` or `erlcloud_config.erl`).

Will pull request if I get the time.

paulo-ferraz-oliveira:
(others may follow suite, in the future)

Closes #698.

`erlcloud` is a nice lib. to build more complex lib.s/app.s around, but sometimes its scope just gets in their way, especially when structures are passed around without the consumer having to know the internals.

This presents a (redundant) albeit useful type for referring to ~opaque~ type `aws_config()`.

### DIFF
--- a/src/erlcloud.erl
+++ b/src/erlcloud.erl
@@ -5,6 +5,8 @@
 -include("erlcloud_aws.hrl").
 -define(APP, erlcloud).
 
+-export_type([aws_config/0]).
+
 start() ->
     application:load(?APP),
     {ok, Apps} = application:get_key(?APP, applications),


nalundgaard:
why `t()` and not just `aws_config()`?

paulo-ferraz-oliveira:
> why `t()` and not just `aws_config()`?

I didn't want to repeat myself 😄 but I can do that: `erlcloud_config:aws_config()`. I've seen `t()` so many times it's almost become like a convention to me: "a lib.'s got a config. module that exposes a type? call it `t()` 😄".

Will update: naming, for me, is definitely not the most important part here.

paulo-ferraz-oliveira:
> Edit: hope it doesn't bring a conflict.

Argh. It can't be done. Any better idea for a name? How about `config`? It'll still be `erlcloud_config:config()`, though. Yeah, I'll push that and then I can easily revert if you don't like it.

src/erlcloud_config.erl
```diff
@@ -48,20 +51,20 @@
 %% Library initialization.
 %%------------------------------------------------------------------------------
 
--spec new(string(), string()) -> aws_config().
+-spec new(string(), string()) -> config().
```
paulo-ferraz-oliveira, line 54: Maybe this will even start shifting `:new` uses away from `aws_config()` which is the desired outcome.

nalundgaard:
What is the conflict with `aws_config()`? Naturally you can't redefine it in `erlcloud_config`, but you should be able to _just export it directly_, right?

On the other hand, would it be more succinct to export these types from the `erlcloud` module, i.e. `erlcloud:aws_config()`? That seems like what I'd prefer to do to minimize the module/type redundancy here.

paulo-ferraz-oliveira:
> Naturally you can't redefine it in erlcloud_config

This.

paulo-ferraz-oliveira:
> On the other hand, would it be more succinct to export these types from the erlcloud module, i.e. erlcloud:aws_config()?

Sure, works for me. Redoing.

paulo-ferraz-oliveira:
There we go. Nice and simple.

nalundgaard (approved):

nalundgaard:
@motobob any objection to this approach?

paulo-ferraz-oliveira:
@nalundgaard, any chance this gets merged soon? I'd like to use an Hex.pm -tagged version in our lib. soon 😄

nalundgaard:
Sorry, I'll just go ahead and do it! Thanks!

paulo-ferraz-oliveira:
That's awesome. Thanks much.

PR #707, opened 2021-08-31
Update erlcloud_aws.erl

pankajsoni19:
Fix secret manager endpoint


nalundgaard (approved):
+1, this matches the [reference from AWS](https://docs.aws.amazon.com/general/latest/gr/asm.html), thanks.

PR #711, opened 2021-10-31
VPC endpoint configuration: cache availability zone

nalundgaard:
Follow up to https://github.com/erlcloud/erlcloud/pull/638 (see [comment](https://github.com/erlcloud/erlcloud/pull/638#discussion_r407576219)).

We have a service that uses erlcloud to generate an AWS configuration for accessing the AWS backend based on HTTP requests. It probably should use a shared cache for those configuration structures, but did not need them originally. When we added a VPC Endpoint configuration, the service immediately became degraded, and testing confirmed that it was because we were calling the instance metadata endpoint in a DOS fashion, causing `erlcloud_ec2_meta:get_instance_metadata/2` to mostly take 5 seconds and return `{error, {socket_error, timeout}}`, breaking the VPC endpoint configuration in addition to making most requests to the service take 5+ seconds.

This seems like a hard edge for at-scale uses. It's very hard to understand this problem if you don't understand how this code works, because it doesn't break per se—it just becomes unbearably slow.

To solve the problem, I propose we cache `ok` results from the AZ lookup permanently in the application environment. It's not possible for this value to change during run-time, so this should be totally fine.

I also added some unit tests for this logic.

### DIFF
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -824,7 +824,7 @@ service_host( Service, Region ) when is_binary(Service) andalso is_binary(Region
 % magic can be done via EC2 DescribeVpcEndpoints/filter by VPC/filter by AZ.
 % however, permissions and describe* API throttling is not what we want to deal with here.
 get_host_vpc_endpoint(Service, Default) when is_binary(Service) ->
-    VPCEndpointsByService = application:get_env(erlcloud, services_vpc_endpoints, []),
+    VPCEndpointsByService = get_vpc_endpoints(),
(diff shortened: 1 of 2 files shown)


motobob (approved):
usually we cache out of the erlcloud and on top level but it's fine here for this case.

motobob:
>Line 942 Column 75: Guard test binary() =:= X__X::{'ok',<<_:248>>} can never succeed
Line 946 Column 1: Function test_get_host_vpc_endpoint_no_os_env/0 has no local return

tests

PR #712, opened 2021-11-10
Support EC2 calls against LocalStack

g-andrade:
(For anyone who doesn't know, [LocalStack](https://github.com/localstack/localstack) allows one to run a bunch of AWS services locally, which is ideal for testing code that will later be use the real stuff.)

In this PR I propose making EC2 protocol & port number configurable under `#aws_config{}`, as has been done for other services - this makes it possible to invoke LocalStack's EC2 service over HTTP and port 4566.


include/erlcloud_aws.hrl
```diff
@@ -20,7 +20,9 @@
 
 -record(aws_config, {
           as_host="autoscaling.amazonaws.com"::string(),
+          ec2_protocol="https"::string(),
           ec2_host="ec2.amazonaws.com"::string(),
+          ec2_port=433::non_neg_integer(),
```
g-andrade, line 25:
```suggestion
          ec2_port=443::non_neg_integer(),
```

rebar.lock
```diff
@@ -1,4 +1,4 @@
-{"1.1.0",
+{"1.2.0",
```
g-andrade, line 1: Orthogonal change which I performed with a "might as well" mindset.

nalundgaard, line 1: I don't think we want this yet?

g-andrade: It's backwards compatible, I figured it was a good a time as any to do it, but I can revert it if you prefer.

nalundgaard: I don't care too much, @motobob do you?

motobob: not in this repo. let it be as is 
in our private ones I want to prereserve 1.1.0.

include/erlcloud_aws.hrl
```diff
@@ -20,7 +20,9 @@
 
 -record(aws_config, {
           as_host="autoscaling.amazonaws.com"::string(),
+          ec2_protocol="https"::string(),
```
motobob, line 23: maybe it's time to generalise it for all? like we have VPCe

rebar.lock
g-andrade: eb5351c

include/erlcloud_aws.hrl
g-andrade: Like `services_protocol` / `services_port` settings at the level of application `env` config?

g-andrade: (Which would override any corresponding settings in `#aws_config{}` that were kept to their default values)

motobob: yes. smth like: 
if env_set use env:get(*_protocol)
else if aws_config{*_protocol}
else use default

granted, probably need to handle port variances for different service so env would looks like proplist/map and not just 2 variables.

g-andrade: That sounds reasonable!

Doing it for all services will bit a bit work though, and I won't find the time for it soon (the EC2 change is somewhat urgent), so I'll leave this PR open and use my fork for now.

motobob: I did not mean to ask to rework this PR. You send a follow up. We are not heavy users of local stack, only that. 

I’ll re-review this one later and probably merge and tag.

g-andrade: Ahh, alright. Thanks!

include/erlcloud_aws.hrl
```diff
@@ -20,7 +20,9 @@
...
           ec2_host="ec2.amazonaws.com"::string(),
+          ec2_port=443::non_neg_integer(),
```
motobob, line 25: to be backward compatible this should default to `undefined`
otherwise this makes ALL EC2 calls do `:443/` and not `/`
https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_aws.erl#L246-L247

granted it does not change much for your app but can affect other mecking lhttpC calls

https://docs.aws.amazon.com/general/latest/gr/rande.html#regional-endpoints

g-andrade: Handled in eff118f

motobob (approved):

motobob:
https://hex.pm/packages/erlcloud/3.5.8
sadly i forgot to squash the merge but ok.

PR #713, opened 2021-11-19
Fallback on unknown attribute names

neoascetic:
AWS rolled-out a new field and clients started to fail with `{function_clause ,[{erlcloud_sqs,decode_attribute_name,["SqsManagedSseEnabled"]...`

### DIFF
--- a/src/erlcloud_sqs.erl
+++ b/src/erlcloud_sqs.erl
@@ -46,7 +46,7 @@
                                     kms_master_key_id | kms_data_key_reuse_period_seconds |
                                     approximate_number_of_messages_not_visible | visibility_timeout |
                                     created_timestamp | last_modified_timestamp | policy |
-                                    queue_arn | deduplication_scope | fifo_throughput_limit).
+                                    queue_arn | deduplication_scope | fifo_throughput_limit | sqs_managed_sse_enabled).
 
 -type(batch_entry() ::   {string(), string()}
                        | {string(), string(), [message_attribute()]}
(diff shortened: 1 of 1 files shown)


src/erlcloud_sqs.erl
```diff
@@ -261,7 +262,9 @@ decode_attribute_name("KmsMasterKeyId") -> kms_master_key_id;
 decode_attribute_name("KmsDataKeyReusePeriodSeconds") -> kms_data_key_reuse_period_seconds;
 decode_attribute_name("FifoQueue") -> fifo_queue;
 decode_attribute_name("DeduplicationScope") -> deduplication_scope;
-decode_attribute_name("FifoThroughputLimit") -> fifo_throughput_limit.
+decode_attribute_name("FifoThroughputLimit") -> fifo_throughput_limit;
+decode_attribute_name("SqsManagedSseEnabled") -> sqs_managed_sse_enabled;
+decode_attribute_name(Name) -> Name.
```
motobob, line 267: not sure it's a good idea to fail later in request instead of here with func_clause...

nalundgaard: should we just leave these out?

nalundgaard: That is, filter out unknown attributes.

src/erlcloud_sqs.erl
```diff
@@ -46,7 +46,7 @@
                                     kms_master_key_id | kms_data_key_reuse_period_seconds |
                                     approximate_number_of_messages_not_visible | visibility_timeout |
                                     created_timestamp | last_modified_timestamp | policy |
-                                    queue_arn | deduplication_scope | fifo_throughput_limit).
+                                    queue_arn | deduplication_scope | fifo_throughput_limit | sqs_managed_sse_enabled).
```
nalundgaard, line 49: Should this be `| string()` too?

src/erlcloud_sqs.erl
neoascetic: Yeah, maybe this is a better way to do this. Want me to do this instead?

motobob: IMO - leave only SqsManagedSseEnabled and not decode_attribute_name(Name)

neoascetic: But that's what we had previously and now our services are having badtimes because of this new undocumented attribute :)

nalundgaard: I think we should leave it for now, handling the new attribute is enough to solve the problem. This is a fairly emergent issue so we should address it ASAP. Let's revisit how to handle unknown attributes in a follow-up. They definitely should be handled in some non-error/crash way end to end.

nalundgaard (approved):
Let's ship this now. Thanks @neoascetic, we'll discuss long-term handling here separately and as a follow-up.

src/erlcloud_sqs.erl
nalundgaard: I guess not, technically, since we don't support it on input?

PR #714, opened 2021-11-25
Fix SNS message attribute Name field

neoascetic:
As per https://docs.aws.amazon.com/sns/latest/api/API_Publish.html


motobob:
>MessageAttributes.entry.N.Name (key)
MessageAttributes.entry.N.Value (value)

motobob (approved):
(faceplam) but let's test in real app before merging

neoascetic:
That's why I've created this PR: it didn't work for me otherwise :)
Well, I was using localstack, but anyway.

nalundgaard (approved):
I'll ship this today along with the License fix for https://github.com/erlcloud/erlcloud/issues/715

PR #717, opened 2022-01-03
Handle invalid xml response in erlcloud_sns

cody-friedrichsen:
During an AWS incident of increased error rates on SNS, Topic Publish operations were disobeying API response format defined by [AWS Docs](https://docs.aws.amazon.com/sns/latest/api/API_Publish.html). Instead of well formed XML responses with required elements included, the response body for Publish calls was empty. In this period, errors were being generated from xmerl https://github.com/erlang/otp/blob/master/lib/xmerl/src/xmerl_scan.erl#L575 . Errors were of the form :
```
** exception exit: {fatal,{expected_element_start_tag,{file,file_name_unknown},
                                                              {line,1},
                                                              {col,1}}}
             in function  xmerl_scan:fatal/2 (xmerl_scan.erl, line 4124)
             in call from xmerl_scan:scan_document/2 (xmerl_scan.erl, line 572)
             in call from xmerl_scan:string/2 (xmerl_scan.erl, line 291)
```

This PR intends to handle an error of this form from xmerl and transform it into an sns error with form `{sns_error, invalid_xml_response_document}` . 

Additionally this PR introduces an `eunit.config` which redirects error reports from a terminal output into an `eunit.log` file to prevent error reports from cluttering the terminal at each test invocation.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -37,3 +37,5 @@
            ]}.
 
 {post_hooks, [{clean, "rm -f .dialyzer_plt"}]}.
+
+{pre_hooks, [{clean, "rm -rf erl_crash.dump *.log *.coverage.xml"}]}.
--- a/rebar.config
+++ b/rebar.config
@@ -38,4 +38,4 @@
 
 {post_hooks, [{clean, "rm -f .dialyzer_plt"}]}.
 
-{pre_hooks, [{clean, "rm -rf erl_crash.dump *.log *.coverage.xml"}]}.
+{pre_hooks, [{clean, "rm -rf erl_crash.dump *.log"}]}.
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -73,23 +73,23 @@
 
 aws_request_xml(Method, Host, Path, Params, #aws_config{} = Config) ->
     Body = aws_request(Method, Host, Path, Params, Config),
-    element(1, xmerl_scan:string(binary_to_list(Body))).
+    raw_xml_response(Body).
 aws_request_xml(Method, Host, Path, Params, AccessKeyID, SecretAccessKey) ->
     Body = aws_request(Method, Host, Path, Params, AccessKeyID, SecretAccessKey),
-    element(1, xmerl_scan:string(binary_to_list(Body))).
+    raw_xml_response(Body).
 aws_request_xml(Method, Protocol, Host, Port, Path, Params, #aws_config{} = Config) ->
     Body = aws_request(Method, Protocol, Host, Port, Path, Params, Config),
-    element(1, xmerl_scan:string(binary_to_list(Body))).
+    raw_xml_response(Body).
 aws_request_xml(Method, Protocol, Host, Port, Path, Params, AccessKeyID, SecretAccessKey) ->
     Body = aws_request(Method, Protocol, Host, Port, Path, Params, AccessKeyID, SecretAccessKey),
-    element(1, xmerl_scan:string(binary_to_list(Body))).
+    raw_xml_response(Body).
 
 aws_request_xml2(Method, Host, Path, Params, #aws_config{} = Config) ->
     aws_request_xml2(Method, undefined, Host, undefined, Path, Params, Config).
 aws_request_xml2(Method, Protocol, Host, Port, Path, Params, #aws_config{} = Config) ->
     case aws_request2(Method, Protocol, Host, Port, Path, Params, Config) of
         {ok, Body} ->
-            {ok, element(1, xmerl_scan:string(binary_to_list(Body)))};
+            format_xml_response(Body);
         {error, Reason} ->
             {error, Reason}
     end.
@@ -99,7 +99,7 @@ aws_request_xml4(Method, Host, Path, Params, Service, #aws_config{} = Config) ->
 aws_request_xml4(Method, Protocol, Host, Port, Path, Params, Service, #aws_config{} = Config) ->
     case aws_request4(Method, Protocol, Host, Port, Path, Params, Service, Config) of
         {ok, Body} ->
-            {ok, element(1, xmerl_scan:string(binary_to_list(Body)))};
+            format_xml_response(Body);
         {error, Reason} ->
             {error, Reason}
     end.
@@ -330,6 +330,20 @@ encode_params(Params, Headers) ->
     ?DEFAULT_CONTENT_TYPE -> {erlcloud_http:make_query_string(Params), LowerCaseHeaders};
     _ContentType -> {Params, LowerCaseHeaders}
   end.
+
+raw_xml_response(Body) ->
+    case format_xml_response(Body) of
+        {ok, XML} -> XML;
+        Error -> Error
+    end.
+
+format_xml_response(Body) ->
+    try element(1, xmerl_scan:string(binary_to_list(Body))) of
+        XML -> {ok, XML}
+    catch
+        exit:{fatal,{expected_element_start_tag, _, _, _}} ->
+            {aws_error, {invalid_xml_response_document, Body}}
+    end.
 
 %%%---------------------------------------------------------------------------
 -spec default_config() -> aws_config().
--- a/src/erlcloud_aws.erl
+++ b/src/erlcloud_aws.erl
@@ -89,7 +89,10 @@ aws_request_xml2(Method, Host, Path, Params, #aws_config{} = Config) ->
 aws_request_xml2(Method, Protocol, Host, Port, Path, Params, #aws_config{} = Config) ->
     case aws_request2(Method, Protocol, Host, Port, Path, Params, Config) of
         {ok, Body} ->
-            format_xml_response(Body);
+            case format_xml_response(Body) of
+                {ok, XML} -> {ok, XML};
+                Error -> {error, Error}
+            end;
         {error, Reason} ->
             {error, Reason}
     end.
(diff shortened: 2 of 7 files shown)


src/erlcloud_sns.erl
```diff
@@ -580,12 +580,15 @@ publish(Type, RecipientArn, Message, Subject, Attributes, Config) ->
             Subject -> [{"Subject", Subject}]
         end,
     AttributesParam = message_attributes(Attributes),
-    Doc =
-        sns_xml_request(
+    try sns_xml_request(
```
motobob, line 583: should this be applied to other xml based APIs?

motobob (approved):

test/erlcloud_sns_tests.erl
```diff
@@ -121,7 +122,11 @@ validate_params(Body, Expected) ->
 %% Validates the query body and responds with the provided response.
 -spec input_expect(string(), [expected_param()]) -> fun().
 input_expect(Response, Expected) ->
-    fun(_Url, post, _Headers, Body, _Timeout, _Config) ->
+    fun
+        (_Url, post, _Headers, "test message" = Body, _Timeout, _Config) ->
+            validate_params(Body, Expected),
+            {ok, {{200, "OK"}, [], <<"">>}};
```
nalundgaard, line 128:
```suggestion
            {ok, {{200, "OK"}, [], <<>>}};
```

src/erlcloud_sns.erl
```diff
@@ -580,12 +580,15 @@ publish(Type, RecipientArn, Message, Subject, Attributes, Config) ->
...
             Config, "Publish",
-            RecipientParam ++ MessageParams ++ SubjectParam ++ AttributesParam),
-    erlcloud_xml:get_text(
-        "PublishResult/MessageId", Doc).
+            RecipientParam ++ MessageParams ++ SubjectParam ++ AttributesParam) of
+        Doc ->
+            erlcloud_xml:get_text("PublishResult/MessageId", Doc)
+    catch
+        exit:{fatal,{expected_element_start_tag, _, _, _}}:_Stack ->
+            erlang:error({sns_error, invalid_xml_response_document})
```
nalundgaard, line 590: I would really like this to return the raw response, like `{aws_error, {invalid_xml_response_document, Response}}`. Is it possible to catch the above error farther up where we can collect this response?

I don't think anyone at the center of all this will care about us handling this exception and returning something generic that isn't garbage and clearly points to the real problem.

nalundgaard (changes requested):
If you add the change to put error logs in eunit.log, please make sure you add a cleanup step to the rebar.config to delete the log when calling `rebar3 clean`.

test/erlcloud_sns_tests.erl
cody-friedrichsen: Removed this clause as it was not being used

src/erlcloud_sns.erl
cody-friedrichsen: Updated to handle invalid xml in `erlcloud_aws` enabling a generic error return containing the invalid body

src/erlcloud_aws.erl
```diff
@@ -331,6 +331,20 @@ encode_params(Params, Headers) ->
...
+        {ok, XML} -> XML;
+        Error -> Error
+    end.
+
+format_xml_response(Body) ->
+    try element(1, xmerl_scan:string(binary_to_list(Body))) of
+        XML -> {ok, XML}
+    catch
+        exit:{fatal,{expected_element_start_tag, _, _, _}} ->
+            {aws_error, {invalid_xml_response_document, Body}}
```
nalundgaard, line 345: I don't think `{aws_error, _}` is a valid form here; it should be `{error, {aws_error, {invalid_xml_response_document, Body}}}` IMO. when raising an error, it should be raised as `{aws_error, Reason}`, but this isn't consistent with the other library forms and might mess some consumers up.

src/erlcloud_aws.erl
```diff
@@ -73,23 +73,23 @@
 
 aws_request_xml(Method, Host, Path, Params, #aws_config{} = Config) ->
     Body = aws_request(Method, Host, Path, Params, Config),
```
nalundgaard, line 75: It looks like `aws_request` [raises](https://github.com/erlcloud/erlcloud/blob/e5f03228189b7582d6e825c605089bd4ee652f3b/src/erlcloud_aws.erl#L116) `erlang:error({aws_error, Reason})`. I think `aws_request_xml` should remain consistent to that.

src/erlcloud_aws.erl
```diff
@@ -331,6 +331,20 @@ encode_params(Params, Headers) ->
     _ContentType -> {Params, LowerCaseHeaders}
   end.
 
+raw_xml_response(Body) ->
+    case format_xml_response(Body) of
+        {ok, XML} -> XML;
+        Error -> Error
```
nalundgaard, line 337: I think you  can just raise the exception here?

nalundgaard (changes requested):
Makes more sense, a few thoughts....

src/erlcloud_aws.erl
cody-friedrichsen: Error handling in `erlcloud_sns` matches `{error, Reason}` and raises that as `{sns_error, Reason}` . The end result of an invalid xml response document in `aws_request` then becomes ` {sns_error,{aws_error,{invalid_xml_response_document,<<>>}}}` . I believe this is the response that conforms with current error handling in `erlcloud_sns` as everything that fails out of that module raises `{sns_error, Reason}` .

src/erlcloud_aws.erl
cody-friedrichsen: Updated to raise an exception in `aws_request` path.  `aws_request2` and `aws_request4` will return error tuples to the calling modules.

rebar.config
```diff
@@ -37,3 +37,5 @@
            ]}.
 
 {post_hooks, [{clean, "rm -f .dialyzer_plt"}]}.
+
+{pre_hooks, [{clean, "rm -rf erl_crash.dump *.log *.coverage.xml"}]}.
```
nalundgaard, line 41: I don't think this repo outputs `*.coverage.xml`, doesn't look like it's set up for coverage reporting at all TBH.

src/erlcloud_aws.erl
```diff
@@ -331,6 +337,20 @@ encode_params(Params, Headers) ->
...
+        Error -> erlang:error(Error)
+    end.
+
+format_xml_response(Body) ->
+    try element(1, xmerl_scan:string(binary_to_list(Body))) of
+        XML -> {ok, XML}
+    catch
+        exit:{fatal,{expected_element_start_tag, _, _, _}} ->
+            {aws_error, {invalid_xml_response_document, Body}}
+    end.
```
nalundgaard, line 352: Thinking about this a little more, do we need to narrowly match errors? I am not sure about XML decoding, but I know JSON decoding can raise a variety of exceptions. I'd be fine to just blind-match them and treat them all as invalid XML document. Any reason not to?
```suggestion
format_xml_response(Body) ->
    XML = try 
        element(1, xmerl_scan:string(binary_to_list(Body)))
    catch
        _:_ ->
            {aws_error, {invalid_xml_response_document, Body}}
    end,
    {ok, XML}.
```

src/erlcloud_sns.erl
```diff
@@ -580,12 +580,13 @@ publish(Type, RecipientArn, Message, Subject, Attributes, Config) ->
...
     AttributesParam = message_attributes(Attributes),
-    Doc =
-        sns_xml_request(
+    case sns_xml_request(
             Config, "Publish",
-            RecipientParam ++ MessageParams ++ SubjectParam ++ AttributesParam),
-    erlcloud_xml:get_text(
-        "PublishResult/MessageId", Doc).
+            RecipientParam ++ MessageParams ++ SubjectParam ++ AttributesParam) of
+        {ok, Doc} ->
```
nalundgaard, line 586: This is no longer valid... `sns_xml_request` returns `Doc` or raises an error and every other use in the module appears to treat it this way. I think this `case` change can be undone completely.

test/erlcloud_sns_tests.erl
```diff
@@ -168,7 +169,12 @@ output_test(Fun, {Line, {Description, Response, Result}}) ->
       fun() ->
               meck:expect(erlcloud_httpc, request, output_expect(Response)),
               erlcloud_ec2:configure(string:copies("A", 20), string:copies("a", 40)),
-              Actual = Fun(),
+              Actual = try
+                  Fun()
+                catch
+                  _Class:Error:_Stack ->
```
nalundgaard, line 175: Technically no need to match `:_Stack` here, it's just wasted characters.
```suggestion
                  _Class:Error ->
```

nalundgaard (commented):
LGTM but a couple of minor things, and one bug.

src/erlcloud_aws.erl
cody-friedrichsen: I can think of a few bad reasons, but not any good ones.

rebar.config
cody-friedrichsen: Removed coverage pattern

src/erlcloud_sns.erl
cody-friedrichsen: removed case. Back to original code here

test/erlcloud_sns_tests.erl
cody-friedrichsen: Removed stacktrace matching

nalundgaard (approved):

src/erlcloud_aws.erl
```diff
@@ -331,6 +337,21 @@ encode_params(Params, Headers) ->
...
+
+format_xml_response(Body) ->
+    XML = try 
+        element(1, xmerl_scan:string(binary_to_list(Body)))
+    catch
+        _:_ ->
+            {aws_error, {invalid_xml_response_document, Body}}
+    end,
+    {ok, XML}.
```
nalundgaard, line 353: Sorry, I was thinking about this one raising an error, looks like I borked it.

```suggestion
format_xml_response(Body) ->
    try 
        {ok, element(1, xmerl_scan:string(binary_to_list(Body)))}
    catch
        _:_ ->
            {aws_error, {invalid_xml_response_document, Body}}
    end.
```

PR #718, opened 2022-01-05
erlcloud_sqs: get_queue_attributes improvement, test

nalundgaard:
Follow-up to https://github.com/erlcloud/erlcloud/pull/713.

When AWS added `SqsManagedSseEnabled`, we were lucky that it was a Boolean value, so it didn't break `erlcloud_sqs:decode_attribute_value/1`, which assumes that all attribute values that aren't specifically known are Boolean or integer values. This change anticipates a future when they add a string attribute value (like `QueueArn` or `KmsMasterKeyId `), so that we don't crash decoding THAT.

Finally, adding some `get_queue_attributes` testing to `erlcloud_sqs_tests` to cover this functionality.

Edit: also added these enhancements to get tests to pass:

> erlcloud_sqs: extend sqs_queue_attribute_names 
> * Reorder `encode_attribute_name` and `decode_attribute_name` to use the ordering from [GetQueueAttributes API Reference](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_GetQueueAttributes.html) `AttributeName.N` definition (alphabetical, with common params first, then SSE & FIFO specific attribute names in sections below).
> * Add `FifoQueue` and `ContentBasedDeduplication` to `encode_attribute_name` and `decode_attribute_name`, previously missed FIFO parameters that should be supported.
> * Refactor `sqs_queue_attribute_name` type to include all known types.

### DIFF
--- a/src/erlcloud_sqs.erl
+++ b/src/erlcloud_sqs.erl
@@ -275,7 +275,13 @@ decode_attribute_value("DeduplicationScope", Value) -> Value;
 decode_attribute_value("FifoThroughputLimit", Value) -> Value;
 decode_attribute_value(_, "true") -> true;
 decode_attribute_value(_, "false") -> false;
-decode_attribute_value(_, Value) -> list_to_integer(Value).
+decode_attribute_value(_, Value) ->
+    try
+        list_to_integer(Value)
+    catch
+        error:badarg ->
+            Value
+    end.
 
 
 -spec list_queues() -> [string()].
--- a/src/erlcloud_sqs.erl
+++ b/src/erlcloud_sqs.erl
@@ -42,11 +42,14 @@
                                   approximate_first_receive_timestamp |
                                   wait_time_seconds |
                                   receive_message_wait_time_seconds).
--type(sqs_queue_attribute_name() :: all | approximate_number_of_messages |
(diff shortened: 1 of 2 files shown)


motobob (approved):

cody-friedrichsen (approved):

nalundgaard:
Note that I added a new commit to cover a couple more issues:

> erlcloud_sqs: extend sqs_queue_attribute_names
> * Reorder `encode_attribute_name` and `decode_attribute_name` to use the ordering from [GetQueueAttributes API > Reference](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_GetQueueAttributes.html) `AttributeName.N` definition (alphabetical, with common params first, then SSE & FIFO specific attribute names in sections below).
> * Add `FifoQueue` and `ContentBasedDeduplication` to `encode_attribute_name` and `decode_attribute_name`, previously missed FIFO parameters that should be supported.
> * Refactor `sqs_queue_attribute_name` type to include all known types.

The `sqs_queue_attribute_name` type spec was what made `dialyzer` angry in OTP 24, because someone had missed `delay_seconds` and `receive_message_wait_time_seconds` in the spec, and I was passing those into my argument. Lovely. Hopefully all G2G now.

PR #723, opened 2022-02-15
Use new EC2 API version for regions discovery.

kkuzmin:
Add filters and `optInStatus` property to [`DescribeRegions`](
https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeRegions.html)

### DIFF
--- a/src/erlcloud_ec2.erl
+++ b/src/erlcloud_ec2.erl
@@ -21,7 +21,7 @@
          %% Availability Zones and Regions
          describe_availability_zones/0, describe_availability_zones/1,
          describe_availability_zones/2,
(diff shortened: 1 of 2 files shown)


src/erlcloud_ec2.erl
```diff
@@ -1689,16 +1689,23 @@ describe_regions(Config)
   when is_record(Config, aws_config) ->
     describe_regions([], Config);
 describe_regions(RegionNames) ->
-    describe_regions(RegionNames, default_config()).
+    describe_regions(RegionNames, none, default_config()).
```
motobob, line 1692: imo this change is unnecessary

motobob (approved):
minor change please

PR #725, opened 2022-04-15
use dot in S3 host name

key-master:
### Problem
I have issues accessing S3 with erlcloud locally

### Solution
according to doc https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html
`s3-<Region>.amazonaws.com` host naming schema is a legacy one and the current schema is `s3.<Region>.amazonaws.com` (dot symbol instead of dash)

also the hostname with dash does not work for me locally, but the dot works

@motobob please take a look


carlisom (approved):

nalundgaard (approved):
LGTM. Do we still need this one?

```erlang
service_host( <<"s3">>, <<"us-east-1">> ) -> "s3-external-1.amazonaws.com";
```

It looks like from [Amazon S3 endpoints](https://docs.aws.amazon.com/general/latest/gr/s3.html) that `s3.us-east-1.amazonaws.com` works here.

kkuzmin (approved):

PR #727, opened 2022-06-06
Support AWS cognito

Amershan:
This is. a partial support for aws cognito sdk

The code was written with support from [@arthurnovak](https://github.com/arthurnovak)


arthurnovak:
hey @motobob @nalundgaard, we've recently supported AWS Cognito interface in erlcloud. Can you have a look please

src/erlcloud_util.erl
```diff
@@ -195,9 +198,20 @@ next_token(Path, XML) ->
...
+    lists:filter(fun({_Name, Value}) -> Value =/= undefined end, List);
+filter_undef(Map) when is_map(Map) ->
+    maps:filter(fun(_Key, Value) -> Value =/= undefined end, Map).
+
+-spec filter_empty_map(maps:map()) -> maps:map().
+filter_empty_map(Map) ->
+    maps:filter(fun(_Key, Value) -> Value =/= #{} end, Map).
+
+-spec filter_empty_list(maps:map()) -> maps:map().
+filter_empty_list(Map) ->
```
motobob, line 213: need map guard

src/erlcloud_util.erl
```diff
@@ -195,9 +198,20 @@ next_token(Path, XML) ->
...
-    lists:filter(fun({_Name, Value}) -> Value =/= undefined end, List).
+-spec filter_undef(proplists:proplist() | maps:map()) ->
+    proplists:proplist() | maps:map().
+filter_undef(List) when is_list(List) ->
+    lists:filter(fun({_Name, Value}) -> Value =/= undefined end, List);
+filter_undef(Map) when is_map(Map) ->
+    maps:filter(fun(_Key, Value) -> Value =/= undefined end, Map).
+
+-spec filter_empty_map(maps:map()) -> maps:map().
+filter_empty_map(Map) ->
```
motobob, line 209: need map guard

src/erlcloud_cognito.erl
```diff
@@ -0,0 +1,906 @@
...
+                         AttributeMapping, IdpIdentifiers, Config) ->
+  Body = #{
+    <<"UserPoolId">>       => UserPoolId,
+    <<"ProviderName">>     => ProviderName,
+    <<"ProviderType">>     => ProviderType,
+    <<"ProviderDetails">>  => ProviderDetails,
+    <<"AttributeMapping">> => AttributeMapping,
+    <<"IdpIdentifiers">>   => IdpIdentifiers
+  },
+  request(Config, "CreateIdentityProvider", erlcloud_util:filter_empty_map(Body)).
```
motobob, line 776: such use of erlcloud_util:filter_empty_map() across the file makes the API ignore mandatory specs for AWS API and effectively renders everything as optional leading to unnecessary dev mistakes for bad requests. 

https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_CreateIdentityProvider.html

src/erlcloud_cognito.erl
```diff
@@ -0,0 +1,906 @@
...
+
+list_all(Fun, Args, Config, Key, TokenAlias, NextToken, Acc) ->
+    UpdArgs = Args ++ [?MAX_RESULTS, NextToken, Config],
+    case erlang:apply(Fun, UpdArgs) of
+        {ok, Map} ->
+            UpdAcc   = Acc ++ maps:get(Key, Map),
+            NewToken = maps:get(TokenAlias, Map, undefined),
+            case NewToken of
+                undefined ->
+                    {ok, #{Key => UpdAcc}};
```
motobob, line 900: e.g. https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AdminListDevices.html

is this really a good way of returning list_all? does not it mingle the return format completely off the non list_all version?

motobob (changes requested):
first pass

src/erlcloud_cognito.erl
arthurnovak: `list_all` returns data in same format as non list_all, but skips `PaginationToken` as it's not needed anymore. So in `AdminListDevices` example it'd look like
```
{
   "Devices": [ 
      ...
   ]
}
```

we can return just `UpdAcc` here and wrap it with `#{Key => ListAllResult}` later if that's what you mean

motobob: probably i'm missing something. let me check one more thing

motobob (approved):
@nalundgaard @kkuzmin any objections/oppinions?

src/erlcloud_cognito.erl
```diff
@@ -0,0 +1,915 @@
...
+
+request(Config0, OperationName, Request) ->
+    case erlcloud_aws:update_config(Config0) of
+        {ok, Config} ->
+            Body       = jsx:encode(Request),
+            Operation  = "AWSCognitoIdentityProviderService." ++ OperationName,
+            Headers    = get_headers(Config, Operation, Body),
+            AwsRequest = #aws_request{service         = 'cognito-idp',
+                                      uri             = get_url(Config),
+                                      method          = post,
```
kkuzmin, line 859: I wonder if all API requests are actually POSTs, even like `AdminGetUser`?

arthurnovak: it looks okay, at least it works with any other services like that 🤷

Amershan: As far as i know it is a post

src/erlcloud_cognito.erl
```diff
@@ -0,0 +1,917 @@
...
+        Error   -> Error
+    end.
+
+make_request_body(Mandatory, Optional) ->
+    maps:merge(Mandatory, erlcloud_util:filter_empty_map(Optional)).
+
+get_headers(#aws_config{cognito_host = Host} = Config, Operation, Body) ->
+    Headers = [{"host", Host},
+               {"x-amz-target", Operation},
+               {"version", ?API_VERSION},
```
kkuzmin, line 880: According to this [doc](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/CommonParameters.html) both `Operation` and `version` should be query parameters, not headers.

kkuzmin (changes requested):

src/erlcloud_cognito.erl
arthurnovak: `host` and `x-amz-target` should come as headers

arthurnovak: this [pdf](https://docs.aws.amazon.com/cognitoidentity/latest/APIReference/cognito-identity-api.pdf) contains request examples with target as a header

not sure about version though.. do we really need the version? It wasn't used in other modules afaik, so probably just skip it. What's you thoughts?

Amershan: I have utilized this:
`erlcloud_aws:sign_v4(post, "/", Config, Headers, Body, Region, "cognito-idp", [{"version", ?API_VERSION}])` to send the version as QS and got this error back from aws:
**The request signature we calculated does not match the signature you provide**

If it is added to the header and use the:
`erlcloud_aws:sign_v4_headers(Config, Headers, Body, Region, "cognito-idp")` 
the request does return  the desired response.

kkuzmin: @arthurnovak it looks like `x-amx-target` header is used incorrectly. In this implementation it is always `AWSCognitoIdentityProviderService.DescribeIdentityProvider ` 
From your example:
> X-AMZ-TARGET: com.amazonaws.cognito.identity.model.AWSCognitoIdentityService.DescribeIdentityPool

It looks like we're using two separate service APIs here: [Amazon Cognito Federated Identities](https://docs.aws.amazon.com/cognitoidentity/latest/APIReference/API_DescribeIdentityPool.html) and [Amazon Cognito](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_AdminGetUser.html)

BTW, API version for Amazon Cognito Federated Identities is 2014-06-30

kkuzmin (changes requested):

kkuzmin (changes requested):
I think we need to split this change into two modules to seprate the APIs: `erlcloud_cognito_idp.erl` and `erlcloud_cognito_identity.erl`.

src/erlcloud_cognito.erl
Amershan: If you referring to the create/describe/update identity providers like this:
[https://github.com/Amershan/erlcloud/blob/cognito/src/erlcloud_cognito.erl#L507](https://github.com/Amershan/erlcloud/blob/cognito/src/erlcloud_cognito.erl#L507)

These are the user pool identity providers not the federated ones. See [here](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_DescribeIdentityProvider.html)

arthurnovak: @kkuzmin it was just an example to show that `x-amz-target` can be used in headers and it definitely works not only for Cognito interface, but in other erlcloud modules. If we want to investigate if headers are used correctly further, i'm happy to be involved in that. However, it'll affect whole erlcloud and seems to be out of scope of this Cognito interface implementation.

as @Amershan mentioned above, this implementation covers only User Pools interface and not Identity Pools one (shouldn't be mixed up with identity providers, which belong to User Pools). Confusing aws naming :/

Amershan: @kkuzmin 
Renamed it to erlcloud_cognito_user_pools

include/erlcloud_aws.hrl
```diff
@@ -37,6 +37,8 @@
           s3_bucket_access_method=vhost::vhost|path|auto,
           s3_bucket_after_host=false::boolean(),
           sdb_host="sdb.amazonaws.com"::string(),
+          cognito_host="cognito-idp.eu-west-1.amazonaws.com"::string(),
```
kkuzmin, line 40: please rename `cognito_host` to `cognito_idp_host`.
Same for the scheme below

kkuzmin (changes requested):

test/erlcloud_cognito_tests.erl
```diff
@@ -0,0 +1,1059 @@
+-module(erlcloud_cognito_tests).
```
kkuzmin, line 1: please rename this module as well.

kkuzmin (changes requested):

include/erlcloud_aws.hrl
Amershan: @kkuzmin  done

kkuzmin (approved):

Amershan:
@motobob Should i fixup the commits to one?

motobob:
>@motobob Should i fixup the commits to one?

git can do it for us)

Amershan:
@motobob fixed the specifications

PR #730, opened 2022-07-27
Fix markdown issues raised by markdownlint

kianmeng:
See detailed description of the rules is available at
https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,14 @@
-# erlcloud: AWS APIs library for Erlang #
+# erlcloud: AWS APIs library for Erlang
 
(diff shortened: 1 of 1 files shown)


motobob:
i'm not fan of it but ok

kkuzmin (approved):

paulo-ferraz-oliveira:
This is especially interesting when integrated with CI, e.g. as in https://github.com/2Latinos/nuntius/blob/main/.github/workflows/lint.yml#L14

Markdownlint, on the other hand, is also very configurable...

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

PR #731, opened 2022-08-01
Fix reading error message from failed DDB requests

gabrieltholsgard:
I noticed when trying to handle specific errors that reading the message from the DDB error body wasn't correct resulting in an empty message.
This fixes such that we will get this message.
For example:
```
(infranet@127.0.0.1)42> erlcloud_ddb2:put_item(<<"a_table">>, [{<<"id">>, <<"dummy">>}, {<<"some_attr">>, Data}]).
{error,{<<"ValidationException">>,<<>>}}
```
with this fix the result is:
```
(infranet@127.0.0.1)45> erlcloud_ddb2:put_item(<<"a_table">>, [{<<"id">>, <<"dummy">>}, {<<"some_attr">>, Data}]).
{error,{<<"ValidationException">>,
        <<"Item size has exceeded the maximum allowed size">>}}
```

### DIFF
--- a/src/erlcloud_ddb_impl.erl
+++ b/src/erlcloud_ddb_impl.erl
@@ -248,7 +248,7 @@ client_error(Body, DDBError) ->
                 undefined ->
                     DDBError#ddb2_error{error_type = http, should_retry = false};
                 FullType ->
-                    Message = proplists:get_value(<<"Message">>, Json, <<>>),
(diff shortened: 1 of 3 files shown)


motobob (approved):

motobob:
good catch but please fix the tests too and good to go

PR #733, opened 2022-10-20
Nat gateway id not reflect in route table issue solved.

hardi171:
Intent of this PR, reflect NatGatewayId in the route table with other GateWayID. 

Problem was : the customer has one route with same destination cidr gateway=nat gateway and another with the same destination cidr but getway=vpn gateway, but the response we get from erlcloud when describing route tables doesn’t reflect.


src/erlcloud_ec2.erl
```diff
@@ -1989,7 +1989,12 @@ extract_route(Node) ->
 extract_route_set(Node) ->
     [
      {destination_cidr_block, get_text("destinationCidrBlock", Node)},
-     {gateway_id, get_text("gatewayId", Node)},
+%%   {gateway_id, get_text("gatewayId", Node)},
```
kkuzmin, line 1992: for the sake of simplicity I suggest we just have both
```
{gateway_id, get_text("gatewayId", Node)},
{nat_gateway_id, get_text("natGatewayId", Node)},
```

rebar.lock
```diff
@@ -1,4 +1,4 @@
-{"1.1.0",
+{"1.2.0",
```
kkuzmin, line 1: please revert.

kkuzmin (changes requested):

rebar.lock
hardi171: ok

src/erlcloud_ec2.erl
hardi171: done.

motobob (approved):

kkuzmin (approved):

PR #736, opened 2022-11-03
Support access-analyzer service

alexo-spb:
A new module is added to support basic operations to manipulate access analyzers.

See https://docs.aws.amazon.com/access-analyzer/latest/APIReference/Welcome.html

### DIFF
--- a/src/erlcloud_access_analyzer.erl
+++ b/src/erlcloud_access_analyzer.erl
@@ -1,0 +1,240 @@
+-module(erlcloud_access_analyzer).
+
+-include("erlcloud.hrl").
+-include("erlcloud_aws.hrl").
+
+%% API
+-export([
+    list_analyzers/1,
+    list_analyzers/2,
+    list_analyzers_all/1,
+    list_analyzers_all/2,
+    get_analyzer/2,
+    create_analyzer/2,
+    delete_analyzer/2
+]).
+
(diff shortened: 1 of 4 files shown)


src/erlcloud_access_analyzer.erl
```diff
@@ -0,0 +1,244 @@
...
+        request_headers = SignedHeaders,
+        request_body = Payload
+    }.
+
+norm_path(Path) ->
+    binary_to_list(iolist_to_binary(["/" | lists:join("/", Path)])).
+
+norm_params([] = _Params) ->
+    "";
+norm_params(Params) ->
```
kkuzmin, line 196: looks like we need to export this [function](https://github.com/erlcloud/erlcloud/blob/master/src/erlcloud_aws.erl#L1197) and use it here instead of `norm_params()`

alexo-spb: exported

src/erlcloud_aws.erl
```diff
@@ -1195,13 +1201,18 @@ canonical_headers(Headers) ->
 canonical_query_string([]) ->
     "";
 canonical_query_string(Params) ->
```
kkuzmin, line 1203: please add a`-spec` to this function since it is now exported.

alexo-spb: done

PR #743, opened 2023-01-11
add workgroup option to start, create, list query API calls

nastiona-kitten:
Add workgroup option to start_query_execution, create_named_query, list_named_queries and list_query_executions API calls.

### DIFF
--- a/src/erlcloud_athena.erl
+++ b/src/erlcloud_athena.erl
@@ -437,7 +437,7 @@ list_query_executions(Config) when is_record(Config, aws_config) ->
 list_query_executions(PaginationMap) when is_map(PaginationMap) ->
     list_query_executions(PaginationMap, [], default_config()).
 
--spec list_query_executions(map(), aws_config()) ->
+-spec list_query_executions(map(), aws_config() | list_query_executions_opts()) ->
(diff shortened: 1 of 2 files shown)


motobob (approved):

src/erlcloud_athena.erl
```diff
@@ -566,6 +632,16 @@ get_url(#aws_config{athena_scheme = Scheme,
...
+encode_params(Params) ->
+  encode_params(Params, []).
+
+encode_params([], Acc) ->
+  maps:from_list(Acc);
+encode_params([{_, undefined} | T], Acc) ->
+  encode_params(T, Acc);
+encode_params([{description, Description} | T], Acc) when is_binary(Description) ->
+  encode_params(T, [{<<"Description">>, Description} | Acc]);
+encode_params([{workgroup, WorkGroup} | T], Acc) when is_binary(WorkGroup) ->
```
motobob, line 644: optional:
I would eliminate this atom to binary conversion and use binary directly. but optional as other modules here often do this

nastiona-kitten: I wanted to make this module more consistent with other modules. Sort of less thinking required from the usage perspective.

PR #744, opened 2023-02-16
Nat gateway id not reflect in route table issue solved.

hardi171:


src/erlcloud_securityhub.erl
```diff
@@ -0,0 +1,127 @@
+-module(erlcloud_securityhub).
```
motobob, line 1: need tests for the module

PR #752, opened 2023-05-16
Remove rebar2 compatibility

savonarola:
And fix mix compatibility.
We are unlikely to start new `rebar2` projects, but `rebar.config.script` prevents the project from working properly as a `mix` dependency.


motobob:
good idea to remove rebar2. clean it up and i'll all for it

PR #754, opened 2023-05-29
Support IMDSv2 session token authentication

dsrosario:
This PR allows the consumer to use IMDSv2. Checks 
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-metadata-v2-how-it-works.html for the detail.


motobob (approved):
a better replacemtns of #750

PR #759, opened 2023-09-21
Add hex.pm publishing section to readme, badges, make target

rmpalomino:
* Add hex.pm version and license badges to readme
* Add short hex.pm publishing section to readme
* Add hex-publish make target to publish a new hex package from the latest version of the libary


carlisom (approved):
➕ 💯 on adding the new  badges

PR #762, opened 2023-12-09
Fix for s3 domains: cn-northwest-1 & cn-north-1

ariel-anieli:
Wrote down the proposal of @devdutt-fp


ariel-anieli:
@carlisom, could you have a look?

carlisom (approved):

PR #783, opened 2025-02-14
efs: Implement `DescribeFileSystems` API

scott-jones-AlertLogic:
Introduce support for the `elasticfilesystem:DescribeFileSystems` operation via the `erlcloud_efs` module.

### Reference
- [DescribeFileSystems - Amazon Elastic File System - API Reference](https://docs.aws.amazon.com/efs/latest/ug/API_DescribeFileSystems.html)


src/erlcloud_efs.erl
```diff
@@ -0,0 +1,203 @@
...
+    case describe_file_systems(AwsConfig, [{<<"Marker">>, Marker0} | Params]) of
+        {ok, FileSystems} ->
+            FileSystems1 = lists:append(FileSystems0, FileSystems),
+            {ok, FileSystems1};
+        {ok, FileSystems, Marker1} ->
+            FileSystems1 = lists:append(FileSystems0, FileSystems),
+            describe_file_systems_next(AwsConfig, Params, Marker1, FileSystems1);
+        {error, Reason} ->
+            {error, Reason}
+    end.
```
rmpalomino, line 107: These appends are not ideiomatic. The Erlangy thing to do would be what erlcloud_ddb_util:flatreverse does. I'm not sure how bad the efficiency impact would actually be realistically with EFS though: https://github.com/erlcloud/erlcloud/blob/ebc0bf4ccc27eff7dddf843404c73723b72180e4/src/erlcloud_ddb_util.erl#L574-L576

src/erlcloud_efs.erl
```diff
@@ -0,0 +1,203 @@
...
+    Request#aws_request{should_retry = false}.
+
+decode_response(AwsRequest) ->
+    case AwsRequest#aws_request.response_body of
+        <<>> ->
+            ok;
+        ResponseBody ->
+            Json = jsx:decode(ResponseBody, [{return_maps, false}]),
+            {ok, Json}
+    end.
```
rmpalomino, line 179: These could all be simplified by pattern matching on an aws_request, e.g.

```erlang
should_retry(#aws_request{response_type = ..., response_status = ...} = Request) when ... ->
     ...

decode_response(#aws_request{response_body = <<>>}) ->
    ....;
decode_response(#aws_request{response_body = ResponseBody}) ->
   ....
```

rmpalomino (commented):
Please take a look, some of this code doesn't feel as Erlangy as it could be.

scott-jones-AlertLogic:
Thanks @rmpalomino! I've rebased my changes to keep the history clean, but the changes made are:
- replace inefficient repeated `lists:append(Acc, L)` calls with more efficient `lists:append(lists:reverse(ListOfListsOfFileSystemProplist))`
- prefer pattern matching directly in the function head over doing so in guards or case statements
- `erlcloud_efs:describe_file_systems()` calls now always return a pagination token (`binary() | undefined`) in a successful result, similar to how the `erlcloud_ec2` module implements its `describe_thing()` and `describe_thing_all()` functions

rmpalomino (approved):

